PR #4, opened 2017-12-14
Swift4 migration.

iccub:
I wanted to check how easy will migration to swift 4 be.

Steps I took:

1. Updated carthage libraries, bumped versions and changed owners of two repos to, check first commit for more details.
2. Upadted sqlite.swift lib. Don't know why, but the don't include this lib with carthage but by manually copying it to `Third-Party Source` folder. So I grabbed newer version and updated it.
3. Hit 'convert to modern swift syntax' on all but `Client` targets, first I wanted to check if frameworks work well. There is problem with one method `effect` from `DeferredUtils` and it block most of frameworks to convert to swift4. Xcode allows to mix 3.2 and 4.0 now so there is no problem with leaving some targets on older versions. Specific targets that are still on 3.2:
    - NotificationService
    - SendTo
    - ShareTo
    - ViewLater
    - Storage
    - Sync
4. Client target conversion, went smoothly, ~80% of code was converted rest needed to be done manually. There were also few classes that belonged to both Client and one of 3.2 targets so I had to use `#if swift(>=4.0)`. There are also parts of code where I added just to make it compile, these are marked with `// FIXME: swift 4` so they can be easily located.
5. App compiled and seems to run fine on ip7 simulator.

Few notes:
- ~~Haven't checked tests yet.~~
- 99% of problems are caused by just one method [DeferredUtils.effect()](https://github.com/brave/brave-ios/blob/27b67852d2f821733cb4eea1ec67fdd9aba54de3/Shared/DeferredUtils.swift#L132)
Just change swift version of Storage target and build it to see the error. Deferred lib is still not updated to swift 4 from what i've seen so maybe this is lib problem, not the `effect` method.
On other hand, most of this deferred usage seems to be related to sqlite, sync and accounts which we won't probably use at all, so maybe it's not worth fixing that, depending on how big changes we are going to make.
- Diff shows 1500 deleted lines, this is related to new sqlite lib.


iccub:
Update: all tests have been updated to swift4, including l10n screenshots

iccub:
Ok so now the project is 100% converted to swift 4, all looks good to me, no workarounds needed anymore.

jhreis:
No longer needed after re-fork.

PR #6, opened 2017-12-18
Brave branding

jhreis:


Client/Assets/Images.xcassets/AppIcon_Developer.appiconset/Contents.json
```diff
@@ -1,63 +1,67 @@
 {
   "images" : [
     {
-      "idiom" : "iphone",
       "size" : "20x20",
+      "idiom" : "iphone",
+      "filename" : "icon-40.png",
```
iccub, line 6: Pedantic, but what about a more descriptive filename? `appicon` would suffice.

iccub (commented):
2 questions:
- are we moving to white background logo? Brave 1.x has dark background
- Not all app icon dimensions are filled compared to Brave 1.x, any reason for that? I see FF doesn't have them filled too.

Other that that, looks good to me.

Client/Assets/Images.xcassets/AppIcon_Developer.appiconset/Contents.json
jhreis: Good thought, I don't think it is a huge deal though, since it is under the `AppIcon.appiconset` which is always what we interact with in Xcode.

jhreis:
- Yes, we are moving to a white background, part of the v2 refresh
- We just replaced whatever FF was using, we will be replacing some of these icons in the future (e.g. for beta) and can see if we need additional coverage.

iccub (approved):

PR #9, opened 2017-12-19
Updated app icon, updated colors and other theme styles, update launc…

jamesmudgett:
Consolidated colors throughout the app. Simplified color naming, replaced icons. More icons and colors to come at a later date.

- [ ] My patch has a standard commit message that looks like `Bug 12345678 - This fixes something something`

## Screenshots


Client/Frontend/Browser/TabToolbar.swift
```diff
@@ -244,14 +244,14 @@ class TabToolbar: UIView {
     }
 
     override func draw(_ rect: CGRect) {
-        if let context = UIGraphicsGetCurrentContext() {
-            drawLine(context, start: CGPoint(x: 0, y: 0), end: CGPoint(x: frame.width, y: 0))
-        }
+//        if let context = UIGraphicsGetCurrentContext() {
```
jhreis: Just remove, if we don't want this.

jhreis (approved):
lgtm

PR #14, opened 2018-03-21
Tabs bar

iccub:
Removed existing firefox implementation of tabs on iPads and copied tabs from 1.0.


iccub:
Added ability to move tabs in the tabs tray.

Client/Frontend/BraveUX.swift
```diff
@@ -0,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+
+struct BraveUX {
```
jhreis: What about just naming this `UX`?

Client/Frontend/Browser/TabManager.swift
```diff
@@ -130,6 +130,21 @@ class TabManager: NSObject {
...
     }
 
+    var currentDisplayedIndex: Int? {
+        assert(Thread.isMainThread)
+
+        guard let selectedTab = self.selectedTab else {
+            return nil
+        }
+
+        return UIApplication.isInPrivateMode ? privateTabs.index(of: selectedTab) : normalTabs.index(of: selectedTab)
```
jhreis: This should be able to just use:
`return displayedTabsForCurrentPrivateMode.index(of: selectedTab)`

Since that variable will automatically give us the correct `[Tab]`

Client/Frontend/Browser/TabScrollController.swift
```diff
@@ -52,11 +53,17 @@ class TabScrollingController: NSObject {
...
-            headerTopConstraint?.update(offset: headerTopOffset)
+            headerTopConstraint?.update(offset: headerTopOffset - tabsBarOffset)
             header?.superview?.setNeedsLayout()
         }
     }
 
+    fileprivate var tabsBarOffset: CGFloat {
+        guard let tabsBar = tabsBar else { return 0 }
+
+        return (scrollDirection == .down && tabsBar.view.isHidden == false) ? BraveUX.TabsBar.height : 0
```
jhreis: `isHidden` here is not an optional, right? So can just use `!tabsBar.view.isHidden`?

Client/Frontend/Browser/TabTrayController.swift
```diff
@@ -778,6 +788,77 @@ fileprivate class TabManagerDataSource: NSObject, UICollectionViewDataSource {
...
+            if index < urls.count {
+                url = urls[index]
+            }
+        }
+
+        // Ensure we actually have a URL for the tab being dragged and that the URL is not local.
+        // If not, just create an empty `NSItemProvider` so we can create a drag item with the
+        // `Tab` so that it can at still be re-ordered.
+        var itemProvider: NSItemProvider
+        if url != nil, !(url?.isLocal ?? true) {
```
jhreis: We should unwrap `url` here, much cleaner:

`if let url = url, !url.isLocal`

Client/Frontend/Browser/TabTrayController.swift
```diff
@@ -778,6 +788,77 @@ fileprivate class TabManagerDataSource: NSObject, UICollectionViewDataSource {
...
+            itemProvider = NSItemProvider(contentsOf: url) ?? NSItemProvider()
+        }  else {
+            itemProvider = NSItemProvider()
+        }
+
+        let dragItem = UIDragItem(itemProvider: itemProvider)
```
jhreis: Rather than having the `?? NSItemProvider()` above, let's put it here instead:

`let dragItem = UIDragItem(itemProvider: itemProvider ?? .init())`

Client/Frontend/Browser/TabsBar/TabBarCell.swift
```diff
@@ -0,0 +1,124 @@
...
+        closeButton.snp.makeConstraints{ make in
+            make.top.bottom.equalTo(self)
+            make.right.equalTo(self).inset(2)
+            make.width.equalTo(30)
+        }
+
+        separatorLine.snp.makeConstraints { make in
+            make.left.equalTo(self)
+            make.width.equalTo(0.5)
+            make.height.equalTo(self)
```
jhreis, line 76: Let's be consistent here. Above we combine `top.bottom`, here `left` and `height` are separated.

Client/Frontend/Browser/TabsBar/TabBarCell.swift
```diff
@@ -0,0 +1,124 @@
...
+            make.centerY.equalTo(self.snp.centerY)
+        }
+
+        separatorLineRight.snp.makeConstraints { make in
+            make.right.equalTo(self)
+            make.width.equalTo(0.5)
+            make.height.equalTo(self)
```
jhreis, line 83: Ditto.

Client/Frontend/Browser/TabsBar/TabBarCell.swift
```diff
@@ -0,0 +1,124 @@
...
+            make.centerY.equalTo(self.snp.centerY)
+        }
+    }
+
+    override var isSelected: Bool {
+        didSet(selected) {
+            closeButton.tintColor = UIApplication.isInPrivateMode ? UIColor.white : UIColor.black
```
jhreis, line 90: We may need to revisit using `isInPrivateMode` to set styles, we should be creating style classes for this instead. But we should probably talk a bit more about this.

Client/Frontend/Browser/TabsBar/TabsBarViewController.swift
```diff
@@ -0,0 +1,307 @@
...
+        view.bounces = false
+        view.delegate = self
+        view.dataSource = self
+        view.allowsSelection = true
+        view.decelerationRate = UIScrollViewDecelerationRateNormal
+        view.register(TabBarCell.self, forCellWithReuseIdentifier: "TabCell")
+        return view
+    }()
+
+    private let leftOverflowIndicator = CAGradientLayer()
```
jhreis: Move these on top of the lazy vars

Client/Frontend/Browser/TabsBar/TabsBarViewController.swift
```diff
@@ -0,0 +1,307 @@
...
+    private let rightOverflowIndicator = CAGradientLayer()
+
+    weak var tabManager: TabManager?
```
jhreis: Can we make this private and take it in via the `init`?

Client/Frontend/Browser/TabsBar/TabsBarViewController.swift
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    func addTabPressed() {
+        tabManager?.addTabAndSelect()
+    }
+    
+    func updateData() {
+        tabList = WeakList<Tab>()
+
+        guard let tabsToUpdate = UIApplication.isInPrivateMode ? tabManager?.privateTabs : tabManager?.normalTabs else {
```
jhreis: Can just use `displayedTabsForCurrentPrivateMode` instead, right?

Client/Frontend/Browser/TabsBar/TabsBarViewController.swift
```diff
@@ -0,0 +1,307 @@
...
+        }
+
+        overflowIndicators()
+        reloadDataAndRestoreSelectedTab()
+    }
+    
+    func reloadDataAndRestoreSelectedTab() {
+        collectionView.reloadData()
+
+        if let selectedTab = tabManager?.selectedTab, let tabManager = tabManager {
```
jhreis: Unwrap `tabManager` first, as you use that for the `selectedTab`

Client/Frontend/Browser/TabsBar/TabsBarViewController.swift
```diff
@@ -0,0 +1,307 @@
...
+    
+    func handleLongGesture(gesture: UILongPressGestureRecognizer) {
+        switch(gesture.state) {
+        case UIGestureRecognizerState.began:
+            guard let selectedIndexPath = self.collectionView.indexPathForItem(at: gesture.location(in: self.collectionView)) else {
+                break
+            }
+            collectionView.beginInteractiveMovementForItem(at: selectedIndexPath)
+        case UIGestureRecognizerState.changed:
+            collectionView.updateInteractiveMovementTargetPosition(gesture.location(in: gesture.view!))
```
jhreis: Remove force unwrapping here.

Client/Frontend/Browser/TabsBar/TabsBarViewController.swift
```diff
@@ -0,0 +1,307 @@
...
+            let selectedIndex = tabList.index(of: selectedTab) ?? 0
+            if selectedIndex < tabList.count() {
+                collectionView.selectItem(at: IndexPath(row: selectedIndex, section: 0), animated: (!tabManager.isRestoring), scrollPosition: .centeredHorizontally)
+            }
+        }
+    }
+    
+    func handleLongGesture(gesture: UILongPressGestureRecognizer) {
+        switch(gesture.state) {
+        case UIGestureRecognizerState.began:
```
jhreis: Should be able to just use `.began`, `.ended`, etc...

Client/Frontend/Browser/TabsBar/TabsBarViewController.swift
```diff
@@ -0,0 +1,307 @@
...
+        case UIGestureRecognizerState.ended:
+            collectionView.endInteractiveMovement()
+        default:
+            collectionView.cancelInteractiveMovement()
+        }
+    }
+
+    private func tabOverflowWidth(_ tabCount: Int) -> CGFloat {
+        let overflow = CGFloat(tabCount) * BraveUX.TabsBar.minimumWidth - collectionView.frame.width
+        return overflow > 0 ? overflow : 0
```
jhreis: `return max(overflow, 0)` is cleaner

Client/Frontend/Browser/TabsBar/TabsBarViewController.swift
```diff
@@ -0,0 +1,307 @@
...
+        // super lame place to put this, need to find a better solution.
+        plusButton.tintColor = UIApplication.isInPrivateMode ? UIColor.white : UIColor.black
+        collectionView.backgroundColor = UIApplication.isInPrivateMode ? UIColor(white: 0.0, alpha: 0.2) : UIColor(white: 0.0, alpha: 0.075)
+
+        addScrollHint(for: .leftSide, maskLayer: leftOverflowIndicator)
+        addScrollHint(for: .rightSide, maskLayer: rightOverflowIndicator)
+        
+        let offset = Float(collectionView.contentOffset.x)
+        let startFade = Float(30)
+        if offset < startFade {
```
jhreis: Can remove this entire if/else block
`leftOverflowIndicator.opacity = min(1, offset / startFade)`
No divide issues here as `startFade` is a constant.

Would need to read docs, but pretty sure it is totally fine to set `opacity` > `1` too, so just setting it to `offset / startFade` may work fine.

Client/Frontend/Browser/TabsBar/TabsBarViewController.swift
```diff
@@ -0,0 +1,307 @@
...
+            leftOverflowIndicator.opacity = offset / startFade
+        } else {
+            leftOverflowIndicator.opacity = 1
+        }
+        
+        // all the way scrolled right
+        let offsetFromRight = collectionView.contentSize.width - CGFloat(offset) - collectionView.frame.width
+        if offsetFromRight < CGFloat(startFade) {
```
jhreis: Same as above, can remove this if/else

Client/Frontend/Browser/TabsBar/TabsBarViewController.swift
```diff
@@ -0,0 +1,307 @@
...
+        let tab = tabList.at(indexPath.row)
+        tabManager?.selectTab(tab)
+    }
+}
+
+// MARK: - UICollectionViewDelegateFlowLayout
+extension TabsBarViewController: UICollectionViewDelegateFlowLayout {
+    func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, sizeForItemAt indexPath: IndexPath) -> CGSize {
+        if tabList.count() == 1 {
+            return CGSize(width: view.frame.width, height: view.frame.height)
```
jhreis: We can just `return view.frame.size`

Client/Frontend/Browser/TabsBar/TabsBarViewController.swift
```diff
@@ -0,0 +1,307 @@
...
+        }
+
+        let newTabButtonWidth = CGFloat(UIDevice.current.userInterfaceIdiom == .pad ? BraveUX.TabsBar.buttonWidth : 0)
+        let tabsAndButtonWidth = CGFloat(tabList.count()) * BraveUX.TabsBar.minimumWidth
+        if tabsAndButtonWidth < collectionView.frame.width - newTabButtonWidth {
+            let maxWidth = (collectionView.frame.width - newTabButtonWidth) / CGFloat(tabList.count())
```
jhreis: Just want to verify that `tabList` cannot be empty at this point, right? If it can, this will throw a divide by 0 error.

Client/Frontend/Browser/TabsBar/TabsBarViewController.swift
```diff
@@ -0,0 +1,307 @@
...
+        guard let tab = tabList.at(indexPath.row) else { return cell }
+
+        cell.tabManager = tabManager
+        cell.tab = tab
+        cell.titleLabel.text = tab.displayTitle
+        cell.currentIndex = indexPath.row
+        cell.separatorLineRight.isHidden = (indexPath.row != tabList.count() - 1)
+
+        cell.closeTabCallback = { [weak self] tab in
+            guard let strongSelf = self, let tabManager = self?.tabManager, let previousIndex = self?.tabList.index(of: tab) else { return }
```
jhreis: After unwrapping `self`, we should use that in the other unwrappings. Also, probably just cleaner to do:

`guard let `self` = self, let ...`, rather than giving it a new name. Makes it harder to explicitly use the incorrect one.

Shared/WeakList.swift
```diff
@@ -53,6 +53,18 @@ open class WeakList<T: AnyObject>: Sequence {
             return nil
         }
     }
+
+    open func count() -> Int {
+        return self.items.count
+    }
+
+    open func at(_ index: Int) -> T? {
```
jhreis: Thoughts about using Swift subscript here? Would use like:

```
let list = WeakList<String> // fill with data
let secondItem = list[1]
```

See [docs](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Subscripts.html).

jhreis (changes requested):

Client/Frontend/BraveUX.swift
iccub: There is also `UIConstants` class already, I could merge it together.

Client/Frontend/Browser/TabsBar/TabBarCell.swift
iccub: Edge constraints(`top|bottom|left|right`) and `width|height` values are separate concepts, personally, I would not combine them.

Client/Frontend/Browser/TabsBar/TabsBarViewController.swift
iccub: I took `strongSelf` convention from https://github.com/raywenderlich/swift-style-guide#extending-object-lifetime. We can decide whether we want to adopt this style over 'self'

Shared/WeakList.swift
iccub: For some reason it throws `Type 'WeakList<Tab>' has no subscript members`, but the same code works fine in swift playgrounds. Looks like it doesn't work with swift 3.3

jhreis: Did you create a subscript? You will need to implement the functions yourself, or is that what you are referring to?

Client/Frontend/Browser/TabManager.swift
```diff
@@ -278,14 +293,10 @@ class TabManager: NSObject {
...
-        
-        let currentTabs = privateMode ? privateTabs : normalTabs
-        let fromIndex = tabs.index(of: currentTabs[visibleFromIndex]) ?? tabs.count - 1
-        let toIndex = tabs.index(of: currentTabs[visibleToIndex]) ?? tabs.count - 1
-        
+
         let previouslySelectedTab = selectedTab
         
-        tabs.insert(tabs.remove(at: fromIndex), at: toIndex)
+        tabs.insert(tabs.remove(at: visibleFromIndex), at: visibleToIndex)
```
jhreis: This should be moved to the model, _eventually_. Given that we are swapping out our database, may not make sense to do now, but definitely model logic.

Client/Frontend/Browser/TabTrayController.swift
```diff
@@ -778,6 +788,75 @@ fileprivate class TabManagerDataSource: NSObject, UICollectionViewDataSource {
...
+        let dragItem = UIDragItem(itemProvider: itemProvider ?? .init())
+        dragItem.localObject = tab
+        return [dragItem]
+    }
+}
+
+@available(iOS 11.0, *)
+extension TabManagerDataSource: UICollectionViewDropDelegate {
+    func collectionView(_ collectionView: UICollectionView, performDropWith coordinator: UICollectionViewDropCoordinator) {
+        guard isDragging, let destinationIndexPath = coordinator.destinationIndexPath, let dragItem = coordinator.items.first?.dragItem, let tab = dragItem.localObject as? Tab, let sourceIndex = tabs.index(of: tab) else {
```
jhreis: Add some line breaks 😄

Client/Frontend/Browser/TabTrayController.swift
```diff
@@ -778,6 +788,75 @@ fileprivate class TabManagerDataSource: NSObject, UICollectionViewDataSource {
...
+    }
+
+    func collectionView(_ collectionView: UICollectionView, dropSessionDidUpdate session: UIDropSession, withDestinationIndexPath destinationIndexPath: IndexPath?) -> UICollectionViewDropProposal {
+        guard let localDragSession = session.localDragSession, let item = localDragSession.items.first, let tab = item.localObject as? Tab else {
+            return UICollectionViewDropProposal(operation: .forbidden)
+        }
+
+        // If the tab doesn't exist by the time we get here, we must return a
+        // `.cancel` operation continuously until `isDragging` can be reset.
+        if !isDragging, tabs.index(of: tab) == nil {
```
jhreis: Since no unwrappings here I think using `&&` is probably more clear.

Client/Frontend/Browser/TabTrayController.swift
```diff
@@ -778,6 +788,75 @@ fileprivate class TabManagerDataSource: NSObject, UICollectionViewDataSource {
...
+        isDragging = false
+
+        let destinationIndex = destinationIndexPath.item
+        tabManager.moveTab(isPrivate: tab.isPrivate, fromIndex: sourceIndex, toIndex: destinationIndex)
+        tabs.insert(tabs.remove(at: sourceIndex), at: destinationIndex)
+        collectionView.moveItem(at: IndexPath(item: sourceIndex, section: 0), to: destinationIndexPath)
+    }
+
+    func collectionView(_ collectionView: UICollectionView, dropSessionDidUpdate session: UIDropSession, withDestinationIndexPath destinationIndexPath: IndexPath?) -> UICollectionViewDropProposal {
+        guard let localDragSession = session.localDragSession, let item = localDragSession.items.first, let tab = item.localObject as? Tab else {
```
jhreis: Line breaks may be nice here too.

Client/Frontend/Browser/TabsBar/TabBarCell.swift
jhreis: Gotcha, I think that makes sense. Let's remember this for future PRs.

Shared/WeakList.swift
iccub: Nvm needed to add `open` to it as WeakList is in `Shared` module 🤷‍♂️

iccub:
Closing in favor of https://github.com/brave/brave-ios/pull/50

PR #25, opened 2018-05-01
Adblock iframe module

jhreis:
This is the current Shields code we will be utilizing. It breaks Shields into its own _dependent_ framework. In the future it will probably be broken into an independent one.


jhreis:
Closing this, as we will not be using it currently, will plan on keeping the branch around though.

PR #51, opened 2018-06-14
Removed leanplum from project file linking.

jhreis:


kylehickinson (approved):
Looks good to me. Probably still compiled before because the framework was still in the build products directory?

PR #115, opened 2018-07-17
Fix #114: Deleted OldStrings.swift to avoid translating unnecessary s…

### LINKED ISSUE #114: Delete Unused NSLocalizedStrings
These strings get clumped into our l10n translation bucket, and we should avoid translating unnecessary strings, as we are starting from scratch.

jhreis:
…trings.

*Thank you for submitting a pull request, your contributions are greatly appreciated!*

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.
## Pull Request Checklist

## Screenshots


Client.xcodeproj/project.pbxproj
```diff
@@ -4294,7 +4288,6 @@
 				D3A9949D1A3686BD008AD1AC /* Tab.swift in Sources */,
 				EB11A1042044A90E0018F749 /* ContentBlockerHelper+Whitelist.swift in Sources */,
 				A93067E81D0FE18E00C49C6E /* NightModeHelper.swift in Sources */,
-				3B39EDCB1E16E1AA00EF029F /* CustomSearchViewController.swift in Sources */,
```
tmancey, line 4297: AppSettingsTableViewController.swift, CustomSearchViewController.swift and RequirePasscodeIntervalViewController.swift have also been removed from the project, was this intentional?

jhreis: No, odd, thanks for pointing out...

jhreis: So looked into this. Most of these files are actually already deleted, and are not linked in Xcode, must be something I triggered when I deleted the OldStrings file. Model8 is still linked, so must just be de-linking from this specific place. 

I tested this, and it seems to work well, you can verify, but these files will keep going red on us, until they are committed somehow.

tmancey: 👍 maybe worth raising another ticket to manually remove these orphans from the project file?

tmancey (approved):

PR #118, opened 2018-07-18
Fix #89: Prevent loading home panel while restoring tabs.

### LINKED ISSUE #89: Home panel showing for a split second on restored tab.

iccub:
Check for error page may seem weird but it is used by session restore to trigger a redirect when loaded.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -720,7 +720,7 @@ class BrowserViewController: SensitiveViewController {
                 hideHomePanelController()
                 return
             }
-            if url.isAboutHomeURL {
+            if url.isAboutHomeURL && !url.isErrorPageURL {
                 showHomePanelController(inline: true)
             } else if !url.isLocalUtility || url.isReaderModeURL {
                 hideHomePanelController()


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -720,7 +720,7 @@ class BrowserViewController: SensitiveViewController {
                 hideHomePanelController()
                 return
             }
-            if url.isAboutHomeURL {
+            if url.isAboutHomeURL && !url.isErrorPageURL {
```
tmancey, line 723: Would detecting ```"/about/sessionrestore"``` be a cleaner solution? See ```isLocalUtility``` function in ```URLExtensions.swift```

iccub: The URL to disable is `http://localhost:6571/errors/error.html?url=http%3A//localhost%3A6571/about/home/%23panel%3D0`

At this point the url is unwrapped out of session restore I believe.

tmancey: 👍 my misunderstanding of the issue title "while restoring tabs"

jhreis (approved):
Please use the provided PR template in the future.

PR #124, opened 2018-07-23
Fix #65: Added search engine equality

### LINKED ISSUE #65: Search Engine Equality
Very current search engines are equivalent or a strict superset of current [browser-ios](https://github.com/brave/browser-ios) project.

Also make sure browser source for related search engines is "Brave" (e.g. DDG).

tmancey:
## Pull Request Checklist

## Screenshots


jhreis (changes requested):
From what I can tell these were just directly copied into the project, correct? If so, I think we need to come up with a better approach. We will no longer be able to leverage the existing search plugin script, as that will overwrite all of our changes.

I think we should either find a way to "fork" the existing repos/endpoints that the script pulls from, or put ours in another location and update the `scrape_plugins.py` script to pull upstream changes, and then copy ours over.

tmancey (commented):
That is correct, a new list.json file was created based upon Brave 1.x, along with changes from Firefox in 2.x. I have now added scripts to automatically generate the Brave search engines, please see the README.md file in the Client/Assets/Search directory. The Firefox scripts fail, however if you would like working versions (I can either commit the fixes or provide these files).

Client/Frontend/Browser/DefaultSearchPrefs.swift
```diff
@@ -46,6 +58,9 @@ class DefaultSearchPrefs {
         // If the engineList is empty then go ahead and use the default
         var usersEngineList = engineList ?? defaultSearchList
 
+        // Append "all locales" search engines to the users engine list
+        usersEngineList = usersEngineList + allLocalesSearchList
```
jhreis: `+=` should work here:
```
usersEngineList += allLocalesSearchList
```

Issue #20, opened 2018-04-25
Cleanup Deprecation Warnings


jhreis:
Lots of deprecation or other random warnings. This should deal with any moderately easy warnings to fix. Anything requiring larger refactors or significant API changes should be dealt with individually.

PR #129, opened 2018-07-26
Ref #20: Cleanup warnings part 1

cndouglas:
## Pull Request Checklist

## Notes

Part 1 of #20

This cleans up a few dozen of the warnings that are easiest to fix. Summary of changes:

- Change `flatMap` (deprecated) to `compactMap`
- Change `!` to `?` in method signatures
- Add `String(describing:)` for string interpolation of optionals
- Convert `string.substring` (deprecated) to modern subscripts or `string.suffix(with:)`
- Remove unnecessary casting
- Remove `weak` from variables in protocol declarations
- Add `@discardableResult` where needed
- Change from `string.characters` (deprecated) to just `string`
- Remove unused `let error` in `catch`
- Change `UIApplication.shared.openURL()` (deprecated in iOS 10.0) to `UIApplication.shared.open()`
- Include `self.init()` in UIEdgeInsets initializer

### DIFF
--- a/Data/sync/Sync.swift
+++ b/Data/sync/Sync.swift
@@ -256,7 +256,7 @@ class Sync: JSInjector {
     
     var syncSeedArray: [Int]? {
         let splitBytes = syncSeed?.components(separatedBy: CharacterSet(charactersIn: "[], ")).filter { !$0.isEmpty }
-        let seed = splitBytes?.map{ Int($0) }.flatMap{ $0 }
+        let seed = splitBytes?.compactMap { Int($0) }
         return seed?.count == Sync.SeedByteLength ? seed : nil
     }
     
@@ -508,7 +508,7 @@ extension Sync {
 
         guard let fetchedRecords = recordType.fetchedModelType?.syncRecords(recordJSON) else { return }
 
-        let ids = fetchedRecords.map { $0.objectId }.flatMap { $0 }
+        let ids = fetchedRecords.compactMap { $0.objectId }
         let localbookmarks = recordType.coredataModelType?.get(syncUUIDs: ids, context: DataController.shared.workerContext) as? [Bookmark]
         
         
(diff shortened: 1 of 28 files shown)


Data/sync/Sync.swift
```diff
@@ -449,7 +449,7 @@ extension Sync {
                     
                 // TODO: Needs favicon
                 if clientRecord == nil {
-                    recordType.coredataModelType?.add(rootObject: fetchedRoot, save: false, sendToSync: false, context: context)
+                    _ = recordType.coredataModelType?.add(rootObject: fetchedRoot, save: false, sendToSync: false, context: context)
```
jhreis: We should add `@discardableResult` to the `add` function here, so we don't need to explicitly ignore it (e.g. `_ = `

Client/Frontend/Browser/TabManager.swift
```diff
@@ -760,7 +760,7 @@ extension TabManager {
         }
         
         // Clean up any screenshots that are no longer associated with a tab.
-        imageStore?.clearExcluding(savedUUIDs)
+        _ = imageStore?.clearExcluding(savedUUIDs)
```
jhreis: Same, add `@discardableResult` to `clearExcluding`

jhreis (changes requested):
Wow, this is great, thanks for hammering a lot of these out! Just one small thing (in two places). I know you didn't mark this as upliftable, yet, but figured I'd give it a quick run through anyway 😄

Client/Frontend/Browser/TabManager.swift
cndouglas: Done.

Data/sync/Sync.swift
cndouglas: Done. Interestingly, the warning does not go away. Appears to be a bug with Xcode.

cndouglas:
@jhreis When you say "mark this as upliftable", do you mean add the `needsuplift` label to the issue? I can't add labels. 😉

jhreis (approved):

jhreis:
@Liunkae Totally forgot about that. We'll update the PR template on this, thanks for pointing that out 😄

jhreis:
First community PR for Brave on this repo 😄Congrats!! 🎉

PR #152, opened 2018-08-10
Update state of edit bookmarks button on bookmarks count change

shubham01:
## What?

- [x] The edit button in the bookmarks screen will be enabled only when there are bookmarks.
- [x] The editing mode will be switched off when all the bookmarks are deleted.

## Why?

The edit button was confusing me when the bookmarks were empty. So I thought we can just disable it when there are no bookmarks, for a minor UX improvement.

### DIFF
--- a/Client/Frontend/Menu/BookmarksViewController.swift
+++ b/Client/Frontend/Menu/BookmarksViewController.swift
@@ -182,6 +182,8 @@ class BookmarksViewController: SiteTableViewController, HomePanel {
       print(error.description)
     }
     
+    editBookmarksButton?.isEnabled = (self.bookmarksFRC?.fetchedObjects?.count ?? 0) > 0
+    
     super.reloadData()
   }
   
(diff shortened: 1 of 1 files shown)


Client/Frontend/Menu/BookmarksViewController.swift
```diff
@@ -182,6 +182,8 @@ class BookmarksViewController: SiteTableViewController, HomePanel {
       print(error.description)
     }
     
+    editBookmarksButton?.isEnabled = (self.bookmarksFRC?.fetchedObjects?.count ?? 0) > 0
```
jhreis, line 185: ```if self.bookmarksFRC?.fetchedObjects?.isEmpty == false``` is probably a bit cleaner.

jhreis (changes requested):
Thanks for submitting this PR! Love the fix, I think we should probably restructure it a bit. Controlling the button's state is so many places (currently 4) becomes heavily error prone with small modifications. Ideally we'd use something like KVO on the FRC's item count, however the `fetchedObjects` is a native Swift object and non-dynamic, so we can't observe this AFAIK.

I think we'd be better off creating a function that updates the edit button state. Something like `updateEditBookmarksButtonState`. Inside here we could put the necessary checks.

Then, rather than providing business logic throughout the VC those places just call the update function, which handles all logic internally.

jhreis:
Closing due to inactivity. Feel free to resubmit a new PR at some point. Thanks again for your interest!

PR #157, opened 2018-08-14
Fix #101, fix #102: core data unit tests.

### LINKED ISSUE #101: Unit Tests: CoreData
**CoreDataStackTests**

- [ ] testWorkerToMainMerge
- [ ] testMainToWorkerMerge
- [ ] testSaveContext

**CD Utilities:**

- [ ] compareContexts // will validate that both new workerContext, and mainContext have same data

### LINKED ISSUE #102: Unit Tests: Bookmarks
## Files:

**BookmarkTests**

- [ ] testAddFolder
- [ ] testAddBookmark
- [ ] testAddNestedBookmark
- [ ] testReordering
- [ ] testRemoveBookmark
- [ ] testRemoveFolder
- [ ] testRemoveFolderWithBookmarks
- [ ] testEditBookmark
- [ ] testEditFolder
- [ ] testEditTopLevelFolder

iccub:
- [x] `DataController` tests
- [x] `Bookmark` tests
- [x] `TabMO` tests
- [x] `Device` tests
- [x] `History` tests
- [x] `Domain` tests
- [x] `FaviconMO` tests

## Pull Request Checklist

## Screenshots


Data/models/DataController.swift
```diff
@@ -30,7 +30,7 @@ private let log = Logger.browserLogger
 // Follow the stack design from http://floriankugler.com/2013/04/02/the-concurrent-core-data-stack/
 
 public class DataController: NSObject {
-    public static let shared = DataController()
+    public static var shared: DataController? = DataController()
```
jhreis: What's the reason for making this an optional? Even if we allow it to be reset, we don't need to make it an optional. Just not a constant.

Data/models/DataController.swift
```diff
@@ -146,14 +153,24 @@ public class DataController: NSObject {
             }
         }
     }
-}
-
-extension NSManagedObjectContext {
-    static var mainThreadContext: NSManagedObjectContext {
-        return DataController.shared.mainThreadContext
+    
+    public static func resetDatabase() {
```
jhreis: This should probably be in a more unit test environment, putting unit test functions in general classes can get pretty messy.

Data/models/Device.swift
```diff
@@ -110,4 +110,13 @@ class Device: NSManagedObject, Syncable {
         }
     }
     
+    // Needed for testing only. No need to set it public, @testable annotation allows use of internal methods.
+    class func clearSharedDevice() {
```
jhreis: Same, why does this need to be here instead of in the unit test target?

jhreis (changes requested):
Coverage lgtm, just some more general stuff in the normal codebase.

Data/models/DataController.swift
iccub: I would need to make initializer at least `internal` then. If you prefer it this way I can change it.

Data/models/Device.swift
iccub: Same goes here, I need to drop `private` access for `shared`

jhreis (approved):

PR #193, opened 2018-08-24
Fix #192: Remove unneeded deployment config

### LINKED ISSUE #192: Remove unneeded deployment target config
Warning:

The UnitTests target has a custom deployment target config:

The target-level config should be removed so that the target inherits the project-level config.

cndouglas:
## Pull Request Checklist

Fix #192

## Screenshots
Here's what the config looks like after this change:


kylehickinson (approved):
We should make sure for those of us using Xcode 10 (I believe this might just be @iccub and myself), that we remove the added iOS 12 deployment target that it sets by default when making a new target

PR #306, opened 2018-10-03
iOS 10 CD model support for back-porting.

jhreis:


iccub (approved):
The app works for me and screenshots are restoring correctly.

PR #324, opened 2018-10-10
Fix README: added hint for installing swiftlint first

itinance:
This PR is adding a note in README that `swiftlint` should be installed before `bootstrap.sh` can be executed. 
Currently it is not mentioned, but setting up this project for development for the very first time, will end up like this:

```
$ sh ./bootstrap.sh
Brave requires the command: swiftlint
Please install it via Homebrew or directly from https://github.com/realm/SwiftLint/releases
```

Therefor i have added a specific lint, that swiftlint needs to be installed first.

## Pull Request Checklist

No testing required as no code was added, removed or modified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Getting involved
 
 We encourage you to participate in this open source project. We love Pull Requests, Bug Reports, ideas, (security) code reviews or any kind of positive contribution.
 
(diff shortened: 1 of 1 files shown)


jhreis (approved):
Sweet thanks!

Issue #37, opened 2018-05-16
Sync Integration


jamesmudgett:

tmancey:
Blocked until feature is complete on 1.6

iccub:
Unblocking, we decided to skip 1.6 integration and port it directly to 1.7

PR #386, opened 2018-11-07
Fix: #37, sync integration

iccub:
Most of the code is a direct 1.6 port, with few small improvements.

Notes:
- I moved `brave/sync` dependency from carthage to npm. It doesn't make sense to place it in Carthage, and it also makes setup script a bit cleaner. See `bootstrap.sh` and `package.json` for changes. Also `brave/crypto` dep is removed, because `brave/sync` relies on it already, no need to duplicate it
- I created a separate folder for soft-linked sync dependencies. Also added a small readme for it.
- Sync callout in Bookmark panel was not added. I will add it in another commit if we want to keep it.
- We continue to utilize `NO_SYNC` flag. It was not added to the project so sync is enabled by default.


iccub:
Changing to `[WIP]` because tests needs to be updated, but all the main code is ready for review.

iccub:
Tests are fixed, the PR should be ready to go.

It also fixes issues with sorting bookmarks when there is over 10 of them.

Client/Frontend/Browser/HomePanel/favorites/FavoritesDataSource.swift
```diff
@@ -76,7 +76,8 @@ class FavoritesDataSource: NSObject, UICollectionViewDataSource {
             return UICollectionViewCell()
         }
         
-        let ftd = FavoritesTileDecorator(url: url, cell: cell, indexPath: indexPath, color: fav.color != nil ? UIColor(colorString: fav.color!) : nil)
+        // BRAVE TODO: Update Favorites logic, color argument is outdated here. Setting nil for now.b
```
jhreis: Is there at ticket to track this?
Also, remove extraneous "b"

Client/Frontend/Settings/SettingsViewController.swift
```diff
@@ -156,6 +165,33 @@ class SettingsViewController: TableViewController {
...
+                        
+                        self.navigationController?.pushViewController(syncSettingsVC, animated: true)
+                    } else {
+                        let view = SyncWelcomeViewController()
+                        view.dismissHandler = {
+                            view.navigationController?.popToRootViewController(animated: true)
+                        }
+                        self.navigationController?.pushViewController(view, animated: true)
+                    }
+                    }, accessory: .disclosureIndicator)
```
jhreis: Fix indentation.

jhreis (changes requested):

Client/Frontend/Browser/HomePanel/favorites/FavoritesDataSource.swift
iccub: This was fixed in https://github.com/brave/brave-ios/pull/383, I only need to merge latest development changes to this rbanch

iccub:
Fixed a bug with sending empty bookmarks to sync server and added missing strings for Sync(that were removed in the latest localization cleanup)

BraveShared/BraveStrings.swift
```diff
@@ -297,6 +297,58 @@ public extension Strings {
     public static let ErrorPagesVisitOnceButton = NSLocalizedString("ErrorPagesVisitOnceButton", tableName: "BraveShared", value: "Visit site anyway", comment: "Button label to temporarily continue to the site from the certificate error page")
 }
 
+// MARK: - Sync
+public extension Strings {
```
jhreis, line 301: Please add `value` to all of these strings. For now, the key can just be the variable name. For example:

public static let QRCode = NSLocalizedString("QRCode, value: "QR Code", comment: "QR Code section title")

Data/sync/Sync.swift
```diff
@@ -64,101 +66,138 @@ enum SyncActions: Int {
...
+    
     // TODO: Move to a better place
     fileprivate let prefNameId = "device-id-js-array"
     fileprivate let prefNameName = "sync-device-name"
     fileprivate let prefNameSeed = "seed-js-array"
     fileprivate let prefFetchTimestamp = "sync-fetch-timestamp"
+    fileprivate let prefBaseOrder = "sync-base-order"
     
-//    #if DEBUG
+    //    #if DEBUG
```
jhreis: Make these variables `lazy var` and return based on the build type.

jhreis (changes requested):
Great work. 

Please also file a ticket for converting our `lazy var` usage to using `then`. Do not do it now, just file away so we can track it.

Data/sync/Sync.swift
```diff
@@ -64,101 +66,137 @@ enum SyncActions: Int {
...
     fileprivate let prefNameId = "device-id-js-array"
     fileprivate let prefNameName = "sync-device-name"
     fileprivate let prefNameSeed = "seed-js-array"
     fileprivate let prefFetchTimestamp = "sync-fetch-timestamp"
+    fileprivate let prefBaseOrder = "sync-base-order"
+    
+    fileprivate lazy var isDebug: Bool = { return AppConstants.BuildChannel == .developer }()
+    
+    fileprivate lazy var serverUrl: String = {
+        return AppConstants.BuildChannel == .developer ?
```
jhreis: Use the above `isDebug`, that way logic is centralized:
`return isDebug ? "" : ""`

Data/sync/Sync.swift
```diff
@@ -64,101 +66,137 @@ enum SyncActions: Int {
...
+    }
+    
     // TODO: Move to a better place
     fileprivate let prefNameId = "device-id-js-array"
     fileprivate let prefNameName = "sync-device-name"
     fileprivate let prefNameSeed = "seed-js-array"
     fileprivate let prefFetchTimestamp = "sync-fetch-timestamp"
+    fileprivate let prefBaseOrder = "sync-base-order"
+    
+    fileprivate lazy var isDebug: Bool = { return AppConstants.BuildChannel == .developer }()
```
jhreis: Switch to `isRelease`, so betas also are treated as `debug=true`

jhreis (approved):

Data/sync/Sync.swift
```diff
@@ -64,101 +66,136 @@ enum SyncActions: Int {
...
+    }
+    
     // TODO: Move to a better place
     fileprivate let prefNameId = "device-id-js-array"
     fileprivate let prefNameName = "sync-device-name"
     fileprivate let prefNameSeed = "seed-js-array"
     fileprivate let prefFetchTimestamp = "sync-fetch-timestamp"
+    fileprivate let prefBaseOrder = "sync-base-order"
+    
+    fileprivate lazy var isDebug: Bool = { return AppConstants.BuildChannel != .release }()
```
jhreis:
```suggestion
    fileprivate lazy var isDebug: Bool = { return !AppConstants.BuildChannel.isRelease }()
```

jhreis (approved):

PR #414, opened 2018-11-16
Fix #462, add Brave.sqlite to new core data stack for 1.6 migration.

### LINKED ISSUE #462: Upgrade from 1.6 doesn't preserve bookmarks, history etc.
A database under different name is used on 1.6 and after upgrade to 1.7 all this data is lost.

iccub:
This is port of https://github.com/brave/browser-ios/pull/1805 PR. It does not need to be merged until 1.6->1.7 upgrade tests

## Pull Request Checklist

## Screenshots

### DIFF
--- a/Data/models/DataController.swift
+++ b/Data/models/DataController.swift
@@ -8,6 +8,8 @@ import XCGLogger
 private let log = Logger.browserLogger
 
 public class DataController: NSObject {
+    private let databaseName = "Brave.sqlite"
+    
     public static var shared: DataController = DataController()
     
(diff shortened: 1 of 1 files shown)


iccub:
I made a sample test today.
Installed 1.6.6, added few bookmarks, then installed 1.7 under same bundle id.
And looks like it works
Bookmarks, history and opened tabs were preserved between launches. 
Even light migration to `model8` didn't mess up anything.

jhreis (approved):

PR #436, opened 2018-11-23
Fix #422, malware/phishing protection.

### LINKED ISSUE #422: Phishing/Malware Protection
In browser-ios, the adblock library is used to flag phishing or malware domains. In brave-ios we will be building this manually (for now at least). There is a combination of two files that are used for this:

Spam404 Domain Blacklist:
https://raw.githubusercontent.com/Dawsey21/Lists/master/adblock-list.txt
Select only the rows with || before them and ^ at the end of them.
Ignore the rows with ## in them (anywhere) and ! before them.

Disconnect Simple Malware:
https://s3.amazonaws.com/lists.disconnect.me/simple_malware.txt

Using static lists for initial implementation is fine, and we'll follow up with creating dynamic update mechanisms for these files in some form. Using some sort of hash list will be optimal to avoid high levels of iterations. Lists may need to be modified, but should be done through some script, as opposed to manually. 

Having a script allows easier updating in the future.

Other points to research:
- What to do when one of these is hit.

iccub:
## Pull Request Checklist

## Screenshots


Client/WebFilters/SafeBrowsing/Lists/MalwareList.swift
```diff
@@ -0,0 +1,43 @@
...
+    /// by using the regex syntax.
+    let regex: NSRegularExpression?
+    
+    static func generate() -> [MalwareList] {
+        var lists = [MalwareList]()
+        
+        // Adblock Plus scam domains
+        
+        // Adblock's domains come in format ||domain.com^.
+        let adblockListRegex = "(?<=\\|\\|).+?(?=\\^)"
```
jhreis, line 27: Are we dealing with the "survey scams" that start with `##`?

Shared/Extensions/URLExtensions.swift
```diff
@@ -361,7 +361,14 @@ extension URL {
 extension URL {
     public var isErrorPageURL: Bool {
         if let host = self.host {
-            return self.scheme == "http" && host == "localhost" && path == "/errors/error.html"
+            return self.scheme == "http" && host == "localhost" && path.contains("/errors/")
```
jhreis: Can't this entire function just be:
`return `return self.scheme == "http" && self.host == "localhost" && path.contains("/errors/")`

Since this would account for `host` being `nil` (as in not equal to "localhost")?

Shared/Extensions/URLExtensions.swift
```diff
@@ -361,7 +361,14 @@ extension URL {
...
+        }
+        return false
+    }
+    
+    public var safeBrowsingErrorURL: Bool {
+        if let host = self.host {
+            return self.scheme == "http" && host == "localhost" && path.contains("/errors/SafeBrowsingError.html")
```
jhreis: Same?

jhreis (changes requested):

Client/WebFilters/SafeBrowsing/Lists/MalwareList.swift
iccub: In https://github.com/brave/brave-ios/issues/422 you say to ignore all other filters, if I understood it correctly

jhreis: My bad, you're right, slipped my mind. Thanks.

Shared/Extensions/URLExtensions.swift
```diff
@@ -360,10 +360,11 @@ extension URL {
...
 extension URL {
     public var isErrorPageURL: Bool {
-        if let host = self.host {
-            return self.scheme == "http" && host == "localhost" && path == "/errors/error.html"
-        }
-        return false
+        return scheme == "http" && host == "localhost" && path.contains("/errors/")
+    }
+    
+    public var safeBrowsingErrorURL: Bool {
```
jhreis, line 366: Isn't this exact same as `isErrorPageURL`? If so, maybe remove, or else maybe just return `isErrorPageURL`?

iccub: No there is `errors/error.html` for network and certificate errors and `errors/safeBrowsingError.html` for phishing errors only.

We need to check for `safeBrowsingError` separately because we want to enable Brave shields for this page and we don't want to do that for regular error pages.

I could do some refactor or add comments to make it more clear

And `isErrorPageURL` helper applies to both `error` and `safeBrowsingError`

jhreis: But the logic is literally identical:
`return scheme == "http" && host == "localhost" && path.contains("/errors/")`
Is something missing in the `path`? Maybe a copy/paste mistake?

iccub: Yes I copied it and forgot to change, sorry, good catch

jhreis: Got it, so to confirm, `isErrorPageURL` is very generic, the behavior _has_ changed, given the diff (from `/error/error.html` to just `/error/`).

While `safeBrowsingErrorURL` is very specific, right?

iccub: Yes, before there was only `error.html` type of error. I decided to add safe browsing error to the same `/errors/` module because they behave similar in most cases.

`isErrorPageURL` detects network errors and phishing warning html
`safeBrowsingErrorURL` detects only phishing warning html

jhreis (approved):

PR #444, opened 2018-11-26
Fix #432: match DAU behavior to Android's.

### LINKED ISSUE #432: Change DAU Ping Behavior
DAU Ping behavior should align with what Android does:

- No delay on launch
- Setup 5 minute timer that attempts a re-ping. Timer should never intentionally stop.
- Attempt additional ping on every app open (just not launch)

iccub:
- try to send ping each time the app goes to foreground
- try to send ping every 20 minutes while the app is running

## Pull Request Checklist

## Screenshots


BraveShared/Analytics/DAU.swift
```diff
@@ -38,14 +38,19 @@ public class DAU {
...
         if launchTimer != nil { return false }
         launchTimer =
             Timer.scheduledTimer(
-                timeInterval: activeUserDuration,
+                timeInterval: pingRefreshDuration,
                 target: self,
                 selector: #selector(sendPingToServerInternal),
                 userInfo: nil,
-                repeats: false)
+                repeats: true)
```
jhreis, line 53: We should kill this at some point. I think maybe a more flexible approach would be to remove the timer entirely. If the DAU call fails, then we do an `asyncAfter` scheduled for `20.minutes`?

BraveShared/TimeExtensions.swift
```diff
@@ -0,0 +1,23 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+
+extension Int {
+    
+    /// Return number of minutes in seconds.
+    var minutes: TimeInterval {
```
jhreis, line 10: ❤️

jhreis (changes requested):

BraveShared/Analytics/DAU.swift
iccub: Android does it every 20 minutes though, not just once after 20mins. If app is not used for some the timer will be killed anyway. 
I can convert it to launch-only-once if you think it's better

jhreis: It would just be recursive.

- attempt ping
- ping fails, schedule new ping in 20 minutes
- 20 minutes pass
- new ping attempt
- ping fails, schedule new ping in 20 minutes

This way if it is _successful_ it will just automatically stop ping attempts. Or is it supposed to attempt every 20 minutes forever, even if successful? I suppose this would catch situations where users use the app at 11:30pm and are still using it at 12:10am.

jhreis: Android leaves it, so we will too.

jhreis (changes requested):
One thing we still need to fix. We currently update the last ping data before we attempt the ping. If the ping fails, this data will be set incorrectly. We should only set the last ping data if a ping was successful.

BraveShared/Analytics/DAU.swift
```diff
@@ -76,13 +81,26 @@ public class DAU {
             // Ping was successful, next ping should be sent with `first` parameter set to false.
             // This preference is set for future DAU pings.
             Preferences.DAU.firstPingParam.value = false
+            
+            // This preference is used to calculate wheter user used the app in this month and/or day.
```
jhreis: `wheter` spelling.

BraveShared/Analytics/DAU.swift
```diff
@@ -15,7 +15,7 @@ public class DAU {
     private static let baseUrl = "https://laptop-updates.brave.com/\(apiVersion)/usage/ios?platform=ios"
     
     /// Number of seconds that determins when a user is "active"
-    private let activeUserDuration = 10.0
+    private let pingRefreshDuration = 20.minutes
```
jhreis: We should change this to `5.minutes` now.
```suggestion
    private let pingRefreshDuration = 5.minutes
```

BraveShared/Analytics/DAU.swift
```diff
@@ -38,27 +38,32 @@ public class DAU {
             return false
         }
         
+        // Sending ping immediately
+        sendPingToServerInternal()
```
jhreis, line 42: To be insanely safe, I think we should maybe perform a ping on app open too... Feels kind of redundant with a 5 minute timer, but this is technically what Android does.

jhreis (approved):

BraveShared/Analytics/DAU.swift
jhreis: https://github.com/brave/brave-ios/pull/444/files#diff-19fd1a6fe4b2f78eb457726f366733e0R285
See it here, now, thx

PR #454, opened 2018-11-27
Sync improvements

### LINKED ISSUE #447: Nested bookmarks are not removed from sync
### Steps to Reproduce 
  1. Create a folder and add few bookmarks there.
  2. Remove the folder
  3. Join another device to sync group

**Actual result:** 
Bookmarks from the previously deleted folder are shown on the new device

iccub:
This PR fixes a lot of problems with sync.
- threading issues in `get-existing-objects`
- sending updated records(did not work at all previously)
- adding `syncTimestamp` to Bookmark
- removing nested bookmarks
- adding nested bookmarks from sync server in correct order
- deleting a device from sync seems to work more reliable too
- leave sync group when user initialized sync for the first time and `sync-setup-error` happens

## Pull Request Checklist

## Screenshots

### DIFF
--- a/Data/models/Bookmark.swift
+++ b/Data/models/Bookmark.swift
@@ -360,6 +360,40 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
         
         return self.add(rootObject: bookmark, save: true)
     }
+    
+    public func remove(sendToSync: Bool = true) {
+        if isFavorite { delete() }
+        
+        // Before we delete a folder and its children, we need to grab all children bookmarks
+        // and send them to sync with `delete` action.
+        if isFolder && sendToSync {
+            removeFolderAndSendSyncRecords(uuid: syncUUID)
+            return
+        }
+        
+        if sendToSync {
+            Sync.shared.sendSyncRecords(action: .delete, records: [self])
+        }
+        
+        delete()
+    }
+    
+    private func removeFolderAndSendSyncRecords(uuid: [Int]?) {
+        if !isFolder { return }
+        
+        var allBookmarks = [Bookmark]()
+        allBookmarks.append(self)
+        
+        if let allNestedBookmarks = Bookmark.getRecursiveChildren(forFolderUUID: syncUUID) {
+            log.warning("All nested bookmarks of :\(String(describing: title)) folder is nil")
+            
+            allBookmarks.append(contentsOf: allNestedBookmarks)
+        }
+        
+        Sync.shared.sendSyncRecords(action: .delete, records: allBookmarks)
+        
+        delete()
+    }
 }
 
 // TODO: Document well
@@ -423,6 +457,34 @@ extension Bookmark {
         
         let record = first(where: predicate, context: context)
         record?.delete()
+    }
+    
+    /// Gets all nested bookmarks recursively.
+    public static func getRecursiveChildren(forFolderUUID syncUUID: [Int]?,
+                                            context: NSManagedObjectContext = DataController.viewContext) -> [Bookmark]? {
+        guard let searchableUUID = SyncHelpers.syncDisplay(fromUUID: syncUUID) else {
+            return nil
+        }
+        
+        let syncParentDisplayUUIDKeyPath = #keyPath(Bookmark.syncParentDisplayUUID)
+        
+        let predicate = NSPredicate(format: "\(syncParentDisplayUUIDKeyPath) == %@", searchableUUID)
+        
+        var allBookmarks = [Bookmark]()
+        
+        let result = all(where: predicate, context: context)
+        
+        result?.forEach {
+            allBookmarks.append($0)
+            
+            if $0.isFolder {
+                if let nestedBookmarks = getRecursiveChildren(forFolderUUID: $0.syncUUID) {
+                    allBookmarks.append(contentsOf: nestedBookmarks)
+                }
+            }
+        }
+        
+        return allBookmarks
     }
     
     public class func frecencyQuery(context: NSManagedObjectContext, containing: String?) -> [Bookmark] {
--- a/Data/models/Bookmark.swift
+++ b/Data/models/Bookmark.swift
@@ -67,7 +67,6 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
     
     override public func awakeFromInsert() {
         super.awakeFromInsert()
-        created = Date()
         lastVisited = created
     }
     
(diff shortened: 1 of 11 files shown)


iccub:
This PR also fixes some issues with Bookmarks not getting deleted on a device. Removing/updating bookmarks should now work better in general.

Data/models/Bookmark.swift
```diff
@@ -425,6 +426,55 @@ extension Bookmark {
         record?.delete()
     }
     
+    /// Gets all nested bookmarks recursively.
+    public static func getAllChildren(forFolderUUID syncUUID: [Int]?,
```
jhreis: This looks pretty similar to `getChildren`, can we centralize logic any way?

Data/sync/Sync.swift
```diff
@@ -459,12 +463,20 @@ extension Sync {
...
-        if recordType == .prefs, let data = fetchedRecords as? [SyncDevice] {
+        if recordType == .prefs {
             // Devices have really bad data filtering, so need to manually process more of it
             // Sort to not rely on API - Reverse sort, so unique pulls the `latest` not just the `first`
-            fetchedRecords = data.sorted { device1, device2 in
+            fetchedRecords = fetchedRecords.sorted { device1, device2 in
                 device1.syncTimestamp ?? -1 > device2.syncTimestamp ?? -1 }.unique { $0.objectId ?? [] == $1.objectId ?? [] }
             
+        } else if recordType == .bookmark {
+            // Bookmars are sorted to have the oldest bookmark to come in first, so it can be added before it's children.
```
jhreis: How is this changing current behavior?

jhreis (changes requested):
This is a great PR!

Can you please rewrite the commit without a ticket to include the prefix: `No Bug: `?

lmk if you need help rewriting an old commit message.

Data/sync/Sync.swift
iccub: Currently bookmarks from the server are received in random order I believe, but if we want to add them properly we have to sort it.
Otherwise nested bookmark could go in before its parent which resulted in children bookmarks being outside of their parent folder.

Data/models/Bookmark.swift
iccub: Not easy to centralize it without nesting logic, for `getAllChildren` we never want to ignore folders while `getChildren` has this option. And recursive search never wants to skip folders.

I think of renaming it to `getAllInsideOf(folderUUID syncUUID..`

jhreis: Oh this is recursive? If so, should be renamed, I think `getAllInsideOf` is probably better, helps. Or `getRecursiveChildren` or something.

Data/sync/Sync.swift
jhreis: I see, this should be handled when a folder is added though, when a folder is added we already attempt to find all orphaned children.
https://github.com/brave/brave-ios/pull/454/files#diff-01f57e7d7609dba0eec131782b05f57dR210

iccub: But if we first save children and parent next, the children will stay at root level. It doesn't remove the Bookmark and move it into a correct place.

jhreis: Alright, we'll take care of this separately.

Data/models/Bookmark.swift
```diff
@@ -425,6 +426,55 @@ extension Bookmark {
...
+    /// Gets all nested bookmarks recursively.
+    public static func getRecursiveChildren(forFolderUUID syncUUID: [Int]?,
+                                            context: NSManagedObjectContext = DataController.viewContext) -> [Bookmark]? {
+        guard let searchableUUID = SyncHelpers.syncDisplay(fromUUID: syncUUID) else {
+            return nil
+        }
+        
+        let syncParentDisplayUUIDKeyPath = #keyPath(Bookmark.syncParentDisplayUUID)
+        
+        let predicate = NSPredicate(format: "\(syncParentDisplayUUIDKeyPath) == %@", searchableUUID)
```
jhreis, line 472: Why use `\(foo)` for one and `%@` for the other?

Data/models/Bookmark.swift
```diff
@@ -425,6 +426,55 @@ extension Bookmark {
...
+                if let nestedBookmarks = getRecursiveChildren(forFolderUUID: $0.syncUUID) {
+                    allBookmarks.append(contentsOf: nestedBookmarks)
+                }
+            }
+        }
+        
+        return allBookmarks
+    }
+    
+    public func removeFolderAndSendSyncRecords(uuid: [Int]?) {
```
jhreis: Is this needed? Can't the normal `remove` function take care of this? Rather than identifying which function to call?

Data/models/Bookmark.swift
```diff
@@ -425,6 +426,55 @@ extension Bookmark {
...
+        
+        if let allNestedBookmarks = Bookmark.getRecursiveChildren(forFolderUUID: syncUUID) {
+            log.warning("All nested bookmarks of :\(String(describing: title)) folder is nil")
+            
+            allBookmarks.append(contentsOf: allNestedBookmarks)
+        }
+        
+        Sync.shared.sendSyncRecords(action: .delete, records: allBookmarks)
+        
+        delete()
```
jhreis: We sure that all children will get deleted from CD too this way?

Data/sync/Sync.swift
jhreis: `Bookmars` typo.

jhreis (changes requested):

Data/models/Bookmark.swift
iccub: Normal `remove` drops folder and children all at once, while for sync, each record to delete must be passed separately.

Data/models/Bookmark.swift
iccub: Because `\(foo)` is static, and the second one is dynamic perhaps?
This convention is widely used in apple docs https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CoreData/FetchingObjects.html
minus the `#keypath` trick.

Data/models/Bookmark.swift
iccub: Yes, I verified it with core data

Data/models/Bookmark.swift
iccub: So after you drop folder, we have no children bookmark references to send them to sync's remove

jhreis: Yeah, I just mean, maybe the `remove` function should be smarter.

jhreis (approved):

PR #474, opened 2018-11-28
Fix #467, dau first ping param conversion.

### LINKED ISSUE #467: DAU ping after upgrade is treated as first ping
### Steps to Reproduce 
  1. Install brave1.6.6 launch the app, wait few second so DAU ping happens
  2. Upgrade to 1.7
  3. Launch and run the app for 15 seconds so the DAU ping is sent

Second DAU ping on 1.7 is sent with `first=true` parameter set.

On 1.7 we introduced new preference `firstPingParam` which needs to be set up during migration process

iccub:
On old codebase we checked existence of `dau_stat` to determine whether it's first server ping.
We need to translate that to use the new `firstPingParam` preference.

## Pull Request Checklist

## Screenshots

### DIFF
--- a/BraveShared/Preferences.swift
+++ b/BraveShared/Preferences.swift
@@ -29,16 +29,10 @@ extension Preferences {
     final class DAU {
         static let lastLaunchInfo = Option<[Int?]?>(key: "dau.last-launch-info", default: nil)
         static let weekOfInstallation = Option<String?>(key: "dau.week-of-installation", default: nil)
-        static let firstPingParam: Option<Bool> = {
-            var defaultValue = true
-            // On old codebase we checked existence of `dau_stat` to determine whether it's first server ping.
-            // We need to translate that to use the new `firstPingParam` preference.
-            if Preferences.DAU.lastLaunchInfo.value != nil {
-                defaultValue = false
-            }
-            
-            return Option<Bool>(key: "dau.first-ping", default: defaultValue)
-        }()
+        // On old codebase we checked existence of `dau_stat` to determine whether it's first server ping.
+        // We need to translate that to use the new `firstPingParam` preference.
+        static let firstPingParam: Option<Bool> =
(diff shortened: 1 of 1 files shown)


BraveShared/Preferences.swift
```diff
@@ -147,6 +147,11 @@ extension Preferences {
     public class func migrateBraveShared(keyPrefix: String) {
         // DAU
         migrate(keyPrefix: keyPrefix, key: "dau_stat", to: Preferences.DAU.lastLaunchInfo)
+        // On old codebase we checked existence of `dau_stat` to determine whether it's first server ping.
+        // We need to translate that to use the new `firstPingParam` preference.
+        if Preferences.DAU.lastLaunchInfo.value != nil {
```
jhreis: Just place this logic when setting the `default` on `firstPingParam`:

```
static let firstPingParam = Option<Bool>(key: "dau.first-ping", default: Preferences.DAU.lastLaunchInfo.value == nil)
```

jhreis (changes requested):

BraveShared/Preferences.swift
```diff
@@ -29,7 +29,17 @@ extension Preferences {
     final class DAU {
         static let lastLaunchInfo = Option<[Int?]?>(key: "dau.last-launch-info", default: nil)
         static let weekOfInstallation = Option<String?>(key: "dau.week-of-installation", default: nil)
-        static let firstPingParam = Option<Bool>(key: "dau.first-ping", default: true)
+        static let firstPingParam: Option<Bool> = {
```
jhreis: We can literally just set the `defaultValue` inline:

(key: "", default: Preferences.DAU.lastLaunchInfo.value == nil`)

This then would set `default` to the correct value.

jhreis (changes requested):

jhreis (approved):

PR #559, opened 2018-12-04
Fix #243: Added translations from some translated languages

### LINKED ISSUE #243: Pull finalized localized strings from Transifex

danishjafri88:
Some translations are missing, not yet translated.

## Pull Request Checklist

## Screenshots


danishjafri88:
The pull request is huge. Could not do anything from keeping it small. Please give it a test run on multiple languages to verify.

Client/Frontend/Strings.swift
```diff
@@ -17,63 +17,63 @@ private func applicationBundle() -> Bundle {
 }
 
 extension Strings {
-    public static let OKString = NSLocalizedString("OKString", value: "OK", comment: "OK button")
-    public static let ThirdPartySearchFailedTitle = NSLocalizedString("ThirdPartySearchFailedTitle", value: "Failed", comment: "A title explaining that we failed to add a search engine")
+    public static let OKString = NSLocalizedString("OKString", bundle: Bundle.shared, value: "OK", comment: "OK button")
```
jhreis, line 20: Why are we bundling all of these? They aren't technically in the `shared` bundle either.

danishjafri88: They are in the shared bundle. Check the target membership of the file.

jhreis: Ah got it, they in the wrong place. Alright, fine by me.

iccub (changes requested):
Edit: last commit fixed the issue for me!

App name is bad on non english regions. I verified it doesn't happen on development branch.

zrzut ekranu 2018-12-4 o 15 33 00

BraveShareTo/it.lproj/InfoPlist.strings
```diff
@@ -0,0 +1,6 @@
+/* Bundle display name */
+"CFBundleDisplayName" = "Apri con Brave";
+
+/* Bundle name */
+"CFBundleName" = "$(PRODUCT_NAME)";
```
jhreis: We should remove all these bundle names.

jhreis (changes requested):

BraveShareTo/it.lproj/InfoPlist.strings
danishjafri88: Ah forgot these, removed

BraveShareTo/zh.lproj/InfoPlist.strings
```diff
@@ -0,0 +1,6 @@
+/* Bundle display name */
+"CFBundleDisplayName" = "在 Brave 中打开";
+
+/* Bundle name */
+"CFBundleName" = "$（PRODUCT_NAME）";
```
jhreis: Remove

BraveShared/zh.lproj/InfoPlist.strings
```diff
@@ -0,0 +1,3 @@
+/* Bundle name */
+"CFBundleName" = "$（PRODUCT_NAME）";
```
jhreis: Remove

jhreis (changes requested):

jhreis (approved):

iccub (approved):

PR #603, opened 2018-12-07
Fix #510: Fix punycode decoder and disable smart dashes in URL bar

### LINKED ISSUE #510: NULL deref in updateTextWithURL - TabLocationView.swift
### Description: 
NULL deref while opening a punycode URL

### Steps to Reproduce 
  1. Open http://xn--eby-7cd.com/

**Actual result:** 
Crash

**Expected result:**
No crash

Always

**Brave Version:** 
Dev (commit - 60a82d37c7987162971f1d23bf5a8639131a27b5)

**Device details:** 
Simulator (iOS 12.1 - iPhone XS)

cndouglas:
Fixes #510

See https://github.com/brave/brave-ios/issues/510#issuecomment-445312822 for an explanation of the problem.

Explanation of changes:

- **Punycode.swift line 126**: find the last hyphen (the delimiter) instead of the first hyphen, in case the URL has other hyphens unrelated to punycode syntax
- **Punycode.swift line 191**: avoid using magic numbers in code: use the length of the punycode prefix instead
- **URLBarView.swift line line 361**: disable smart dashes for the URL field to prevent `--` from being automatically converted to `—`.

## Pull Request Checklist

## Screenshots

Smart dashes are now disabled in the URL bar (compare to https://github.com/brave/brave-ios/issues/510#issuecomment-445312822):

There are two problems that each have separate verification steps.

Handling hyphens in punycode URL:
1. Open Brave and tap the URL bar.
2. Type `test-😄.com` and tap Go.
3. The app should not crash and should show an error page (the URL is not valid).

Handling the punycode prefix:
1. Open Brave and tap the URL bar.
2. Type `xn--`.
3. The text should remain `xn--` and not change to `xn—` (like the screen capture above).

### DIFF
--- a/Client/Frontend/Browser/Punycode.swift
+++ b/Client/Frontend/Browser/Punycode.swift
@@ -123,7 +123,7 @@ extension String {
         var n = initialN
         var bias = initialBias
         var pos = 0
-        if let ipos = input.index(of: delimiter) {
+        if let ipos = input.lastIndex(of: delimiter) {
             pos = ipos
             output.append(contentsOf: input[0 ..< pos])
             pos += 1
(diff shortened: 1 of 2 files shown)


kylehickinson (approved):
Nice!! Probably going to wait until post 1.7 submission to merge this

PR #613, opened 2018-12-10
Show user a popup after Sync being disabled.

iccub:
## Pull Request Checklist

## Screenshots

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -415,7 +415,7 @@ class BrowserViewController: UIViewController {
         let sync = Sync.shared
         
         #if NO_SYNC
-        if sync.syncSeed == nil { return }
+        if sync.syncSeedArray == nil { return }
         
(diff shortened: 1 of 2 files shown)


jhreis (approved):

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -412,7 +412,24 @@ class BrowserViewController: UIViewController {
     /// Added to keyWindow, since it can then be utilized from any VC (e.g. settings modal).
     /// This also inits sync at app launch.
     private func initializeSyncWebView() {
-        Sync.shared.webView.alpha = 0.01
+        let sync = Sync.shared
+        
+        #if NO_SYNC
+        if sync.syncSeed == nil { return }
```
jhreis: Use `syncSeedArray` instead, as that is already `public` and also `readonly`. Don't want to expose `syncSeed` writtable.

jhreis (changes requested):

jhreis (approved):

PR #667, opened 2019-01-02
Fix #696: Enable adding multiple bookmarks of the same url.

### LINKED ISSUE #696: Decide and implement how to deal with duplicated Bookmarks.
### Steps to Reproduce 
  1. Join two devices into a sync group.
  2. Add Wikipedia bookmark on device A
  3. Add the same url Wikipedia bookmark on device B

**Actual result:** 
We need to decide how to handle this case, currently it is not handled at all

**Expected result:**

**Brave Version:** 

**Device details:** 

**Website problems only:**

iccub:
This is a prerequisite for Sync, as we need to have an option for having multiple bookmarks with the same url.

A user still can remove bookmarks either by swiping on them or by tapping on edit button.

## Pull Request Checklist

## Screenshots


Client/Frontend/Menu/HomeMenuController.swift
```diff
@@ -176,11 +175,7 @@ class HomeMenuController: UIViewController, PopoverContentComponent {
     //switch to bookmarks 'tab' in case we're looking at history and tapped the add/remove bookmark button
     onClickPageButton(bookmarksButton)
     
-    if Bookmark.contains(url: url) {
-      Bookmark.remove(forUrl: url)
-    } else {
-      Bookmark.add(url: url, title: tabState.title, parentFolder: bookmarksController.currentBookmarksPanel().currentFolder)
-    }
+    Bookmark.add(url: url, title: tabState.title, parentFolder: bookmarksController.currentBookmarksPanel().currentFolder)
```
jhreis, line 178: I don't think this is still correct, the toggle should function as normal, it is just that the models should allow duplicate bookmarks to come from sync.

Or if the user edits two bookmarks to the same URL, that should be allowed (which I believe it currently is), but if the user hits the toggle while having a matching url bookmark it should be removed.

Other platforms currently delete a random bookmark if there are > 1 that match the given URL, and we have been told to follow suite on this too.

jhreis (changes requested):

Client/Frontend/Menu/HomeMenuController.swift
iccub: So no code changes are needed. Removing bookmark by urls does just this, removes a random bookmark of a given url. 
This PR can be closed then

PR #671, opened 2019-01-03
Fix #409, don't save Domain on view context.

### LINKED ISSUE #409: Refactor `getOrCreateForUrl` for Background Writes
Currently when fetching a URL, if one does not exist it is created on the requesting `context`.

This is potentially problematic when the requesting context is `viewContext`, as save operations will happen on it.

A solution where a `newBackgroundContext()` is used to handle the creation of a new `Domain` should be investigated and considered. 

Current code as of this ticket creation:
```
        var newDomain: Domain!
        context.performAndWait {
            newDomain = Domain(entity: Domain.entity(context), insertInto: context)
            newDomain.url = domainString
            DataController.save(context: context)
        }
```

New Pseudo code:

```
        let writeContext = DataController.newBackgroundContext()
        var newDomain: Domain!
        writeContext.performAndWait {
            newDomain = Domain(entity: Domain.entity(writeContext), insertInto: writeContext)
            newDomain.url = domainString
            DataController.save(context: writeContext).upon {
                 newDomain = context.getObjectById(newDomain.objectId)
            }
        }
```

iccub:
## Pull Request Checklist

## Screenshots

### DIFF
--- a/Data/models/Domain.swift
+++ b/Data/models/Domain.swift
@@ -38,19 +38,28 @@ public final class Domain: NSManagedObject, CRUD {
             return domain
         }
         
-        // See #409:
-        //  A larger refactor is probably wanted here.
-        //  This can easily lead to a Domain being created on the `viewContext`
-        //  A solution to consider is creating a new background context here, creating, saving, and then re-fetching
-        //   that object in the requested context (regardless if it is `viewContext` or not)
         var newDomain: Domain!
-        context.performAndWait {
-            newDomain = Domain(entity: Domain.entity(context), insertInto: context)
+        
+        // We want to avoid saving data on view context.
+        // So if new domain is triggered on view context we save it synchronously using new background context,
+        // then the same object is refetched on the view context.
+        let writeContext = context == DataController.viewContext ? DataController.newBackgroundContext() : context
+        
+        writeContext.performAndWait {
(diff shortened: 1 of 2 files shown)


DataTests/DomainTests.swift
```diff
@@ -1,4 +1,4 @@
-// This Source Code Form is subject to the terms of the Mozilla Public
+    // This Source Code Form is subject to the terms of the Mozilla Public
```
jhreis, line 1: delete tab

Data/models/Domain.swift
```diff
@@ -38,19 +38,28 @@ public final class Domain: NSManagedObject, CRUD {
...
-                DataController.save(context: context)
+                DataController.save(context: writeContext)
             }
         }
+        
+        // Refetching only for view context. For all other cases we want to keep this object on the same
+        // context to avoid data errors(for example a Domain can be created while adding a Bookmark,
+        // if Bookmark and Domain would be on different contextes that would cause the app to crash)
+        if context == DataController.viewContext {
+            newDomain = context.object(with: newDomain.objectID) as? Domain
```
jhreis, line 60: Are we totally sure that this doesn't introduce a race condition. Although the Domain creation above is sync, the save operation on the writecontext is not, neither is the merging back into the view context.

Although your unit tests may pass, I'm not totally convinced it isn't coincidental. At the very least we should do extensive testing here, I will run some of my own test on this too.

PR #674, opened 2019-01-04
Fix #680: Set daily ping to true if it is a new calendar day.

### LINKED ISSUE #680: Rolling DAU -> Fixed Date (Midnight)
Currently DAU uses a rolling 24 hour period (e.g. `daily=true` _after_ 24 passed since last one).

This can lead to things like `daily=false&monthly=true` if a month changed prior to the 24 hour mark was passed.

This should be modified to use hard midnight based date.

e.g. 
11:59pm daily = true
12:01am daily = true

iccub:
## Pull Request Checklist

## Screenshots

### DIFF
--- a/BraveShared/Analytics/DAU.swift
+++ b/BraveShared/Analytics/DAU.swift
@@ -213,21 +213,27 @@ public class DAU {
             return nil
         }
         
-        let dSecs = Int(today.timeIntervalSince1970) - stat[0]
         let _month = stat[1]
         let _year = stat[2]
-        let SECONDS_IN_A_DAY = 86400
         
-        // On first ping, the user is all three of these
-        let daily = dSecs >= SECONDS_IN_A_DAY
-                
+        let lastPingDate = Date(timeIntervalSince1970: Double(stat[0]))
+        
+        guard let lastPingDay = DAU.calendar.components([.day], from: lastPingDate).day, let currentDay = todayComponents.day else {
+            log.error("Could not unwrap calendar components from date")
+            return nil
+        }
+         
         let weeksMonday = Preferences.DAU.lastPingFirstMonday.value
         // There is no lastPingFirstMondayKey preference set at first launch, meaning the week param should be set to true.
         let isFirstLaunchWeeksMonday = weeksMonday == nil
-        let weekly = todayComponents.weeksMonday != weeksMonday || isFirstLaunchWeeksMonday
-        
         let monthly = month != _month || year != _year
-        log.debug("Dau stat params, daily: \(daily), weekly: \(weekly), monthly:\(monthly), dSecs: \(dSecs)")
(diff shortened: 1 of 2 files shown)


BraveShared/Analytics/DAU.swift
```diff
@@ -213,21 +213,33 @@ public class DAU {
...
+            return nil
+        }
+        
+        let secondsInHours = hour * 60 * 60
+        let secondsInMinutes = minutes * 60
+        
+        let secondsSinceMidnight = secondsInHours + secondsInMinutes + seconds
+        
+        // Daily ping should count in two cases
+        // 1. At least 24 hours has passed.
```
jhreis: This never needs to be accounted for. #2 infers this. We just need to track what the day of year is and if it has changed, toggle daily=yes. The only situation that this would fail is if user used brave 74th day of 2019 and then the 74th day of 2020, so could double check the year for sanity.

So something like:

`daily = newDay != oldDay || newYear != oldYear`

iccub: I could just use `day` and `year` from calendar components to make it even simpler, what do you think?

jhreis: Yeah, that's basically what I meant, I think we can keep this pretty clean and lean.

BraveShared/Analytics/DAU.swift
```diff
@@ -213,21 +213,25 @@ public class DAU {
...
+            log.error("Could not unwrap calendar components from date")
+            return nil
+        }
+         
         let weeksMonday = Preferences.DAU.lastPingFirstMonday.value
         // There is no lastPingFirstMondayKey preference set at first launch, meaning the week param should be set to true.
         let isFirstLaunchWeeksMonday = weeksMonday == nil
+        let monthly = month != _month || year != _year
         let weekly = todayComponents.weeksMonday != weeksMonday || isFirstLaunchWeeksMonday
+        let daily = monthly || lastPingDay != currentDay
```
jhreis, line 234: Can you had a comment here, as to why you included `monthly`?

BraveShared/Analytics/DAU.swift
```diff
@@ -213,21 +213,25 @@ public class DAU {
             return nil
         }
         
-        let dSecs = Int(today.timeIntervalSince1970) - stat[0]
+        let secondsSinceLastPing = Int(today.timeIntervalSince1970) - stat[0]
```
jhreis: Is this even used anymore? I think it is just used in the logging, right? If so, let's remove.

iccub: Good catch, Xcode warning didn't show up because it's used in a debug log

BraveShared/Analytics/DAU.swift
iccub: Added, github didn't outdate this comment

jhreis (approved):
Sick job!

PR #681, opened 2019-01-08
Fix #692, fix #693, general CoreData stack improvements.

### LINKED ISSUE #692: CoreData concurrency issues
### Steps to Reproduce 
  1. Launch the app with `-com.apple.CoreData.ConcurrencyDebug 1` argument

**Actual result:** 
There's a lot of crashes because contextes are accessed on wrong threads.

**Expected result:**

**Brave Version:** 

**Device details:** 

**Website problems only:**

### LINKED ISSUE #693: DataTests are failing intermittently
### Steps to Reproduce 
  1. Launch DataTests few times

**Actual result:** 
Sometimes tests fail, sometimes not

**Expected result:**

**Brave Version:** 

**Device details:** 

**Website problems only:**

iccub:
Goals:
- avoid concurrency problems
- avoid using multiple contextes per call
- synchronize writes so no data corruption/merge conflicts will happen
- treat background context task as a database transaction, thread-safe one off task
- never write with view context

The basic concept is to add a `OperationQueue` on top of background context, so saves to database will be done in background but one after each other to prevent any weird merging.

## Pull Request Checklist

## Screenshots


Client.xcodeproj/xcshareddata/xcschemes/Fennec.xcscheme
```diff
@@ -102,6 +102,16 @@
             ReferencedContainer = "container:Client.xcodeproj">
          </BuildableReference>
       </MacroExpansion>
+      <CommandLineArguments>
+         <CommandLineArgument
+            argument = "-com.apple.CoreData.ConcurrencyDebug 1"
```
iccub, line 112: This flag allows us to catch concurrency bugs, accessing contextes from wrong thread etc. I turned it on by default on development and test schemes.

Client.xcodeproj/xcshareddata/xcschemes/Fennec.xcscheme
```diff
@@ -102,6 +102,16 @@
...
+            isEnabled = "YES">
+         </CommandLineArgument>
+         <CommandLineArgument
+            argument = "-com.apple.CoreData.SQLDebug 1"
```
iccub, line 116: This shows us database mount location and SQL queries, it has 4 levels of verbosity. Turned off by default.

Data/models/Bookmark.swift
```diff
@@ -96,59 +96,62 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
...
-            // so we need to fetch the object using background context.
-        } else if managedObjectContext?.concurrencyType != .privateQueueConcurrencyType {
-            contextToUpdate = DataController.newBackgroundContext()
-        }
+                       sendToSync: Bool = false, context: NSManagedObjectContext? = nil) {
         
-        guard let bookmarkToUpdate = (try? contextToUpdate?.existingObject(with: objectID)) as? Bookmark,
-            let context = contextToUpdate else {
-                return
+        if let context = context {
```
iccub: This is a pattern I use in many places in code. It's supposed to prevent launching multiple background tasks for a given code. 

When context is provided we never have to worry about any threading because the  code is already wrapped in a safe thread.
When context is not provided we can then safely launch a new background task.

This solution is not perfect because you need to be aware of passing correct context all the time, as not providing any context results in launching a new thread.

Data/models/DataController.swift
```diff
@@ -37,6 +37,12 @@ public class DataController: NSObject {
         return container
     }()
     
+    private lazy var operationQueue: OperationQueue = {
+        let queue = OperationQueue()
+        queue.maxConcurrentOperationCount = 1
```
iccub: This should not be a bottleneck even for heavy operations. We can always use operations with a different priority, so more important tasks will be finished first.
From my testing the queue count never exceeded 2 items, this will probably increase after implementing Sync.

Client/Frontend/Browser/TabManager.swift
```diff
@@ -315,7 +315,13 @@ class TabManager: NSObject {
         let type: TabType = isPrivate ? .private : .regular
         let tab = Tab(configuration: configuration, type: type)
         if !isPrivate {
-            tab.id = id ?? TabMO.create().syncUUID
+            if id != nil {
+                tab.id = id
+            } else {
+                DataController.performBackgroundTask { context in
```
jhreis: Not sure I follow this change. This function previously just used its own context and returned. Since this context isn't persisted (just used to pass into `create`), I think this is basically functionally equivalent, no?

Client/Frontend/Browser/TabManager.swift
```diff
@@ -356,17 +365,14 @@ class TabManager: NSObject {
...
             for (i, tab) in self.allTabs.enumerated() {
-                guard let managedObject = TabMO.get(fromId: tab.id, context: context) else { 
+                guard let managedObject = TabMO.get(fromId: tab.id, context: context) else {
                     log.error("Error: Tab missing managed object")
                     continue
                 }
                 managedObject.order = Int16(i)
             }
-            
-            DataController.save(context: context)
```
jhreis, line 383: Why get rid of this save? `performBackgroundTask` doesn't auto-save, does it?

Client/Frontend/ContentBlocker/TrackingProtectionPageStats.swift
```diff
@@ -64,7 +64,17 @@ class TPStatsBlocklistChecker {
         }
         
         DispatchQueue.global().async {
-            let enabledLists = BlocklistName.blocklists(forDomain: domain).on
+            var domainOnBackgroundContext: Domain?
+            var enabledLists = Set<BlocklistName>()
+            
+            DataController.performBackgroundTask { context in
+                domainOnBackgroundContext = context.object(with: domain.objectID) as? Domain
```
jhreis: I think inlining this into the `if-let` below would be cleaner:
```
if let domainOnBackgroundContext = context.object(with: domain.objectID) as? Domain {
    enabledLists = BlocklistName.blocklists(forDomain: bgDomain).on
}
```

Client/Frontend/ContentBlocker/TrackingProtectionPageStats.swift
```diff
@@ -64,7 +64,17 @@ class TPStatsBlocklistChecker {
         }
         
         DispatchQueue.global().async {
-            let enabledLists = BlocklistName.blocklists(forDomain: domain).on
+            var domainOnBackgroundContext: Domain?
+            var enabledLists = Set<BlocklistName>()
+            
+            DataController.performBackgroundTask { context in
```
jhreis: Given that these lists are only _read_ (I don't see any manipulation of them), and we don't even return the `Domain` that is pulled, I think using the `viewContext` is probably fine. Or do you think this is better since it doesn't _need_ to be on main thread?

Data/models/Bookmark.swift
jhreis: I think we should put in an assert
`assert(!isMainThread, "No saving updating on main thread")`
And use this liberally, assuming there are no more known instances of updating on main thread.

If this is not on the mainthread it can then be reduced more simply to:
```
let performingContext = context ?? DataController.newBackgroundContext
DataController.perform(with: performingContext) { // See convo on `performBackgroundTask` to understand this
     guard let bookmarkToUpdate = performingContext.object(with: self.objectID) as? Bookmark else { return }
     bookmarkToUpdate.updateImpl(customTitle: customTitle, url: url, newSyncOrder: newSyncOrder, save: save,
                           sendToSync: sendToSync, context: backgroundContext)
}
```

AFAIK, this is true for most of the other instances around too.

Data/models/Bookmark.swift
```diff
@@ -96,59 +96,62 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
...
+            updateImpl(customTitle: customTitle, url: url, newSyncOrder: newSyncOrder, save: save, sendToSync: sendToSync, context: context)
+        } else {
+            DataController.performBackgroundTask { [weak self] backgroundContext in
```
jhreis: Don't think we should use a `weak self` here, if something else deallocated this bookmark from memory it would never get updated in CD.

Data/models/Bookmark.swift
```diff
@@ -170,12 +173,28 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
                    save: Bool = false,
                    sendToSync: Bool = false,
                    parentFolder: Bookmark? = nil,
-                   context: NSManagedObjectContext = DataController.newBackgroundContext()) -> Bookmark? {
+                   context: NSManagedObjectContext? = nil) {
+        
+        if let context = context {
```
jhreis: To give another example from convo above:

```
let performingContext = context ?? DataController.newBackgroundContext
DataController.perform(with: performingContext) { // See convo on `performBackgroundTask` to understand this
    addImpl(rootObject: root, save: save, sendToSync: sendToSync, parentFolder: parentFolder, context: backgroundContext)
}
```

Or could even just throw a context at `addImpl` and have it deal with this internally.

Data/models/Bookmark.swift
```diff
@@ -238,12 +252,10 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
     }
     
     public class func add(from list: [(url: URL, title: String)]) {
-        let context = DataController.newBackgroundContext()
-        context.performAndWait {
+        DataController.performBackgroundTask { context in
```
jhreis: This is no longer synchronous, does this impact anything?

Data/models/DataController.swift
```diff
@@ -72,10 +78,33 @@ public class DataController: NSObject {
         return backgroundContext
     }
     
+    public static func performBackgroundTask(save: Bool = true, _ code: @escaping (NSManagedObjectContext) -> Void) {
```
jhreis: I think if we made everything pass in a context, and if we know these are all writish operations, assert on the `viewContext` being passed in. This would be a great stopping point to catch any main thread writes. It would alos allow us to simplify all of the above logic I believe (see those code snippets about the `if context = context {}` code.

Data/models/DataController.swift
```diff
@@ -72,10 +78,33 @@ public class DataController: NSObject {
...
+            backgroundContext.performAndWait {
+                code(backgroundContext)
+                
+                if save {
+                    do {
+                        if backgroundContext.hasChanges {
+                            try backgroundContext.save()
+                        }
+                    } catch {
+                        print(error)
```
jhreis: `log.error`

jhreis (changes requested):

Client/Frontend/Browser/TabManager.swift
iccub: You are right, we could simplify this code and do `performBackgroundTask` inside of the `TabMO.create()` method.
Basically the only difference is that creating TabMO object is put on an operation queue

Client/Frontend/Browser/TabManager.swift
iccub: It saves by default, I'm even considering removing `DataController.save()` function from code.
We almost always save after doing operations within a context, so this is a good default behavior to have imo

Client/Frontend/ContentBlocker/TrackingProtectionPageStats.swift
iccub: Yeah we can also use view context inside of `performAndWait` block, otherwise it throws an error. Not sure what's better in this case

Data/models/Bookmark.swift
iccub: Not from my observations, it adds the preloaded favorites which are automatically fetched by fetch results controller.

Data/models/Bookmark.swift
iccub: Sounds good to me, I wasn't exactly sure how to simplify all these `if let context = context {..` code I added

Client/Frontend/Browser/TabManager.swift
danishjafri88: I suggest not using any DataController functions directly in non NSManagedObject class(Unless required of course). Everything can be managed in the data classes themselves.

iccub: Good observation, I will try to remove all DataController code from non `Data` framework classes, to keep it more separate

iccub:
This PR also fixes Core Data unit tests, however I still haven't enable them in Travis, will do that in a separate PR

BraveShared/Preferences.swift
```diff
@@ -55,6 +55,27 @@ extension Preferences {
         public static let adblock = Option<String?>(key: "blockfile.adblock", default: nil)
         public static let httpse = Option<String?>(key: "blockfile.httpse", default: nil)
     }
+    
+    public final class Shields {
```
iccub, line 59: No code logic was changed regarding to shields, but I had to move some of the code to BraveShared, otherwise it was hard to separate `Client` logic from `Data`

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -158,8 +158,9 @@ extension BrowserViewController: WKNavigationDelegate {
             
             if let urlHost = url.normalizedHost {
                 if let mainDocumentURL = navigationAction.request.mainDocumentURL, url.scheme == "http" {
-                    let domainForShields = Domain.getOrCreateForUrl(mainDocumentURL, context: DataController.viewContext)
-                    if domainForShields.isShieldExpected(.HTTPSE) && HttpsEverywhereStats.shared.shouldUpgrade(url) {
+                    let domainForShields = Domain.getOrCreateForUrl(mainDocumentURL)
+                    let isPrivateBrowsing = PrivateBrowsingManager.shared.isPrivateBrowsing
```
iccub, line 162: `Domain` is in data and don't have access to `PrivateBrowsingManager`, so I have to pass it here. This happens in few other places too.

Client/Frontend/Browser/Search/SearchLoader.swift
```diff
@@ -92,23 +92,6 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<[Site], SearchViewController> {
             }
         }
     }
-    
-    // TODO: This is not a proper frecency query, it just gets sites from the past week. See issue #289
-    fileprivate func getSitesByFrecency(_ containing: String? = nil) -> Deferred<[Site]> {
```
iccub, line 97: This is moved to `Data`, it still uses background context to get websites.

Client/Frontend/Browser/TabManager.swift
```diff
@@ -329,7 +329,7 @@ class TabManager: NSObject {
         let type: TabType = isPrivate ? .private : .regular
         let tab = Tab(configuration: configuration, type: type)
         if !isPrivate {
-            tab.id = id ?? TabMO.create().syncUUID
+            tab.id = id ?? TabMO.create()
```
iccub, line 332: This now creates a new tab and returns its `syncUUID`. 
Since create is asynchronous I don't want to return whole TabMO object, we only use syncUUID after creation anyway.
The downside is that if the save will fail we will still return the id, in the future, deferred could be used for such cases(same with `getOrCreate` Domain)

Client/Frontend/Browser/TabManager.swift
```diff
@@ -370,18 +370,8 @@ class TabManager: NSObject {
...
-                    log.error("Error: Tab missing managed object")
-                    continue
-                }
-                managedObject.order = Int16(i)
-            }
-            
-            DataController.save(context: context)
-        }
+        let allTabIds = allTabs.compactMap { $0.id }
+        TabMO.saveTabOrder(tabIds: allTabIds)
```
iccub, line 374: Logic moved to `Data`.
`Tab` is in `Client` so can't use that, that's why I'm mapping and sending only ids of tabs.

Data/models/Bookmark.swift
```diff
@@ -59,11 +59,6 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
         return nil
     }
     
-    override public func awakeFromInsert() {
-        super.awakeFromInsert()
-        lastVisited = created
```
iccub, line 64: This was actually a bug, `awakeFromInsert` is called before we set `created` so it was always nil at this place. I moved all this logic into `add()` Bookmark method.

Data/models/Bookmark.swift
```diff
@@ -245,19 +222,29 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
...
+            site.location = url?.absoluteString
+            
+            let bookmark = SyncBookmark()
+            bookmark.isFavorite = isFavorite
+            bookmark.isFolder = isFolder
+            
+            var parentFolderOnCorrectContext: Bookmark?
+            
+            if let parent = parentFolder {
+                parentFolderOnCorrectContext = context.object(with: parent.objectID) as? Bookmark
```
iccub: Only difference here is getting parent folder on the background context

Data/models/Bookmark.swift
```diff
@@ -81,57 +76,42 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
...
-            // so we need to fetch the object using background context.
-        } else if managedObjectContext?.concurrencyType != .privateQueueConcurrencyType {
-            contextToUpdate = DataController.newBackgroundContext()
-        }
+                       sendToSync: Bool = true, context: NSManagedObjectContext? = nil) {
         
-        guard let bookmarkToUpdate = (try? contextToUpdate?.existingObject(with: objectID)) as? Bookmark,
-            let context = contextToUpdate else {
+        DataController.performTask(context: context) { context in
+            guard let bookmarkToUpdate = context.object(with: self.objectID) as? Bookmark else { return }
```
iccub: Only difference is getting Bookmark on the background context

Data/models/Bookmark.swift
```diff
@@ -276,76 +263,167 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
...
+        if dest === src {
+            log.error("Source and destination bookmarks are the same!")
+            return
+        }
         
-        DataController.save(context: context)
+        // To avoid mixing threads, the FRC can't be used within background context operation.
+        // We take data that relies on the FRC on main thread and pass it into background context
+        // in a safely manner.
+        guard let data = getReorderData(fromFrc: frc, sourceIndexPath: sourceIndexPath,
```
iccub: Reordering got a bit more complex so I added a lot of comments, I hope they are helpful

Data/models/Bookmark.swift
```diff
@@ -368,38 +446,14 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
...
             }
         }
-        
-        DataController.save(context: context)
-    }
-    
-    // TODO: Migration syncUUIDS still needs to be solved
-    // Should only ever be used for migration from old db
-    // Always uses worker context
-    class func addForMigration(url: String?, title: String, customTitle: String, parentFolder: Bookmark?, isFolder: Bool?) -> Bookmark? {
```
iccub, line 378: This is unused code taken from 1.6 codebase

Data/models/Bookmark.swift
```diff
@@ -460,7 +519,8 @@ extension Bookmark {
     }
     
     public static func getChildren(forFolderUUID syncUUID: [Int]?, includeFolders: Bool = true,
-                                   context: NSManagedObjectContext = DataController.viewContext) -> [Bookmark]? {
+                                   context: NSManagedObjectContext? = nil) -> [Bookmark]? {
+        let context = context ?? DataController.viewContext
```
iccub: This looks like it could be passed as a default argument for the method but it can't since `viewContext` is internal(this is what Xcode tells me)

Data/models/DataController.swift
```diff
@@ -10,38 +10,53 @@ private let log = Logger.browserLogger
...
     
-    private var viewContext: NSManagedObjectContext {
-        return container.viewContext
+    // Context object also allows us access to all persistent container data if needed.
+    static var viewContext: NSManagedObjectContext {
+        return DataController.shared.container.viewContext
     }
     
-    public static func save(context: NSManagedObjectContext?) {
+    static func save(context: NSManagedObjectContext?) {
```
iccub, line 70: This will be removed, now it's still used in few places

Data/models/DataController.swift
```diff
@@ -62,25 +77,7 @@ public class DataController: NSObject {
...
-    private func addPersistentStore(for container: NSPersistentContainer) {
-        if AppConstants.IsRunningTest {
-            let description = NSPersistentStoreDescription()
-            description.type = NSInMemoryStoreType
-            
-            container.persistentStoreDescriptions = [description]
-            return
-        }
-        
+    func addPersistentStore(for container: NSPersistentContainer) {
```
iccub, line 91: It's internal so `InMemoryStore` can overwrite it

Data/models/Domain.swift
```diff
@@ -83,58 +84,186 @@ public final class Domain: NSManagedObject, CRUD {
         return newDomain
     }
 
-    class func blockFromTopSites(_ url: URL, context: NSManagedObjectContext) {
```
iccub, line 86: These 3 methods are not used

Data/models/Domain.swift
```diff
@@ -83,58 +84,186 @@ public final class Domain: NSManagedObject, CRUD {
...
-                print(fetchError)
             }
-
-            DataController.save(context: context)
-            DispatchQueue.main.async {
-                completionOnMain()
+        }
+    }
+    
+    // MARK: - Shields
```
iccub: Shields which were previously stored in `Client` as a `Domain` extension

Data/sync/Sync.swift
```diff
@@ -392,14 +393,17 @@ extension Sync {
...
             let evaluate = "callbackList['send-sync-records'](null, '\(recordType.rawValue)',\(json))"
-            self.webView.evaluateJavaScript(evaluate,
-                                            completionHandler: { (result, error) in
-                                                if let error = error {
-                                                    print(error)
-                                                }
-                                                
-                                                completion?(error)
-            })
+            DispatchQueue.main.async {
```
iccub, line 396: `evaluateJavaScript` must be called from main thread

DataTests/CoreDataTestCase.swift
```diff
@@ -8,49 +8,65 @@ import CoreData
...
+        // because some operations are still pending in the app like creating a Domain for localhost
+        // attempting to restore tabs and some shields work.
+        // The workaround is to detect if sqlite database is still used(which means it's first launch of the test suite)
+        // wait a few seconds for it to finish its initialization, then switch the database to in-memory and run
+        // all the tests.
+        //
+        // In the future we should make Data tests more independent of the app state.
+        
+        let storeType = DataController.viewContext.persistentStoreCoordinator!.persistentStores.first!.type
+        if storeType == NSSQLiteStoreType {
```
iccub, line 24: This is a hack but works pretty reliably. I still haven't turned on DataTests for Travis yet.

iccub (commented):
I'm adding few comments regarding what changed and some notes.
Most of the changed code is just wrapping it with `performTask` closures

Data/sync/Sync.swift
```diff
@@ -392,14 +393,17 @@ extension Sync {
...
+                self.webView.evaluateJavaScript(evaluate,
+                                                completionHandler: { (result, error) in
+                                                    if let error = error {
+                                                        print(error)
```
jhreis: Let's get rid of this print, at least use logger.

Data/models/FrecencyQuery.swift
```diff
@@ -0,0 +1,35 @@
...
+import Storage
+import CoreData
+
+protocol WebsitePresentable {
+    var title: String? { get }
+    var url: String? { get }
+}
+
+protocol Frecencyable where Self: WebsitePresentable {
+    static func getByFrecency(query: String?, context: NSManagedObjectContext?) -> [WebsitePresentable]
```
jhreis: I think `byFrecency` is maybe a bit cleaner? Not sure `get` is needed. Thoughts?

Data/models/FrecencyQuery.swift
```diff
@@ -0,0 +1,35 @@
...
+}
+
+public class FrecencyQuery {
+    // TODO: This is not a proper frecency query, it just gets sites from the past week. See issue #289
+    public static func getSitesByFrecency(containing query: String? = nil) -> Deferred<[Site]> {
```
jhreis: I know this was pre-existing, but I think removing `get` here makes sense too.

Client/Frontend/Browser/TabManager.swift
jhreis: How many situations are like this? Maybe we should use deferred now? However, this could block the UI though since we'd have to wait for the save to complete, right?

Data/models/TabMO.swift
```diff
@@ -69,50 +69,61 @@ public final class TabMO: NSManagedObject, CRUD {
...
+        }
     }
 
     public class func getAll() -> [TabMO] {
         let sortDescriptors = [NSSortDescriptor(key: #keyPath(TabMO.order), ascending: true)]
         return all(sortDescriptors: sortDescriptors) ?? []
     }
     
-    public class func get(fromId id: String?, context: NSManagedObjectContext) -> TabMO? {
+    public class func get(fromId id: String?, context: NSManagedObjectContext? = nil) -> TabMO? {
```
jhreis: This concerns me a bit, as the caller may forget to be explicit. At the _very least_, we should set the default context here (`viewContext`) instead of just passing `nil` through. This way if the implementation of `first` changes for some odd reason, it doesn't recursively impact everything that is also using `get` here.

```suggestion
    public class func get(fromId id: String?, context: NSManagedObjectContext = DataController.viewContext) -> TabMO? {
```

Data/models/CRUDProtocols.swift
```diff
@@ -82,12 +78,15 @@ public extension Readable where Self: NSManagedObject {
         return nil
     }
     
-    static func first(where predicate: NSPredicate?, context: NSManagedObjectContext = DataController.viewContext) -> Self? {
-        return all(where: predicate, fetchLimit: 1, context: context)?.first
+    static func first(where predicate: NSPredicate? = nil, sortDescriptors: [NSSortDescriptor]? = nil,
+                      context: NSManagedObjectContext? = nil) -> Self? {
+        let context = context ?? DataController.viewContext
```
jhreis: Why is this being used instead of explicit in the function parameters? Makes inspection harder, when we're processing the context inside the function.

Data/models/DataController.swift
```diff
@@ -10,38 +10,53 @@ private let log = Logger.browserLogger
...
+        queue.addOperation({
+            let backgroundContext = DataController.newBackgroundContext()
+            // performAndWait doesn't block main thread because it fires on OperationQueue`s background thread.
+            backgroundContext.performAndWait {
+                code(backgroundContext)
+                
+                if !save { return }
+                
+                do {
+                    if backgroundContext.hasChanges {
```
jhreis: Can we move this outside/before the `do`?

Data/models/DataController.swift
```diff
@@ -10,38 +10,53 @@ private let log = Logger.browserLogger
...
-        guard let mom = NSManagedObjectModel(contentsOf: modelURL) else {
-            fatalError("Error initializing managed object model from: \(modelURL)")
+    public func storeExists() -> Bool {
+        return FileManager.default.fileExists(atPath: storeURL.path)
+    }
+    
+    // MARK: - Data framework interface
+    
+    static func performTask(context: NSManagedObjectContext? = nil, save: Bool = true,
+                            _ code: @escaping (NSManagedObjectContext) -> Void) {
```
jhreis: I think `code` here isn't very clear, it should also be a named parameter so it _can_ be called more clearly without using `{}`

e.g.
`performTask(execute: myCode)`

Honestly, could even be called `task` and drop the suffix off the function.

`DataController.perform {}`

DataController.perform(task: myCode)`

Client/Frontend/Menu/BookmarksViewController.swift
```diff
@@ -566,10 +566,12 @@ extension BookmarksViewController: NSFetchedResultsControllerDelegate {
   func controller(_ controller: NSFetchedResultsController<NSFetchRequestResult>, didChange anObject: Any, at indexPath: IndexPath?, for type: NSFetchedResultsChangeType, newIndexPath: IndexPath?) {
     switch type {
     case .update:
-      guard let indexPath = indexPath, let cell = tableView.cellForRow(at: indexPath) else {
-        return
-      }
-      configureCell(cell, atIndexPath: indexPath)
+        if let indexPath = indexPath, let cell = tableView.cellForRow(at: indexPath) {
```
jhreis: My hatred for duplicate code will show here 😄 

But I think doing something like:

```
let update = { (path: IndexPath?) in
     if let path = path, let cell = self.tableView.cellForRow(at: path) {
           self.configureCell(cell, atIndexPath: path)
     }
}
[indexPath, newIndexPath].forEach(update)
```

is a bit cleaner. wdyt? Same amount of code, just removes the duplication. 😁

jhreis (changes requested):

Data/models/CRUDProtocols.swift
iccub: `first` is public, `viewContext` is internal, and it can't be used as a default parameter.

I could make `viewContext` public, but sure what is better. Thinking about architecture I wanted to remove all direct references to `DataController` from `Client`

jhreis: Unfortunate, but understandable.

jhreis (approved):
Been a longtime coming, great work and excited to get all of this in!

PR #684, opened 2019-01-08
Fix #683: Localize bookmarks folder deletion alert

### LINKED ISSUE #683: Nonempty bookmarks folder deletion alert is not localized
### Description: 
If the user tries to delete a bookmarks folder with at least one bookmark, an alert is displayed to verify the decision. The alert's title, body, and buttons are not localized.

The following strings in the code need to be localized:
https://github.com/brave/brave-ios/blob/94ce279a8af8c2884f3723ab00a7f62e06a24073/Client/Frontend/Menu/BookmarksViewController.swift#L438-L440

### Steps to Reproduce 
  1. Switch the device to a language other than English.
  2. Open Bookmarks and create a new bookmarks folder.
  3. Add a bookmark to the bookmarks folder.
  4. Try to delete the bookmarks folder.

**Actual result:**
The alert is not localized:

**Expected result:**
The alert should be localized.

Always

**Brave Version:** 
Latest master (94ce279)

**Device details:** 
Any

**Website problems only:**

### Additional Information

cndouglas:
Fix #683

## Pull Request Checklist

The actual localization cannot be tested until translated strings are imported from Transifex.


kylehickinson (approved):
Sweet, nice catch!

PR #773, opened 2019-01-22
Fix #666: Disable top-level navigation for data scheme

### LINKED ISSUE #666: [hackerone] #473654 - Address Bar Spoofing
https://hackerone.com/bugs?report_id=473654

jumde:
fix https://github.com/brave/brave-ios/issues/666

Top-level navigation to `data:` urls are not supported on major mobile browsers. On Brave iOS, navigating to data-urls does not work correctly.

Try opening `data:text/html;base64,SGVsbG8gV29ybGQhCg==` in Desktop vs Mobile Browser

This PR removes support for navigating to data URLs.

## Submitter Checklist:

### Direct navigation to data:urls
1. Navigate to `data:text/html;base64,SGVsbG8gV29ybGQhCg==` should do a search instead of initiating a navigation to a data URL

### Popups and subresources

1.  Navigate to http://dustiest-limitation.000webhostapp.com/data.html
2. Click on Google.com - data URL should not open
3. Images and CSS rules are loaded correctly.

## Reviewer Checklist:

  - [x] `QA/(Yes|No)`
  - [x] `bug` / `enhancement`

### DIFF
--- a/Shared/Extensions/URLExtensions.swift
+++ b/Shared/Extensions/URLExtensions.swift
@@ -103,8 +103,10 @@ extension URL {
 // The list of permanent URI schemes has been taken from http://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml 
 private let permanentURISchemes = ["aaa", "aaas", "about", "acap", "acct", "cap", "cid", "coap", "coaps", "crid", "data", "dav", "dict", "dns", "example", "file", "ftp", "geo", "go", "gopher", "h323", "http", "https", "iax", "icap", "im", "imap", "info", "ipp", "ipps", "iris", "iris.beep", "iris.lwz", "iris.xpc", "iris.xpcs", "jabber", "ldap", "mailto", "mid", "msrp", "msrps", "mtqp", "mupdate", "news", "nfs", "ni", "nih", "nntp", "opaquelocktoken", "pkcs11", "pop", "pres", "reload", "rtsp", "rtsps", "rtspu", "service", "session", "shttp", "sieve", "sip", "sips", "sms", "snmp", "soap.beep", "soap.beeps", "stun", "stuns", "tag", "tel", "telnet", "tftp", "thismessage", "tip", "tn3270", "turn", "turns", "tv", "urn", "vemmi", "vnc", "ws", "wss", "xcon", "xcon-userid", "xmlrpc.beep", "xmlrpc.beeps", "xmpp", "z39.50r", "z39.50s"]
 
-extension URL {
+private let ignoredSchemes = ["data"]
+private let supportedSchemes = permanentURISchemes.filter { !ignoredSchemes.contains($0) }
 
+extension URL {
     public func withQueryParams(_ params: [URLQueryItem]) -> URL {
         var components = URLComponents(url: self, resolvingAgainstBaseURL: false)!
         var items = (components.queryItems ?? [])
@@ -310,7 +312,7 @@ extension URL {
      */
     public var schemeIsValid: Bool {
         guard let scheme = scheme else { return false }
-        return permanentURISchemes.contains(scheme.lowercased())
+        return supportedSchemes.contains(scheme.lowercased())
     }
 
     public func havingRemovedAuthorisationComponents() -> URL {
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -2096,7 +2096,7 @@ extension BrowserViewController: TabManagerDelegate {
 }
 
 /// List of schemes that are allowed to be opened in new tabs.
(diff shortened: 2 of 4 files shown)


jumde:
cc: @tomlowenthal

iccub:
What more needs to be done in this pr @jumde?

Also I would rename the commit to say something like `disable support for data scheme`, it is a generic change that fixes the security issue as well.

jumde:
@iccub - I'm checking out the test suite to see if any test can be added for this change. 

Updated the commit message, thank you

iccub:
`testSuggestClient` broke it seems

iccub:
Tests work on my local machine, I will rerun Travis

iccub (approved):
Looks good to me and the commit message is great, thanks.

Approving and waiting for @jhreis for final approval and merge

Shared/Extensions/URLExtensions.swift
```diff
@@ -101,7 +101,7 @@ extension URL {
...
+private let permanentURISchemes = ["aaa", "aaas", "about", "acap", "acct", "cap", "cid", "coap", "coaps", "crid", "dav", "dict", "dns", "example", "file", "ftp", "geo", "go", "gopher", "h323", "http", "https", "iax", "icap", "im", "imap", "info", "ipp", "ipps", "iris", "iris.beep", "iris.lwz", "iris.xpc", "iris.xpcs", "jabber", "ldap", "mailto", "mid", "msrp", "msrps", "mtqp", "mupdate", "news", "nfs", "ni", "nih", "nntp", "opaquelocktoken", "pkcs11", "pop", "pres", "reload", "rtsp", "rtsps", "rtspu", "service", "session", "shttp", "sieve", "sip", "sips", "sms", "snmp", "soap.beep", "soap.beeps", "stun", "stuns", "tag", "tel", "telnet", "tftp", "thismessage", "tip", "tn3270", "turn", "turns", "tv", "urn", "vemmi", "vnc", "ws", "wss", "xcon", "xcon-userid", "xmlrpc.beep", "xmlrpc.beeps", "xmpp", "z39.50r", "z39.50s"]
```
jhreis: I'm not so sure that we should modify this list, as it is pulled from an external source. What about creating a list of exceptions underneath / above?

Something like:

```
private let permanentURISchemes = // ...
private let ignoredSchemes = ["data"]
private let supportedSchemes = permanentURISchemes.removeAll(where: ignoredSchemes.contains)
```

Then just update `schemeIsValid` to use `supportedSchemes`:

```
    public var schemeIsValid: Bool {
        guard let scheme = scheme else { return false }
        return supportedSchemes.contains(scheme.lowercased())
    }
```

jhreis (changes requested):
Just one small critique. Also we should update the `testschemeIsValid` unit test to include a `"data"` url inside `badurls`.

Shared/Extensions/URLExtensions.swift
```diff
@@ -103,8 +103,10 @@ extension URL {
 // The list of permanent URI schemes has been taken from http://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml 
 private let permanentURISchemes = ["aaa", "aaas", "about", "acap", "acct", "cap", "cid", "coap", "coaps", "crid", "data", "dav", "dict", "dns", "example", "file", "ftp", "geo", "go", "gopher", "h323", "http", "https", "iax", "icap", "im", "imap", "info", "ipp", "ipps", "iris", "iris.beep", "iris.lwz", "iris.xpc", "iris.xpcs", "jabber", "ldap", "mailto", "mid", "msrp", "msrps", "mtqp", "mupdate", "news", "nfs", "ni", "nih", "nntp", "opaquelocktoken", "pkcs11", "pop", "pres", "reload", "rtsp", "rtsps", "rtspu", "service", "session", "shttp", "sieve", "sip", "sips", "sms", "snmp", "soap.beep", "soap.beeps", "stun", "stuns", "tag", "tel", "telnet", "tftp", "thismessage", "tip", "tn3270", "turn", "turns", "tv", "urn", "vemmi", "vnc", "ws", "wss", "xcon", "xcon-userid", "xmlrpc.beep", "xmlrpc.beeps", "xmpp", "z39.50r", "z39.50s"]
 
-extension URL {
+private let ignoredSchemes = ["data"]
+private let supportedSchemes = permanentURISchemes.filter { !ignoredSchemes.contains($0) }
```
jhreis, line 107: Made a mistake previously with the `removeAll`, however the inverse of `filter` is `drop`, and will probably look a bit cleaner:

```suggestion
private let supportedSchemes = permanentURISchemes.drop(where: ignoredSchemes.contains)
```

jhreis: This is wrong too. 😄

jhreis (approved):

PR #794, opened 2019-01-25
No Bug: Fixing DAU ping race condition on app activation.

jhreis:
I need to build out some better unit tests for this, however some larger architecture stuff will probably need to be adjusted a bit.

## Submitter Checklist:

## Reviewer Checklist:

  - [x] `QA/(Yes|No)`
  - [x] `bug` / `enhancement`

### DIFF
--- a/BraveShared/Analytics/DAU.swift
+++ b/BraveShared/Analytics/DAU.swift
@@ -21,12 +21,14 @@ public class DAU {
     fileprivate static var calendar: NSCalendar { return Calendar(identifier: .gregorian) as NSCalendar }
     
     private var launchTimer: Timer?
-    private let today: Date
+    private let today: Date?
+    /// Whether a current ping attempt is being made
+    private var processingPing = false
     private var todayComponents: DateComponents {
-        return DAU.calendar.components([.day, .month, .year, .weekday], from: today)
+        return DAU.calendar.components([.day, .month, .year, .weekday], from: today ?? Date())
     }
     
-    public init(date: Date = Date()) {
+    public init(date: Date? = nil) {
         today = date
     }
     
@@ -61,6 +63,12 @@ public class DAU {
             return
         }
         
+        if processingPing {
+            log.info("Currently processing a ping, blocking ping re-attempt")
+            return
+        }
+        processingPing = true
+        
         // Sending ping to server
         var pingRequest = URLComponents(string: DAU.baseUrl)
         pingRequest?.queryItems = paramsAndPrefs.queryParams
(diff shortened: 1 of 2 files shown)


iccub (approved):

BraveShared/Analytics/DAU.swift
```diff
@@ -38,6 +40,11 @@ public class DAU {
             return false
         }
         
+        if processingPing {
```
danishjafri88: This is redundant once you make the DAU class object global. The previous check also do the same thing. Basically you cant have a sync check for a async updation. Also the check below for launchTimer != nil will make sure no new timers are created.

danishjafri88: This is redundant once you make the DAU class object global. The check below for launchTimer != nil will make sure no new timers are created. Instead put this check at `guard let paramsAndPrefs = paramsAndPrefsSetup() else ` and set `processingPing = true` after the guard

BraveShared/Analytics/DAU.swift
```diff
@@ -56,6 +63,7 @@ public class DAU {
     }
     
     @objc public func sendPingToServerInternal() {
+        processingPing = true
         guard let paramsAndPrefs = paramsAndPrefsSetup() else {
```
danishjafri88: Check for `processingPing ` also and set `processingPing` as true after the guard.

BraveShared/Analytics/DAU.swift
```diff
@@ -86,6 +94,7 @@ public class DAU {
             Preferences.DAU.lastLaunchInfo.value = paramsAndPrefs.lastLaunchInfoPreference
             
             Preferences.DAU.lastPingFirstMonday.value = paramsAndPrefs.lastPingFirstMondayPreference
+            self.processingPing = false
```
danishjafri88: You need to set self.processingPing = false in above `if let e = error {` also to complete the check.

danishjafri88 (changes requested):

danishjafri88:
The change is required, let take a case:
Time fires 1st time: Calls the `sendPingToServerInternal `  and `processingPing ` is set as true.
before the api call is done (Should not if the time is more than api timeout but for architecture sake). The timer fires again and there is no check to make sure the api call is not done.

jhreis:
@danishjafri88 yeah, good point, will adjust.

BraveShared/Analytics/DAU.swift
jhreis: Added the former, but why the latter?

BraveShared/Analytics/DAU.swift
jhreis: Did move, and `dau.sendPingToServer()` is still called manually on app open.

BraveShared/Analytics/DAU.swift
jhreis: yikes, thx

BraveShared/Analytics/DAU.swift
danishjafri88: It will always be processingPing ie true if paramsAndPrefsSetup failed for today but not after midnight

danishjafri88 (approved):

PR #809, opened 2019-01-28
Fix #789: Don't add a Device if it already exists.

### LINKED ISSUE #789: Sometimes new sync device is added twice
### Steps to Reproduce 
  1. Pair two devices together

**Actual result:** 
It does not happen often, but sometimes the device that joins your sync chain is added twice on one device.

**Expected result:**

**Brave Version:** 

**Device details:** 

**Website problems only:**

iccub:
This is a workaround solution. Proper fix will be done once https://github.com/brave/brave-ios/pull/681 is merged in, so we have an OperationQueue to avoid any race conditions.

## Submitter Checklist:

## Reviewer Checklist:

  - [x] `QA/(Yes|No)`
  - [x] `bug` / `enhancement`

### DIFF
--- a/Data/models/Device.swift
+++ b/Data/models/Device.swift
@@ -5,6 +5,8 @@ import CoreData
 import Foundation
 import Shared
 
+private let log = Logger.browserLogger
+
 public final class Device: NSManagedObject, Syncable, CRUD {
     
     // Check if this can be nested inside the method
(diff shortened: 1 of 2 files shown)


Data/models/Device.swift
```diff
@@ -49,10 +51,26 @@ public final class Device: NSManagedObject, Syncable, CRUD {
...
+        // There can't be more than two device with the same syncUUID.
+        // A race condition could sometimes happen while getting two `Sync.resolvedSyncRecords` callbacks at the same time.
+        // Fixing issue #692 should help with it, until then we do a simple guard and disallow adding another device
+        // with the same syncUUID.
+        if let syncDisplayUUID = SyncHelpers.syncDisplay(fromUUID: syncUUID) {
+            let predicate = NSPredicate(format: "syncDisplayUUID == %@", syncDisplayUUID)
+            let first = Device.first(where: predicate, context: context)
+            if first != nil {
+                log.warning("A device with sync UUID \(syncDisplayUUID) already exists. Aborting inserting a new device.")
+                return nil
```
jhreis: What do you think about just failing silently and letting this bleed through?

```
var device = Device.first(where: predicate, context: context)
if device == nil {
  // This is expected, however, see race condition comments above.
  device = Device(entity: Device.entity(context: context), insertInto: context)
  device.created = root?.syncNativeTimestamp ?? Date()
  device.syncUUID = syncUUID
}

device.update(syncRecord: root)
```

iccub: The question is, do we want to eventually update record on insert in this case?
We do something similar in `Sync.resolvedRecords` so that shouldn't be a problem

Data/models/Device.swift
```diff
@@ -49,12 +51,25 @@ public final class Device: NSManagedObject, Syncable, CRUD {
...
+            let predicate = NSPredicate(format: "syncDisplayUUID == %@", syncDisplayUUID)
+            device = Device.first(where: predicate, context: context)
+        }
+        
+        if device == nil {
+            device = Device(entity: Device.entity(context: context), insertInto: context)
+            device?.created = root?.syncNativeTimestamp ?? Date()
+            device?.syncUUID = syncUUID
+            
+            device?.update(syncRecord: root)
```
jhreis: I think we can place this outside the `if`? That way it would update regardless of what device it will use.

jhreis (changes requested):

Data/models/Device.swift
iccub: Why only `update` then and no syncUUID and `created` which also uses `root` object?

jhreis (approved):

PR #818, opened 2019-01-29
WIP: Added logic to save Custom URP cookie 

danishjafri88:
Added logic to save Custom URP cookie to be sent with request for respective URP URLs

WIP + Documentation

## Submitter Checklist:

## Reviewer Checklist:

  - [ ] `QA/(Yes|No)`
  - [ ] `bug` / `enhancement`

### DIFF
--- a/BraveShared/Analytics/UserReferralProgram.swift
+++ b/BraveShared/Analytics/UserReferralProgram.swift
@@ -8,6 +8,8 @@ import Shared
 private let log = Logger.browserLogger
 
 public class UserReferralProgram {
+    
+    private static let urpPartnerDomains: [String] = ["https://coinbase.com", "http://example.com"]
     public static let shared = UserReferralProgram()
     
     private static let apiKeyPlistKey = "API_KEY"
@@ -211,5 +213,47 @@ public class UserReferralProgram {
         }
         
         return nil
+    }
+    
+    public class func loadURPCookies() -> [HTTPCookie] {
+        var cookies: [HTTPCookie] = []
+        for partnerURLString in UserReferralProgram.urpPartnerDomains {
+            if let partnerURL: URL = URL(string: partnerURLString), let cookieDict = UserReferralProgram.customCookiesDict(for: partnerURL) {
+                for (key, value) in cookieDict {
+                    if let cookie: HTTPCookie = HTTPCookie(properties: [
+                        .domain: partnerURL.host!,
+                        .path: "/",
+                        .name: key,
+                        .value: value,
+                        .secure: "TRUE",
+                        .expires: NSDate(timeIntervalSinceNow: 31556926)
+                        ]) {
+                        cookies.append(cookie)
+                    }
+                }
+            }
+        }
+        return cookies
+    }
+    
+    public class func customCookiesDict(for url: URL) -> [String: String]? {
+        guard let customHeadersAsData = Preferences.URP.customHeaderData.value,
+            let customHeaders = (try? NSKeyedUnarchiver.unarchiveTopLevelObjectWithData(customHeadersAsData)) as? [CustomHeaderData],
+            let hostUrl = url.host else { return nil }
+        var cookieDict: [String: String]?
+        for customHeader in customHeaders {
+            // There could be an egde case when we would have two domains withing different domain groups, that would
+            // cause to return only the first domain-header it approaches.
+            for domain in customHeader.domainList {
+                if hostUrl.contains(domain) {
+                    if cookieDict == nil {
+                        cookieDict = [customHeader.headerField: customHeader.headerValue]
+                    } else {
+                        cookieDict?[customHeader.headerField] = customHeader.headerValue
+                    }
+                }
+            }
+        }
+        return cookieDict
     }
 }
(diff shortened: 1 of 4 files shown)


BraveShared/Analytics/UserReferralProgram.swift
```diff
@@ -189,27 +200,64 @@ public class UserReferralProgram {
...
+                let domains = header.domainList.compactMap { URL(string: $0)?.absoluteString }
+                cookies += domains.compactMap {
+                    HTTPCookie(properties: [
+                        // Must include `.` prefix to be included in subdomains
+                        .domain: ".\($0)",
+                        .path: "/",
+                        .name: header.headerField,
+                        .value: header.headerValue,
+                        .secure: "TRUE",
+                        .expires: NSDate(timeIntervalSinceNow: 7.days)
```
jumde, line 255: Do we expect the cookies to be accessed by scripts on coinbase? If not, we should probably add HTTPOnly flag as well: https://www.owasp.org/index.php/HttpOnly

BraveShared/Analytics/UserReferralProgram.swift
```diff
@@ -189,27 +200,64 @@ public class UserReferralProgram {
...
             // There could be an egde case when we would have two domains withing different domain groups, that would
             // cause to return only the first domain-header it approaches.
-            for domain in customHeader.domainList {
-                if hostUrl.contains(domain) {
-                    if let allFields = request.allHTTPHeaderFields, !allFields.keys.contains(customHeader.headerField) {
-                        return (customHeader.headerValue, customHeader.headerField)
-                    }
+            for domain in customHeader.domainList where hostUrl.contains(domain) {
+                // If `domain` is "cookie only", we exclude it from HTTP Headers, and just use cookie approach
+                let cookieOnly = urpCookieOnlyDomains.contains(domain)
```
jumde, line 232: We are checking all request urls, this means the cookie will be set for any URL that using coinbase assets?  Can we limit this check to urls loaded in top-level frame? Let me know if I'm missing something.

BraveShared/Analytics/UserReferralProgram.swift
```diff
@@ -189,27 +200,64 @@ public class UserReferralProgram {
...
+        customHeadersDeferred.upon { customHeaders in
+            var cookies: [HTTPCookie] = []
+            customHeaders.forEach { header in
+                let domains = header.domainList.compactMap { URL(string: $0)?.absoluteString }
+                cookies += domains.compactMap {
+                    HTTPCookie(properties: [
+                        // Must include `.` prefix to be included in subdomains
+                        .domain: ".\($0)",
+                        .path: "/",
+                        .name: header.headerField,
```
fmarier, line 252: Ideally, it would be good to prefix the cookie name with [`__Secure-`](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-02#section-4.1.3) to enforce that the `secure` flag was set.

BraveShared/Analytics/UserReferralProgram.swift
danishjafri88: Webkit looks for cross domain referencing and does not add for such cases. ie if a some other site beside a domain of coinbase should load a coinbase url then the cookie won't be sent. This is not true for iFrames obviously. Apart from this there is no way to limit cookie sent to document url only.

BraveShared/Analytics/UserReferralProgram.swift
jhreis: We do toggle the `secure` which according to docs determines "whether the cookie should be transmitted only over secure channels." Do we still want to manually add the `__Secure-` string? Follows the [RFC 6265](https://tools.ietf.org/html/rfc6265) spec.

fmarier: The `__Secure-` prefix is a complement to the `secure` flag. In browsers that support it (Firefox, Chromium), the prefix prevents someone from setting a cookie with that name without the secure flag. That way the consumer of the cookie (i.e. the server application) can be assured that the cookie received was originally set with that flag and was never sent over HTTP. Without the prefix, there's no way to tell since the `secure` flag isn't part of the cookie request header.

BraveShared/Analytics/UserReferralProgram.swift
jhreis: Unfortunately that flag is `readonly`, and not settable during the creation of the cookie (like all of these other flags).

BraveShared/Analytics/UserReferralProgram.swift
jhreis: Gotcha, is it supported in webkit (safari)? If it won't hurt I can just add it regardless.

jhreis: Please move discussion to: https://github.com/brave/brave-ios/pull/823/files

BraveShared/Analytics/UserReferralProgram.swift
jhreis: Please move discussion to: https://github.com/brave/brave-ios/pull/823/files

BraveShared/Analytics/UserReferralProgram.swift
jhreis: Please move any additional discussion to: https://github.com/brave/brave-ios/pull/823/files

jhreis:
Closing in favor of: https://github.com/brave/brave-ios/pull/823

PR #835, opened 2019-02-04
Fix #698: Launching Brave from external URL or Referral tab now loads/focuses correctly.

### LINKED ISSUE #698: Brave Doesn't Auto-Switch to Tab
### Description: 

Feedback from Discord:

> I'm calling out to the installed Brave browser on my iPhone from some swift code in MY application and passing a URL in the call and have noticed that if the Brave browser is already loaded, the correct action takes place--the iPhone switches over to the running Brave browser, creates a new tab and drops my URL into the address bar and all is well.  **If the Brave browser is NOT running however, the call out to Brave launches the Brave browser, creates a new tab with the passed URL, but DOES NOT SWITCH over to the new tab!**  It stays on whatever was the last tab in focus prior to the Brave app being closed. The end result is that the user doesn't see the URL expected but instead sees the last URL the Brave browser had rendered.  Is there some extra parameter I need to pass to the call to the Brave browser to make the action I want, consistent? **(Emphasis added by Sampson)**

### Steps to Reproduce 

```obj-c
let browser = "brave"
let bundleID = "(browser)://"
if USE_BRAVE && UIApplication.shared.canOpenURL(URL(string: bundleID)!) {
    print("App is install and can be opened")
    let link = link0?.replacingOccurrences(of: "https", with: "http")
                     .replacingOccurrences(of: "http", with: "(browser)://open-url?url=http")
    print("retailerUrl (link as Optional)")
    let url = URL(string:link!)
    if #available(iOS 10.0, *) {
        UIApplication.shared.open(url!, options: [:], completionHandler: nil)
    } else {
        UIApplication.shared.openURL(url!)
    }
}
```

danishjafri88:
The bug used to happen when the app was suspended ie closed. The open url in new tab was followed by the normal load tabs call. 
Fixed it by creating a taskQueue which is performed at the end of tabs load.

## Submitter Checklist:

## Test Plan:
-Go to any other app where you have an option to tap a link. For example, notes, slack or even safari
-Long press a link 
-On the action sheet select `share`
-Select `Brave` next

This should open a new tab with the link and should be in focus. Earlier it was opening the tab but was in focus.

## Reviewer Checklist:

  - [x] `QA/(Yes|No)`
  - [x] `bug` / `enhancement`


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -190,6 +192,15 @@ class BrowserViewController: UIViewController {
         // Lists need to be compiled before attempting tab restoration
         contentBlockListDeferred = ContentBlockerHelper.compileLists()
     }
+    
+    // Use this function to queue a task for after the BVC has loaded and is visible.(After shoing  initial tabs ).
```
jhreis: - [ ] Can you add a bit more info to the comment here, maybe even an `e.g.` (example)

```
// Use this function...
// e.g. present a popup to the user or select a loaded tab
```
(unsure if those are actual situations this could be used for, but adding something there may be useful to understand context and why this is important / needed.


- [ ] Also comment contains weird spacing

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -190,6 +192,15 @@ class BrowserViewController: UIViewController {
...
+    func addTask(task: @escaping () -> Void) {
```
jhreis: Per other comments, I don't think this is necessary anymore.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -56,6 +56,8 @@ class BrowserViewController: UIViewController {
     fileprivate var searchLoader: SearchLoader?
     fileprivate let alertStackView = UIStackView() // All content that appears above the footer should be added to this view. (Find In Page/SnackBars)
     fileprivate var findInPageBar: FindInPageBar?
+    
+    var taskQueue: [()->()] = []
```
jhreis: Using a promise here is perfect. Inside the BVC, just setup a `nil` promise:

```
let onLoad = Deferred<Void>()
```

Then instead of appending a task, just use `upon`

```
onLoad.upon {
  task()
}
```

Then when we're ready:

`onLoad.fill(())`

Then once it is filled all of the "tasks" will automatically be executed. Additionally, this avoids race conditions for us automatically too. Instead of needing to check whether to execute immediately or append to the queue, promises will auto-fire if they have already been filled. Less code, and less error prone.

Client/Application/AppDelegate.swift
```diff
@@ -232,9 +232,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
             if isFirstLaunch {
                 urp.referralLookup { url in
                     guard let url = url else { return }
-                    DispatchQueue.main.asyncAfter(deadline: .now() + 1, execute: {
+                    self.browserViewController.addTask {
```
jhreis: When internalizing this to BVC, this won't be needed `openURLInNewTab` will automatically handle this by adding a trigger on the promise.

Client/Application/AppDelegate.swift
```diff
@@ -259,8 +259,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
         guard let routerpath = NavigationPath(url: url) else {
             return false
         }
-
-        DispatchQueue.main.async {
+        
+        self.browserViewController.addTask {
```
jhreis: Per other comments, I think this can be removed, however we'd do need to add a new function to BVC to handle the `NavigationPath.handle` the function will just call the same function internally, but will need to setup itself up as a `promise.upon`.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -437,22 +448,28 @@ class BrowserViewController: UIViewController {
...
+            
+            if noTabsAdded {
+                // Two scenarios if there are no tabs in tabmanager:
+                // 1. We have not restored tabs yet, attempt to restore or make a new tab if there is nothing.
+                // 2. We are in private browsing mode and need to add a new private tab.
+                tabToSelect = isPrivate ? self.tabManager.addTab(isPrivate: true) : self.tabManager.restoreAllTabs()
+            } else {
+                tabToSelect = self.tabManager.tabsForCurrentMode.last
+            }
+            if self.taskQueue.isEmpty {
```
jhreis: This is a bit more complex with promises, since you can't (AFAIK), identify whether a promise has a closure attached to it, so we probably need to handle this with additional metadata:

```
private class onLoad {
    fileprivate let await = Deferred<Void>()
    fileprivate var containsTasks = false
    fileprivate func upon(_ block: @escaping () -> Void) {
        await.upon(block)
        containsTasks = true
    }
}
```

So then here we'd do something like:
`if onLoad.containsTasks {`

jhreis (changes requested):
I am thinking we should push most of this logic into the BVC itself, so AppDelegate can only call a function on the BVC that the BVC itself will handle this internally. We'd probably need two, one for each of the current AppDelegate calls. Additionally I think using a promise here would be more clear. I added a good bit of information regarding this, lmk what you think.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -435,22 +437,25 @@ class BrowserViewController: UIViewController {
...
+                // 1. We have not restored tabs yet, attempt to restore or make a new tab if there is nothing.
+                // 2. We are in private browsing mode and need to add a new private tab.
+                tabToSelect = isPrivate ? self.tabManager.addTab(isPrivate: true) : self.tabManager.restoreAllTabs()
+            } else {
+                tabToSelect = self.tabManager.tabsForCurrentMode.last
+            }
+            if self.loadQueue.isEmpty {
+                self.tabManager.selectTab(tabToSelect)
+            } else {
+                self.loadQueue.fill(())
```
jhreis: I think this should be called all the time, right?

jhreis (changes requested):

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -435,22 +437,25 @@ class BrowserViewController: UIViewController {
...
-        }
-        
         contentBlockListDeferred?.uponQueue(.main) { _ in
-            self.tabManager.selectTab(tabToSelect)
+            let isPrivate = Preferences.Privacy.privateBrowsingOnly.value
+            let noTabsAdded = self.tabManager.tabsForCurrentMode.isEmpty
+            
+            var tabToSelect: Tab?
+            
+            if noTabsAdded {
```
jhreis: This is a question, but should we even execute all of this code (lines 446-453) if the `loadQueue` has items to execute? All of those (currently anyway), lead to new tab creation, right?

Client/Frontend/Browser/BrowserViewController.swift
danishjafri88: Yup, thanks !!

jhreis (approved):
So much nicer! Great work!

PR #883, opened 2019-02-08
Fixes: #870 Replaced logic to clear data with WebKit api.

danishjafri88:
## Submitter Checklist:

## Reviewer Checklist:

  - [x] `QA/(Yes|No)`
  - [x] `bug` / `enhancement`

### DIFF
--- a/Client/FileManagerExtension.swift
+++ b/Client/FileManagerExtension.swift
@@ -13,6 +13,8 @@ public extension FileManager {
         case webSiteData = "/WebKit/WebsiteData"
     }
     typealias FolderLockObj = (folder: Folder, lock: Bool)
+    
+    //Lock a folder using FolderLockObj provided.
     public func setFolderAccess(_ lockObjects: [FolderLockObj]) -> Bool {
         let baseDir = NSSearchPathForDirectoriesInDomains(.libraryDirectory, .userDomainMask, true)[0]
         for lockObj in lockObjects {
@@ -24,5 +26,19 @@ public extension FileManager {
             }
         }
         return true
+    }
+    
+    // Check the locked status of a folder. Returns true for locked.
+    public func checkLockedStatus(folder: Folder) -> Bool {
+        let baseDir = NSSearchPathForDirectoriesInDomains(.libraryDirectory, .userDomainMask, true)[0]
+        do {
+            if let lockValue = try self.attributesOfItem(atPath: baseDir + folder.rawValue)[.posixPermissions] as? NSNumber {
+                return lockValue == NSNumber(value: 0o755 as Int16)
+            }
+        } catch let e {
+            log.error("Failed to check lock status on item at path \(folder.rawValue) with error: \n\(e)")
+            return false
+        }
+        return false
     }
 }
--- a/Client/FileManagerExtension.swift
+++ b/Client/FileManagerExtension.swift
@@ -19,7 +19,7 @@ public extension FileManager {
         let baseDir = NSSearchPathForDirectoriesInDomains(.libraryDirectory, .userDomainMask, true)[0]
         for lockObj in lockObjects {
             do {
-                try self.setAttributes([.posixPermissions: (lockObj.lock ? NSNumber(value: 0 as Int16) : NSNumber(value: 0o755 as Int16))], ofItemAtPath: baseDir + lockObj.folder.rawValue)
+                try self.setAttributes([.posixPermissions: (lockObj.lock ? 0 : 0o755)], ofItemAtPath: baseDir + lockObj.folder.rawValue)
             } catch {
                 log.error("Failed to \(lockObj.lock ? "Lock" : "Unlock") item at path \(lockObj.folder.rawValue) with error: \n\(error)")
                 return false
@@ -33,11 +33,10 @@ public extension FileManager {
         let baseDir = NSSearchPathForDirectoriesInDomains(.libraryDirectory, .userDomainMask, true)[0]
         do {
             if let lockValue = try self.attributesOfItem(atPath: baseDir + folder.rawValue)[.posixPermissions] as? NSNumber {
-                return lockValue == NSNumber(value: 0o755 as Int16)
+                return lockValue == 0o755
             }
         } catch {
             log.error("Failed to check lock status on item at path \(folder.rawValue) with error: \n\(error)")
-            return false
         }
         return false
     }
--- a/Client/FileManagerExtension.swift
+++ b/Client/FileManagerExtension.swift
@@ -15,13 +15,13 @@ public extension FileManager {
     typealias FolderLockObj = (folder: Folder, lock: Bool)
     
     //Lock a folder using FolderLockObj provided.
-    public func setFolderAccess(_ lockObjects: [FolderLockObj]) -> Bool {
+    @discardableResult public func setFolderAccess(_ lockObjects: [FolderLockObj]) -> Bool {
         let baseDir = NSSearchPathForDirectoriesInDomains(.libraryDirectory, .userDomainMask, true)[0]
         for lockObj in lockObjects {
             do {
                 try self.setAttributes([.posixPermissions: (lockObj.lock ? NSNumber(value: 0 as Int16) : NSNumber(value: 0o755 as Int16))], ofItemAtPath: baseDir + lockObj.folder.rawValue)
-            } catch let e {
-                log.error("Failed to \(lockObj.lock ? "Lock" : "Unlock") item at path \(lockObj.folder.rawValue) with error: \n\(e)")
+            } catch {
+                log.error("Failed to \(lockObj.lock ? "Lock" : "Unlock") item at path \(lockObj.folder.rawValue) with error: \n\(error)")
                 return false
             }
         }
@@ -35,8 +35,8 @@ public extension FileManager {
             if let lockValue = try self.attributesOfItem(atPath: baseDir + folder.rawValue)[.posixPermissions] as? NSNumber {
                 return lockValue == NSNumber(value: 0o755 as Int16)
             }
-        } catch let e {
-            log.error("Failed to check lock status on item at path \(folder.rawValue) with error: \n\(e)")
+        } catch {
+            log.error("Failed to check lock status on item at path \(folder.rawValue) with error: \n\(error)")
             return false
         }
         return false
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -2920,9 +2920,9 @@ extension BrowserViewController: PreferencesObserver {
              Preferences.Shields.fingerprintingProtection.key:
             tabManager.allTabs.forEach { $0.webView?.reload() }
         case Preferences.Privacy.blockAllCookies.key:
-            if Preferences.Privacy.blockAllCookies.value {
-                tabManager.reset()
-            } else {
+            // All `block all cookies` toggle requires a hard rest of Webkit configuration.
+            tabManager.reset()
+            if !Preferences.Privacy.blockAllCookies.value {
                 tabManager.allTabs.forEach {
                     if let url: URL = $0.webView?.url {
                         $0.loadRequest(PrivilegedRequest(url: url) as URLRequest)
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -2920,7 +2920,7 @@ extension BrowserViewController: PreferencesObserver {
              Preferences.Shields.fingerprintingProtection.key:
             tabManager.allTabs.forEach { $0.webView?.reload() }
         case Preferences.Privacy.blockAllCookies.key:
-            // All `block all cookies` toggle requires a hard rest of Webkit configuration.
+            // All `block all cookies` toggle requires a hard reset of Webkit configuration.
             tabManager.reset()
             if !Preferences.Privacy.blockAllCookies.value {
                 tabManager.allTabs.forEach {
--- a/Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
+++ b/Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
@@ -178,7 +178,7 @@ class ClearPrivateDataTableViewController: UITableViewController {
                 self.tabManager.resetConfiguration()
                 // Unlock the folders to allow clearing of data.
                 if Preferences.Privacy.blockAllCookies.value {
-                    _ = FileManager.default.setFolderAccess([
+                    FileManager.default.setFolderAccess([
                         (.cookie, false),
                         (.webSiteData, false)
                         ])
@@ -188,14 +188,13 @@ class ClearPrivateDataTableViewController: UITableViewController {
                 // TODO: add API to avoid add/remove
                 //Lock the local storage back if mismatch.
                 if Preferences.Privacy.blockAllCookies.value, !FileManager.default.checkLockedStatus(folder: .cookie) {
-                    _ = FileManager.default.setFolderAccess([
+                    FileManager.default.setFolderAccess([
                         (.cookie, true),
                         (.webSiteData, true)
                         ])
                 }
                 self.tabManager.removeTab(self.tabManager.addTab())
             }
-            //      }
         })
     }
     
--- a/Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
+++ b/Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
@@ -176,9 +176,23 @@ class ClearPrivateDataTableViewController: UITableViewController {
             // Reset Webkit configuration to remove data from memory
             if clear.contains(where: { $0 is CookiesClearable || $0 is CacheClearable }) {
                 self.tabManager.resetConfiguration()
+                // Unlock the folders to allow clearing of data.
+                if Preferences.Privacy.blockAllCookies.value {
+                    _ = FileManager.default.setFolderAccess([
+                        (.cookie, false),
+                        (.webSiteData, false)
+                        ])
+                }
             }
             ClearPrivateDataTableViewController.clearPrivateData(clear).uponQueue(DispatchQueue.main) { _ in
                 // TODO: add API to avoid add/remove
+                //Lock the local storage back if mismatch.
+                if Preferences.Privacy.blockAllCookies.value, !FileManager.default.checkLockedStatus(folder: .cookie) {
+                    _ = FileManager.default.setFolderAccess([
+                        (.cookie, true),
+                        (.webSiteData, true)
+                        ])
+                }
                 self.tabManager.removeTab(self.tabManager.addTab())
             }
             //      }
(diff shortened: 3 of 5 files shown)


Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
```diff
@@ -172,6 +172,11 @@ class ClearPrivateDataTableViewController: UITableViewController {
              }
              }
              } else {*/
+            
+            // Reset Webkit configuration to remove data from memory
+            if !clear.filter({ $0 is CookiesClearable || $0 is CacheClearable}).isEmpty {
```
iccub: Can be simplified to `!clear.contains(where: { $0 is CookiesClearable || $0 is CacheClearable })`

Client/Frontend/Settings/Clearables.swift
```diff
@@ -126,24 +115,14 @@ class CacheClearable: Clearable {
...
             
-            var err: Error?
-            for item in ["Caches", "Cookies", "WebKit"] {
-                do {
-                    try deleteLibraryFolderContents(item, validateClearedExceptFor: ["Snapshots"])
-                } catch {
-                    err = error
-                }
+            WKWebsiteDataStore.default().removeData(ofTypes: WKWebsiteDataStore.allWebsiteDataTypes(), modifiedSince: Date(timeIntervalSinceReferenceDate: 0)) {
+                // Clear image cache
```
iccub: This comment can be removed, it adds nothing of value, I know it was there before the changes :)

Client/Frontend/Settings/Clearables.swift
```diff
@@ -126,24 +115,14 @@ class CacheClearable: Clearable {
...
+                ImageCache.shared.clear()
+                
+                // Leave the cache off in the error cases above
```
iccub: This says `in the error cases above` but we removed errors with the new implementation, do we still need this? Maybe the comments is just outdated, might require some research

Client/Frontend/Settings/Clearables.swift
```diff
@@ -126,24 +115,14 @@ class CacheClearable: Clearable {
...
             URLCache.shared.removeAllCachedResponses()
             
-            var err: Error?
-            for item in ["Caches", "Cookies", "WebKit"] {
-                do {
-                    try deleteLibraryFolderContents(item, validateClearedExceptFor: ["Snapshots"])
-                } catch {
-                    err = error
-                }
+            WKWebsiteDataStore.default().removeData(ofTypes: WKWebsiteDataStore.allWebsiteDataTypes(), modifiedSince: Date(timeIntervalSinceReferenceDate: 0)) {
```
iccub: Previously we deleted only 3 items `["Caches", "Cookies", "WebKit"]`
Are we sure `allWebsiteDataTypes()` removes the same stuff and doesn't break anything?

Client/Frontend/Settings/Clearables.swift
```diff
@@ -84,26 +85,14 @@ class CookiesClearable: Clearable {
...
-                }
+            WKWebsiteDataStore.default().removeData(ofTypes: [WKWebsiteDataTypeCookies], modifiedSince: Date(timeIntervalSinceReferenceDate: 0)) {
+                UserDefaults.standard.synchronize()
+                result.fill(Maybe<()>(success: ()))
             }
-            UserDefaults.standard.synchronize()
-            
-            // And just to be safe, we also wipe the Cookies directory.
-            do {
-                try deleteLibraryFolderContents("Cookies", validateClearedExceptFor: [])
```
iccub, line 102: `deleteLibraryFolderContents` doesn't seem to be used anymore, 
same with `deleteLibraryFolder`

iccub (changes requested):
I have mostly questions to not brake anything accidentally :)

Client/Frontend/Settings/Clearables.swift
danishjafri88: Yeah actually they used to delete more than required.

Client/Frontend/Settings/Clearables.swift
danishjafri88: Removed them thanks!

Client/Frontend/Settings/Clearables.swift
danishjafri88: Removed the comments as they are outdated

Client/Frontend/Settings/Clearables.swift
danishjafri88: Removed :)

Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
danishjafri88: Hehe missed my mind 👍

Client/Frontend/Settings/Clearables.swift
```diff
@@ -126,24 +75,11 @@ class CacheClearable: Clearable {
...
             }
-            if let err = err {
-                return result.fill(Maybe<()>(failure: ClearableErrorType(err: err)))
-            }
-            
-            // Clear image cache
-            ImageCache.shared.clear()
-            
-            // Leave the cache off in the error cases above
-            URLCache.shared.setupBraveDefaults()
```
iccub, line 145: Looks like `setupBraveDefaults` and the file containing it can be also removed then.

But now I see one place with `URLCache.shared.memoryCapacity = 0`, but no other place when we set different values to it.
Should we still set memory/disk capacities somewhere after the new changes?

danishjafri88: The memory capacity is set to 0 to flush current cache and then reset to defaults. And since we are using `allWebsiteDataTypes` which covers all types of caches (memory/disk) we can safely remove this. .. Removed. Thanks, don't why I am missing these so much :P

Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
jhreis: Can even use `contains(where:)`

`if clear.contains(where: { $0 is CookiesClearable || $0 is CacheClearable}) {`

or if using actual values, this should even work:

`if clear.contains(where: [CookiesClearable, CacheClearable].contains) {`

jhreis:
We should really add some automated tests around this.

- turn on cookie blocking
- clear all private data
- verify locks are still in place

Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
jhreis: I think @iccub solution is cleanest/most obvious, but pointing out this would probably work too:
`clear.contains(where: [CookiesClearable, CacheClearable].contains)`
(although falls apart where polymorphism is required)

Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
```diff
@@ -172,6 +172,11 @@ class ClearPrivateDataTableViewController: UITableViewController {
              }
              }
              } else {*/
+            
+            // Reset Webkit configuration to remove data from memory
+            if clear.contains(where: { $0 is CookiesClearable || $0 is CacheClearable }) {
+                self.tabManager.resetConfiguration()
```
jhreis, line 178: Is there any real downside to calling this all the time?

Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
danishjafri88: Also its a O(n^2) not that it matters in this case.

danishjafri88:
@jhreis I was looking to do so, the problem is that locking will work in actual device, for simulator the folder hierarchy is different and dynamic.

Client/Frontend/Settings/Clearables.swift
```diff
@@ -84,26 +45,14 @@ class CookiesClearable: Clearable {
...
         let result = Deferred<Maybe<()>>()
         // need event loop to run to autorelease UIWebViews fully
         DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
             // Now we wipe the system cookie store (for our app).
-            let storage = HTTPCookieStorage.shared
-            if let cookies = storage.cookies {
-                for cookie in cookies {
-                    storage.deleteCookie(cookie)
-                }
+            WKWebsiteDataStore.default().removeData(ofTypes: [WKWebsiteDataTypeCookies], modifiedSince: Date(timeIntervalSinceReferenceDate: 0)) {
```
diracdeltas: i would expect local storage / websql / indexdb, etc., to be cleared when cookies are cleared (though i know this wasn't the case before). we have an open issue for this in brave-core. might as well do that here if it's easy.

danishjafri88: Added, thanks for the heads up

diracdeltas: thanks, here is the b-c issue for reference: https://github.com/brave/brave-browser/issues/1127

jumde:
Does this not work on a simulator? I'm getting this on enabling `Block all cookies` without resetting data: 

screen shot 2019-02-11 at 3 14 47 pm

jumde:
@jhreis @danishjafri88 - For cookie control, looks like toggling the `block all cookies` switch (on and then off)  deletes all but `LocalStorage`. We should keep it consistent.

Do we need to delete all storage when `Block all cookies` is enabled? As a user I'd expect to log back in to `a.com` if I toggle on and off the switch quickly. This will not happen if all storage is deleted when the switch is turned on.

danishjafri88:
@jumde nothing is deleted when toggling `block all cookies` can you share an example, that may help.
And yeah it does not work on Simulator as the Folder hierarchy is not same as a device.

Client/Frontend/Settings/Clearables.swift
diracdeltas: question - is this tied to the 'clear cache' switch? if so, why does it also clear cookies?

Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
danishjafri88: No actual downside to resetConfigurations here. The session has to be cleared.

tildelowengrimm:
Looks like the cache checkbox clears cookies but not vice versa? That seems weird. Could we just consolidate into "Cookies & site data" and clear cookies, local storage, cache, &c all together? Or is there a need to be able to delete cache without cookies?

danishjafri88:
@tomlowenthal yeah, this was copied from FF. Any changes to current implementation should be done in a new PR as an enhancement. If you could please create an issue for it(Feature Enhancement) that would be great! :)

Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
jhreis: I just meant, can we remove the conditional?

No more:
`if clear.contains(where: { $0 is CookiesClearable || $0 is CacheClearable }) {`
and just call
`resetConfiguration()` every time?

danishjafri88:
@jhreis @iccub @diracdeltas, minor code update. Handled a case where block all cookies might be on and can hinder clearing data. And added a hard reset of Webview configuration for both cases of block-all cookies toggle. This is needed as the Config retains the session in memory.

Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
danishjafri88: Oh I thought you meant calling it while clearing data. History and saved logins are not part of webkit config and a reset is not required. Although no downside to it. ~~If you too think its not a big hassle I can make it happen always.~~ Additionally we need this check for unlocking the folders.

jumde:
> @jumde nothing is deleted when toggling `block all cookies` can you share an example, that may help.
> And yeah it does not work on Simulator as the Folder hierarchy is not same as a device.

@danishjafri88 - 

Steps to repro:

1. Navigate to `http://dustiest-limiation.000webhostapp.com/first_party.html` - sets all storage
2. Navigate to `http://dustiest-limiation.000webhostapp.com/data.html` - no storage api calls here
3. Turn on `Block all cookies`. Using the Safari Inspector verify the cookies and storage is blocked.
4. Turn off `Block all cookies` - Cookies and Session Storage is deleted.

danishjafri88:
@jumde Ah I think this is because of the revert we did recently, I fixed this in the last [commit](https://github.com/brave/brave-ios/pull/883/commits/5912c4a6bccfcad29dee9fd641df58806b99bcb0) Can you check now please. with latest commit. I verified the same with your steps.

jumde:
> @jumde Ah I think this is because of the revert we did recently, I fixed this in the last [commit](https://github.com/brave/brave-ios/pull/883/commits/5912c4a6bccfcad29dee9fd641df58806b99bcb0) Can you check now please. with latest commit. I verified the same with your steps.

@danishjafri88 - this is still the same for me. used the same repro steps. Cookies and session storage are deleted.

jumde:
> > @jumde Ah I think this is because of the revert we did recently, I fixed this in the last [commit](https://github.com/brave/brave-ios/pull/883/commits/5912c4a6bccfcad29dee9fd641df58806b99bcb0) Can you check now please. with latest commit. I verified the same with your steps.
> 
> @danishjafri88 - this is still the same for me. used the same repro steps. Cookies and session storage are deleted.

Discussed this with @danishjafri88 offline, filed a follow-up issue to track this: https://github.com/brave/brave-ios/issues/897

jhreis:
> > > @jumde Ah I think this is because of the revert we did recently, I fixed this in the last [commit](https://github.com/brave/brave-ios/pull/883/commits/5912c4a6bccfcad29dee9fd641df58806b99bcb0) Can you check now please. with latest commit. I verified the same with your steps.
> > 
> > 
> > @danishjafri88 - this is still the same for me. used the same repro steps. Cookies and session storage are deleted.
> 
> Discussed this with @danishjafri88 offline, filed a follow-up issue to track this: #897

👍 

@danishjafri88 regarding this, is [this commit](https://github.com/brave/brave-ios/pull/883/commits/5912c4a6bccfcad29dee9fd641df58806b99bcb0) necessary now, if it does not solve the issue?

jumde:
> @tomlowenthal yeah, this was copied from FF. Any changes to current implementation should be done in a new PR as an enhancement. If you could please create an issue for it(Feature Enhancement) that would be great! :)

Created: https://github.com/brave/brave-ios/issues/898

jumde:
approved

Client/FileManagerExtension.swift
```diff
@@ -25,4 +27,18 @@ public extension FileManager {
...
     }
+    
+    // Check the locked status of a folder. Returns true for locked.
+    public func checkLockedStatus(folder: Folder) -> Bool {
+        let baseDir = NSSearchPathForDirectoriesInDomains(.libraryDirectory, .userDomainMask, true)[0]
+        do {
+            if let lockValue = try self.attributesOfItem(atPath: baseDir + folder.rawValue)[.posixPermissions] as? NSNumber {
+                return lockValue == NSNumber(value: 0o755 as Int16)
+            }
+        } catch let e {
```
iccub: No need for `let e`, you can just use `error` property inside of the catch closure

Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
```diff
@@ -172,8 +172,27 @@ class ClearPrivateDataTableViewController: UITableViewController {
...
+                // Unlock the folders to allow clearing of data.
+                if Preferences.Privacy.blockAllCookies.value {
+                    _ = FileManager.default.setFolderAccess([
```
iccub: Could we add `@discardableResults` to the `setFolderAccess` method?

danishjafri88:
> > > > @jumde Ah I think this is because of the revert we did recently, I fixed this in the last [commit](https://github.com/brave/brave-ios/pull/883/commits/5912c4a6bccfcad29dee9fd641df58806b99bcb0) Can you check now please. with latest commit. I verified the same with your steps.
> > > 
> > > 
> > > @danishjafri88 - this is still the same for me. used the same repro steps. Cookies and session storage are deleted.
> > 
> > 
> > Discussed this with @danishjafri88 offline, filed a follow-up issue to track this: #897
> 
> 👍
> 
> @danishjafri88 regarding this, is [this commit](https://github.com/brave/brave-ios/pull/883/commits/5912c4a6bccfcad29dee9fd641df58806b99bcb0) necessary now, if it does not solve the issue?

@jhreis Yeah still required. This does not fix the problem PJ mentioned, rather the problem that websites won't function properly once you switch off block-all-cookies.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2920,9 +2920,9 @@ extension BrowserViewController: PreferencesObserver {
              Preferences.Shields.fingerprintingProtection.key:
             tabManager.allTabs.forEach { $0.webView?.reload() }
         case Preferences.Privacy.blockAllCookies.key:
-            if Preferences.Privacy.blockAllCookies.value {
-                tabManager.reset()
-            } else {
+            // All `block all cookies` toggle requires a hard rest of Webkit configuration.
```
iccub: typo `rest -> reset`

iccub (changes requested):
Only few small style changes remaining, rest looks good to me

Client/Frontend/Browser/BrowserViewController.swift
danishjafri88: Corrected

Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
danishjafri88: updated

Client/FileManagerExtension.swift
danishjafri88: updated

iccub (approved):

Client/FileManagerExtension.swift
```diff
@@ -13,16 +13,32 @@ public extension FileManager {
...
+    // Check the locked status of a folder. Returns true for locked.
+    public func checkLockedStatus(folder: Folder) -> Bool {
+        let baseDir = NSSearchPathForDirectoriesInDomains(.libraryDirectory, .userDomainMask, true)[0]
+        do {
+            if let lockValue = try self.attributesOfItem(atPath: baseDir + folder.rawValue)[.posixPermissions] as? NSNumber {
+                return lockValue == NSNumber(value: 0o755 as Int16)
+            }
+        } catch {
+            log.error("Failed to check lock status on item at path \(folder.rawValue) with error: \n\(error)")
+            return false
```
jhreis: Is this return needed? I think we can just let it bleed through and return below, right?

Client/FileManagerExtension.swift
```diff
@@ -13,16 +13,32 @@ public extension FileManager {
...
         }
         return true
     }
+    
+    // Check the locked status of a folder. Returns true for locked.
+    public func checkLockedStatus(folder: Folder) -> Bool {
+        let baseDir = NSSearchPathForDirectoriesInDomains(.libraryDirectory, .userDomainMask, true)[0]
+        do {
+            if let lockValue = try self.attributesOfItem(atPath: baseDir + folder.rawValue)[.posixPermissions] as? NSNumber {
+                return lockValue == NSNumber(value: 0o755 as Int16)
```
jhreis: Why `0o755 as Int16`?

jhreis (changes requested):

Client/FileManagerExtension.swift
danishjafri88: No particular reason other than habit. :) In fact we can use it directly too.(w/o Nsnumber)

jhreis (approved):

Issue #887, opened 2019-02-11
security issues with URP logic


diracdeltas:
cc @bbondy @jumde 

there are 2 security improvements we should make to the URP logic:

1. in https://github.com/brave/brave-ios/blob/9a239cddb164590fa98c09c1c976d8d09b078308/BraveShared/Analytics/UserReferralProgram.swift#L191, custom headers fetched from the server should only be accepted if the header name is `X-Brave-Partner`; otherwise log/throw an error. This prevents an attacker who controls the Brave endpoint from being able to set any headers other than `X-Brave-Partner`.

2. for Coinbase, which uses cookies instead of headers, you should do some additional validation before inserting the cookie(s) into the cookie store (https://github.com/brave/brave-ios/pull/823/files#diff-60f6d7911cf19ceb7bcfcde55f9bc740R240): (a) validate that the cookie is being inserted only for coinbase.com, otherwise throw an error; (b) validate that the cookie name is `__Secure-X-Brave-Partner`.

diracdeltas:
On second thought, re: cookies - why fetch them from the server at all?

The original reason for fetching headers from the server was so that the promos could be changed without a browser update; but in the case of cookies, a browser update is needed anyway for updating this list: https://github.com/brave/brave-ios/pull/823/files#diff-60f6d7911cf19ceb7bcfcde55f9bc740R15

jhreis:
Currently the list is a blacklist for partners that do not want the HTTP Header (not for cookie opt-in). This was a short-term patch, and @aekeus was looking at adding additional data to the request to specify which solution (or both) each partner would use.

We just didn't have time, given the intense constraints we had.

diracdeltas:
TODO: make sure header value cannot be maliciously formed to create a new header name

PR #892, opened 2019-02-11
Fixes: #887 Validation for http header key

danishjafri88:
Validation for http header key wrt: https://github.com/brave/brave-ios/issues/887#issue-408629525

## Submitter Checklist:

## Reviewer Checklist:

  - [x] `QA/(Yes|No)`
  - [x] `bug` / `enhancement`


diracdeltas:
This does not fix https://github.com/brave/brave-ios/issues/887 entirely unless you also add validation for the ` __Secure-X-Brave-Partner` cookie name. 

EDIT: @jumde says this is implicitly done because the cookie name is from the header; i still think it would be better to explicitly check it when the cookie is inserted

jhreis:
@diracdeltas we derive the cookies from the headers, and this filters before saving to disk, so this would protect against both cookies and headers, guaranteeing both have the value `X-Brave-Partner` and we manually prepend the `__Secure-` flag when storing the actual cookies. 

Where we fetch the data: https://github.com/brave/brave-ios/blob/9a239cddb164590fa98c09c1c976d8d09b078308/BraveShared/Analytics/UserReferralProgram.swift#L191

This is the function that is called, which calls the method that we have modified in this PR:
https://github.com/brave/brave-ios/blob/521a8b0a1fee17f5da1927bfefc6e33a3088a287/BraveShared/Analytics/UrpService.swift#L95

The data here (after validated to have correct name), fulfills the promise which creates the cookies:
https://github.com/brave/brave-ios/blob/9a239cddb164590fa98c09c1c976d8d09b078308/BraveShared/Analytics/UserReferralProgram.swift#L199

Hardcoded `__Secure` flag:
https://github.com/brave/brave-ios/blob/9a239cddb164590fa98c09c1c976d8d09b078308/BraveShared/Analytics/CustomHeaderData.swift#L36

A little hard via GitHub to show full path, but hope this gives a bit of clarity. Basically, we are filtering on the network connection callback and processing data ASAP so we know whatever is stored is correct and just ignore anything we don't want (non `X-Brave-Partner` names).

BraveShared/Analytics/CustomHeaderData.swift
```diff
@@ -45,7 +45,7 @@ class CustomHeaderData: NSObject {
         var customHeaders: [CustomHeaderData] = []
 
         for (_, object) in json {
-            guard let header = object["headers"].first else { continue }
+            guard let header: (String, JSON) = object["headers"].first, header.0 == "X-Brave-Partner" else { continue }
```
jhreis: Maybe abstract the `X-Brave-Partner` into a constant somewhere, just easier to see what's going on then. Can be in this file, just nice to have broken up.

diracdeltas:
@jhreis yup, see my edit; i did see that code path.

danishjafri88:
@diracdeltas added a check just before insert. Will this work?

diracdeltas:
much clearer now, thanks!

jumde:
PR looks good. Is it possible to add an automated test for this? Or if that's difficult do you know if there is an easy way to test it without setting up a proxy.

danishjafri88:
@jumde we can test the `func customHeaders(from json: JSON)` with different payloads. I will push one in morning.

jhreis:
@jumde we do have a server running that we could try using, but now sure how well localhost works with receiving cookies. @danishjafri88 can you give this a shot and see how far you get?

BraveSharedTests/CustomHeaderDataTests.swift
```diff
@@ -0,0 +1,36 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import XCTest
+@testable import BraveShared
+import SwiftyJSON
+
+class CustomHeaderDataTest: XCTestCase {
+    
+    //Valid Json from server.
+    var expectedJSON = "[{\"domains\":[\"coinbase.com\",\"api.coinbase.com\"],\"headers\":{\"X-Brave-Partner\":\"coinbase\"},\"cookieNames\":[],\"expiration\":31536000000},{\"domains\":[\"marketwatch.com\",\"barrons.com\"],\"headers\":{\"X-Brave-Partner\":\"dowjones\"},\"cookieNames\":[],\"expiration\":31536000000},{\"domains\":[\"townsquareblogs.com\",\"tasteofcountry.com\",\"ultimateclassicrock.com\",\"xxlmag.com\",\"popcrush.com\"],\"headers\":{\"X-Brave-Partner\":\"townsquare\"},\"cookieNames\":[],\"expiration\":31536000000},{\"domains\":[\"cheddar.com\"],\"headers\":{\"X-Brave-Partner\":\"cheddar\"},\"cookieNames\":[],\"expiration\":31536000000}]"
```
jhreis: Use `"""`, then you don't need to escape all of the `"`s

var expectedJSON = """[{"domains":["coinbase.com"]}..."""`

jhreis: Let's use completely fabricated data here, and not real websites/services, since this is all hardcoded (even though this is a test, still will stay in our code for people to see).

BraveSharedTests/CustomHeaderDataTests.swift
```diff
@@ -0,0 +1,36 @@
...
+    var maliciousJSON = "[{\"domains\":[\"coinbase.com\",\"api.coinbase.com\"],\"headers\":{\"Not-X-Brave-Partner\":\"coinbase\"},\"cookieNames\":[],\"expiration\":31536000000},{\"domains\":[\"marketwatch.com\",\"barrons.com\"],\"headers\":{\"Not-X-Brave-Partner\":\"dowjones\"},\"cookieNames\":[],\"expiration\":31536000000},{\"domains\":[\"townsquareblogs.com\",\"tasteofcountry.com\",\"ultimateclassicrock.com\",\"xxlmag.com\",\"popcrush.com\"],\"headers\":{\"Not-X-Brave-Partner\":\"townsquare\"},\"cookieNames\":[],\"expiration\":31536000000},{\"domains\":[\"cheddar.com\"],\"headers\":{\"Not-X-Brave-Partner\":\"cheddar\"},\"cookieNames\":[],\"expiration\":31536000000}]"
+    
+    
+    func testExample() {
+        // This is an example of a functional test case.
+        // Use XCTAssert and related functions to verify your tests produce the correct results.
+        let expectedHeaders = getCustomHeader(for: expectedJSON)
+        let maliciousHeaders = getCustomHeader(for: maliciousJSON)
+        //Tests for non empty array of headers from valid json
+        XCTAssertTrue(!expectedHeaders.isEmpty)
```
jhreis:
```suggestion
        XCTAssertFalse(expectedHeaders.isEmpty)
```

BraveSharedTests/CustomHeaderDataTests.swift
```diff
@@ -0,0 +1,36 @@
...
+        //Tests the non empty array for the correct key
+        XCTAssertTrue(expectedHeaders.filter({$0.headerField != CustomHeaderData.bravePartnerKey}).isEmpty)
```
jhreis:
```suggestion
        XCTAssertFalse(expectedHeaders.contains { $0.headerField != CustomHeaderData.bravePartnerKey })
```

jhreis (changes requested):

BraveSharedTests/CustomHeaderDataTests.swift
danishjafri88: Done

BraveSharedTests/CustomHeaderDataTests.swift
danishjafri88: lol thanks!!!

BraveSharedTests/CustomHeaderDataTests.swift
danishjafri88: updated.

jumde:
approved ++

jhreis (approved):

Issue #897, opened 2019-02-13
Cookies deleted when `Block-all-cookies` is toggled on and then off


jumde:
### Description: 
`Block all cookies` should not delete cookies.

### Steps to Reproduce 
  * Navigate to `http://dustiest-limitation.000webhostapp.com/first_party.html` - sets all storage
  * Navigate to `http://dustiest-limitation.000webhostapp.com/data.html` - no storage api calls here
  * Turn on `Block all cookies`.
  * Navigate to `http://dustiest-limitation.000webhostapp.com/view.html` - to check if storage is not available
  * Turn off `Block all cookies`.
  * Refresh `http://dustiest-limitation.000webhostapp.com/view.html` - to check if storage is available

**Actual result:** 
Cookies are deleted

**Expected result:**
Cookies should not be deleted

**Reproduces how often:** [Easily reproduced, Intermittent Issue]
Easily

cc: @danishjafri88

kjozwiak:
@jumde @joe looks like http://dustiest-limiation.000webhostapp.com/first_party.html doesn't exist anymore. Is there any other sites/STR that we can use to verify that this has been fixed?

**EDITED**:

There was a type in the original STR, links should be the following:

* http://dustiest-limitation.000webhostapp.com/first_party.html
* http://dustiest-limitation.000webhostapp.com/data.html

jumde:
> @jumde @joe looks like http://dustiest-limiation.000webhostapp.com/first_party.html doesn't exist anymore. Is there any other sites/STR that we can use to verify that this has been fixed?

@kjozwiak - Updated the issue, there was a typo: http://dustiest-limitation.000webhostapp.com/first_party.html

kjozwiak:
@jumde hmmm I don't think this is working :( Either the issue wasn't fixed, or the STR that we're using are not correct. Here's the STR that I'm following:

* Navigate to http://dustiest-limitation.000webhostapp.com/first_party.html
* Navigate to http://dustiest-limitation.000webhostapp.com/view.html
* ensure that `Cookies` display `stp=stp` & `Local Storage` displays `stp`
* Enable `Block All Cookies` via `Settings`
* Refresh http://dustiest-limitation.000webhostapp.com/view.html (`Cookies` & `Local Storage` should be blank)
* Disable `Block All Cookies` via `Settings`
* Refresh http://dustiest-limitation.000webhostapp.com/view.html (`Cookies` should display `stp=stp` & `Local Storage` should display `stp`)

[...]

PR #901, opened 2019-02-15
Fixes: #897 Cookies are now saved locally and are loaded back on toggle of block-all cookies.

danishjafri88:
To test this feature follow these steps:
1. https://www.w3schools.com/js/tryit.asp?filename=tryjs_cookie_username
2. On load it will request a username, which on save will return the username on next subsequent runs.
3. Enable Block all cookies.
4. After the page reloads it will not be able to return username even if you enter it again.
5. Disable Block all cookies
6. After page reloads it will return the username entered in 2.

## Submitter Checklist:

## Reviewer Checklist:

  - [x] `QA/(Yes|No)`
  - [ ] `bug` / `enhancement`


danishjafri88:
Logic to save and load cookies on block all cookies is done. Looking for a way to unit test this. Meanwhile this is blocked by #883

danishjafri88:
@jumde can you see if this works?
p.s session storage/cookies can't be salvaged.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2923,11 +2923,20 @@ extension BrowserViewController: PreferencesObserver {
             // All `block all cookies` toggle requires a hard reset of Webkit configuration.
             tabManager.reset()
             if !Preferences.Privacy.blockAllCookies.value {
-                tabManager.allTabs.forEach {
-                    if let url: URL = $0.webView?.url {
-                        $0.loadRequest(PrivilegedRequest(url: url) as URLRequest)
+                HTTPCookie.loadFromDisk { (_) in
```
iccub: You can remove brackets from `(_)` and just leave `_ in`, please recheck your code and do that for other occurrences in this PR

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2923,11 +2923,20 @@ extension BrowserViewController: PreferencesObserver {
...
+                    self.tabManager.reloadSelectedTab()
+                    self.tabManager.allTabs.forEach {
```
iccub: You could change it to `for tab in self.tabManager.allTabs where self.tabManager.selectedTab != tab {`
to remove the `if` inside of the loop. Up to you if you find your way more readable

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2923,11 +2923,20 @@ extension BrowserViewController: PreferencesObserver {
...
+                        //Creating all webviews since they are deleted in reset.
+                        if $0 != self.tabManager.selectedTab {
+                            $0.createWebview()
+                            if let url: URL = $0.webView?.url {
```
iccub: `: URL` can be removed

Client/Frontend/Settings/SettingsViewController.swift
```diff
@@ -225,25 +225,29 @@ class SettingsViewController: TableViewController {
...
-                        
-                        // TODO: Throw Alert to user to try again?
-                        let alert = UIAlertController(title: nil, message: Strings.Block_all_cookies_failed_alert_msg, preferredStyle: .alert)
-                        alert.addAction(UIAlertAction(title: Strings.OKString, style: .default))
-                        self.present(alert, animated: true)
+                        if success {
+                            Preferences.Privacy.blockAllCookies.value = status
+                        } else {
+                            //Revert the changes. Not handling success here to avoid a loop.
+                            _ = FileManager.default.setFolderAccess([
```
iccub: make `setFolderAccess()` `@discardableResult`

Client/HttpCookieExtension.swift
```diff
@@ -0,0 +1,84 @@
...
+         The network process is in suspened maybe?
+         
+         And for some reason fetch cookie records preemptively guarantees this works.
+         Best guess is that fetching records brings network process to active.
+         
+         Same applies to setting cookies back below.
+         */
+        WKWebsiteDataStore.default().fetchDataRecords(ofTypes: [WKWebsiteDataTypeCookies]) { (_) in}
+        cookieStore.getAllCookies { (cookies) in
+            let baseDir = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true)[0]
```
iccub: This will crash if there no `.documentDirectory`, I know it's not going to happen but we can make it more safely there.

I suggest using `URL` mechanism, also see my comment below
Use something like
 `guard let documentsDir = urls(for: .documentDirectory, in: .userDomainMask).first else { return nil }`  
to get URL of documents dir.

Even better will be creating a helper method for it, since we use it in few places
so
`var documentDirectoryUrl: URL? { return urls(for: .documentDirectory, in: .userDomainMask).first }`

Client/HttpCookieExtension.swift
```diff
@@ -0,0 +1,84 @@
...
+            do {
+                let data = try NSKeyedArchiver.archivedData(withRootObject: cookies, requiringSecureCoding: false)
+                try data.write(to: URL(fileURLWithPath: "\(baseDir)/\(filename)"))
```
iccub: Hardcoding user paths is scary as @jhreis said to me few times.

Now that we have a `URL` for documents dir you can append it like:
`let filenameUrl = documentDirectoryUrl.appendingPathComponent(filename)`
and then `data.write(to: fileNameUrl)`

Client/HttpCookieExtension.swift
```diff
@@ -0,0 +1,84 @@
...
+                log.error("Failed to write cookies to disk with error: \(error)")
+                completion?(false)
+                return
+            }
+            completion?(true)
+        }
+    }
+    
+    class func loadFromDisk(_ filename: String = "CookiesData.json", completion: ((Success) -> Void)? = nil) {
+        let baseDir = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true)[0]
```
iccub: see `saveToDisk` comments

Client/HttpCookieExtension.swift
```diff
@@ -0,0 +1,84 @@
...
+        do {
+            let data = try Data(contentsOf: URL(fileURLWithPath: "\(baseDir)/\(filename)"), options: Data.ReadingOptions.alwaysMapped)
```
iccub: see `saveToDisk` comments, use appending component instead of hardcoding file path

Client/HttpCookieExtension.swift
```diff
@@ -0,0 +1,84 @@
...
+            if let cookies: [HTTPCookie] = try NSKeyedUnarchiver.unarchiveTopLevelObjectWithData(data) as? [HTTPCookie] {
```
iccub: `: [HTTPCookie]` can be removed

Client/HttpCookieExtension.swift
```diff
@@ -0,0 +1,84 @@
...
+import Foundation
+import WebKit
+import Shared
+
+private let log = Logger.browserLogger
+
+public extension HTTPCookie {
+    
+    typealias Success = Bool
+    class func saveToDisk(_ filename: String = "CookiesData.json", completion: ((Success) -> Void)? = nil) {
```
iccub: We should hardcode `CookiesData.json` string somewhere, it's used in few places

Client/HttpCookieExtension.swift
```diff
@@ -0,0 +1,84 @@
...
+        })
+        
+        dispatchGroup.notify(queue: .main) {
+            completion?(true)
+        }
+    }
+    
+    class func deleteLocalCookieFile(_ filename: String = "CookiesData.json") {
+        let baseDir = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true)[0]
+        let url: URL = URL(fileURLWithPath: "\(baseDir)/\(filename)")
```
iccub: see `saveToDisk` comments

Client/HttpCookieExtension.swift
```diff
@@ -0,0 +1,84 @@
...
+            log.error("Failed to load cookies from disk with error: \(error)")
+            completion?(false)
+        }
+    }
+    
+    private class func setCookies(_ cookies: [HTTPCookie], completion: ((Success) -> Void)?) {
+        let cookieStore = WKWebsiteDataStore.default().httpCookieStore
+        // For the purpose of the line below read the comment in saveCookies (in this same commit)
+        WKWebsiteDataStore.default().fetchDataRecords(ofTypes: [WKWebsiteDataTypeCookies]) { (_) in}
+        let dispatchGroup = DispatchGroup()
```
iccub, line 74: What's your opinion on using `DispatchGroup` vs `Deferred` library here?
Your code is fine, I just wonder whether we should try to stick with one way to handle synchronization code like this

iccub (changes requested):
Mainly need changes around files and directory access. Logic seems fine, I will do more check once code style review is finished

Client/Frontend/Settings/SettingsViewController.swift
danishjafri88: It was already `@discardableResult ` forgot to remove this escape. :)

Client/HttpCookieExtension.swift
danishjafri88: We cant work with URL as all the methods use Strings and all the getters return strings. Going ahead with putting a guard to make sure it doesn't crash.

Client/HttpCookieExtension.swift
danishjafri88: Cant append Component to String and cant use URL here until this point.

Client/HttpCookieExtension.swift
danishjafri88: Though `Deferred` is just a wrapper over dispatchGroups aka semaphores, in this particular we shouldn't use it because 
1. There can be a large number of cookies, and the extra function stack is not appreciable.
2. I want to keep it as native as possible.
3. Deferred does not provide a very out of the box solution for these cases. 
`private class func setCookies(_ cookies: [HTTPCookie]) -> Deferred<[Bool]> {
        let cookieStore = WKWebsiteDataStore.default().httpCookieStore
        // For the purpose of the line below read the comment in saveCookies (in this same commit)
        WKWebsiteDataStore.default().fetchDataRecords(ofTypes: [WKWebsiteDataTypeCookies]) { (_) in}
        var deferreds: [Deferred<Bool>] = []
        cookies.forEach({
            let deferred = Deferred<Bool>()
            deferreds.append(deferred)
            cookieStore.setCookie($0, completionHandler: {deferred.fill(true)})
        })
        return all(deferreds)
    }` 

plus the code change required at consumer side.  This does not improve the readability neither the lines of code. 
It basically performance required vs style.

Client/Frontend/Settings/SettingsViewController.swift
```diff
@@ -225,25 +225,29 @@ class SettingsViewController: TableViewController {
...
-                        (.webSiteData, status)
-                        ])
-                    if success {
-                        Preferences.Privacy.blockAllCookies.value = status
-                    } else {
-                        //Revert the changes. Not handling success here to avoid a loop.
-                        _ = FileManager.default.setFolderAccess([
-                            (.cookie, false),
-                            (.webSiteData, false)
+                    let completionBlock: (Bool) -> Void = {(_) in
```
iccub: `(_) in` -> `_ in`

Client/HttpCookieExtension.swift
```diff
@@ -0,0 +1,93 @@
...
+         A possibility is https://bugs.webkit.org/show_bug.cgi?id=188242
+         Even with the issue being fixed it still sometimes doesn't work.
+         The network process is in suspened maybe?
+         
+         And for some reason fetch cookie records preemptively guarantees this works.
+         Best guess is that fetching records brings network process to active.
+         
+         Same applies to setting cookies back below.
+         */
+        WKWebsiteDataStore.default().fetchDataRecords(ofTypes: [WKWebsiteDataTypeCookies]) { (_) in}
```
iccub: `(_) in` -> `_ in`

Client/HttpCookieExtension.swift
```diff
@@ -0,0 +1,93 @@
...
+        cookieStore.getAllCookies { (cookies) in
+            if let baseDir = FileManager.documentDirectoryURL {
```
iccub: What about replacing it with `guard` to remove one level of indentation? And do the same in rest of the methods in the file

Client/HttpCookieExtension.swift
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+    }
+    
+    class func loadFromDisk(_ filename: String = HTTPCookie.locallySavedFile, completion: ((Success) -> Void)? = nil) {
+        if let baseDir = FileManager.documentDirectoryURL {
+            do {
+                let data = try Data(contentsOf: baseDir.appendingPathComponent(filename), options: Data.ReadingOptions.alwaysMapped)
+                if let cookies = try NSKeyedUnarchiver.unarchiveTopLevelObjectWithData(data) as? [HTTPCookie] {
+                    HTTPCookie.setCookies(cookies) { (success) in
```
iccub: `(success) in` -> `success in`

Client/HttpCookieExtension.swift
```diff
@@ -0,0 +1,93 @@
...
+                }
+                completion?(true)
+            } else {
+                completion?(false)
+            }
+        }
+    }
+    
+    class func loadFromDisk(_ filename: String = HTTPCookie.locallySavedFile, completion: ((Success) -> Void)? = nil) {
+        if let baseDir = FileManager.documentDirectoryURL {
```
iccub: consider guard

Client/HttpCookieExtension.swift
```diff
@@ -0,0 +1,93 @@
...
+        cookies.forEach({
+            dispatchGroup.enter()
+            cookieStore.setCookie($0, completionHandler: {dispatchGroup.leave()})
+        })
+        dispatchGroup.notify(queue: .main) {
+            completion?(true)
+        }
+    }
+    
+    class func deleteLocalCookieFile(_ filename: String = "CookiesData.json") {
```
iccub: Use constant for `filename` default param

Client/HttpCookieExtension.swift
```diff
@@ -0,0 +1,93 @@
...
+        if let baseDir = FileManager.documentDirectoryURL {
+            let url: URL = URL(fileURLWithPath: "\(baseDir)/\(filename)")
```
iccub: Use `appendingPathComponent`

Client/HttpCookieExtension.swift
```diff
@@ -0,0 +1,93 @@
...
+            dispatchGroup.enter()
+            cookieStore.setCookie($0, completionHandler: {dispatchGroup.leave()})
+        })
+        dispatchGroup.notify(queue: .main) {
+            completion?(true)
+        }
+    }
+    
+    class func deleteLocalCookieFile(_ filename: String = "CookiesData.json") {
+        if let baseDir = FileManager.documentDirectoryURL {
```
iccub: consider guard

ClientTests/HttpCookieExtensionTest.swift
```diff
@@ -0,0 +1,42 @@
...
+        let expectation = XCTestExpectation(description: "Cookie is loaded from file")
+        
+        let properties: [HTTPCookiePropertyKey: Any] = [HTTPCookiePropertyKey.name : "BraveCookieTest",
+                                            HTTPCookiePropertyKey.domain: "brave.com",
+                                            .path: "/",
+                                            .value: "something",
+                                            .secure: "TRUE",
+                                            .expires: NSDate(timeIntervalSinceNow: 30)]
+        if let cookie = HTTPCookie(properties: properties) {
+            WKWebsiteDataStore.default().fetchDataRecords(ofTypes: [WKWebsiteDataTypeCookies]) { (_) in}
```
iccub: replace `(_) in` to `_ in` in this method

ClientTests/HttpCookieExtensionTest.swift
```diff
@@ -0,0 +1,42 @@
...
+            WKWebsiteDataStore.default().httpCookieStore.setCookie(cookie) {
+                HTTPCookie.saveToDisk(completion: { (_) in
+                    WKWebsiteDataStore.default().fetchDataRecords(ofTypes: [WKWebsiteDataTypeCookies]) { (_) in}
+                    WKWebsiteDataStore.default().httpCookieStore.delete(cookie, completionHandler: {
+                        HTTPCookie.loadFromDisk(completion: { (_) in
+                            WKWebsiteDataStore.default().httpCookieStore.getAllCookies({ (cookies) in
```
iccub: Remove brackets from `cookies`

iccub (changes requested):

iccub (approved):

iccub:
I did code review and it looks good to me.
@jumde would you mind testing it with the STR you provided here https://github.com/brave/brave-ios/issues/897?

jumde:
@iccub @danishjafri88 - Tried the steps listed in #897:

```
Steps to Reproduce
Navigate to http://dustiest-limiation.000webhostapp.com/first_party.html - sets all storage
Navigate to http://dustiest-limiation.000webhostapp.com/data.html - no storage api calls here
Turn on Block all cookies. Using the Safari Inspector verify the cookies and storage is blocked.
Turn off Block all cookies - Cookies and Session Storage is deleted.
```

I can't see the cookies after the `Block All Cookies` is turned off. Let me know if I'm missing something.

danishjafri88:
> @iccub @danishjafri88 - Tried the steps listed in #897:
> 
> ```
> Steps to Reproduce
> Navigate to http://dustiest-limiation.000webhostapp.com/first_party.html - sets all storage
> Navigate to http://dustiest-limiation.000webhostapp.com/data.html - no storage api calls here
> Turn on Block all cookies. Using the Safari Inspector verify the cookies and storage is blocked.
> Turn off Block all cookies - Cookies and Session Storage is deleted.
> ```
> I can't see the cookies after the `Block All Cookies` is turned off. Let me know if I'm missing something.
@jumde 
It's because the cookie set is session only. Session cookies cannot be salvaged in any possible way. To test this please follow steps in this pr's description.
Screenshot 2019-03-13 at 12 54 53 PM

jumde:
@danishjafri88 oh yeah, missed that. Thanks! All good, tested on gmail. The user is logged in after `Block all cookies` is toggled on.

Client/HttpCookieExtension.swift
```diff
@@ -0,0 +1,95 @@
...
+    }
+    
+    private class func setCookies(_ cookies: [HTTPCookie], completion: ((Success) -> Void)?) {
+        let cookieStore = WKWebsiteDataStore.default().httpCookieStore
+        // For the purpose of the line below read the comment in saveCookies (in this same commit)
+        WKWebsiteDataStore.default().fetchDataRecords(ofTypes: [WKWebsiteDataTypeCookies]) { _ in}
+        let dispatchGroup = DispatchGroup()
+        cookies.forEach({
+            dispatchGroup.enter()
+            cookieStore.setCookie($0, completionHandler: {dispatchGroup.leave()})
```
jhreis: Do not need the `{}` block wrapper here 
```suggestion
            cookieStore.setCookie($0, completionHandler: dispatchGroup.leave)
```

Client/FileManagerExtension.swift
```diff
@@ -31,7 +37,10 @@ public extension FileManager {
     
     // Check the locked status of a folder. Returns true for locked.
     public func checkLockedStatus(folder: Folder) -> Bool {
-        let baseDir = NSSearchPathForDirectoriesInDomains(.libraryDirectory, .userDomainMask, true)[0]
+        guard let baseDir = NSSearchPathForDirectoriesInDomains(.libraryDirectory, .userDomainMask, true).first else {
```
jhreis: This same logic as https://github.com/brave/brave-ios/pull/901/files#diff-9a7ffe3be300e023d1ceadfbae9ea388R24

Might be good to create a function that returns this result to avoid duplication.

Client/Frontend/Settings/SettingsViewController.swift
jhreis:
```suggestion
                    let completionBlock: Bool -> Void = { _ in
```
(remove two extraneous sets of `()`s

jhreis (changes requested):

jhreis:
Confirmed with @jumde that above comment was formal sign off. Sec review passed.

jhreis (approved):

PR #983, opened 2019-03-18
Added logic to download opensearch xml and load engine

danishjafri88:
This pr includes code to allow user add a custom search engine. This is possible in two ways:
1. Add OpenSearch engine
    - From BVC by selecting a search field, an option to add search engine will appear above the keyboard.
    - From search settings while adding a search engine, if the site supports OpenSearch and option to add it will be provided in the header.

2. Manually
    User can add an engine manually in settings by entering a URL with formatted search query param and a title.

Simulator Screen Shot - iPhone X - 2019-03-29 at 22 36 52
Simulator Screen Shot - iPhone X - 2019-03-29 at 22 36 58
Simulator Screen Shot - iPhone X - 2019-03-29 at 22 37 00
Simulator Screen Shot - iPhone X - 2019-03-29 at 22 37 24
Simulator Screen Shot - iPhone X - 2019-03-29 at 22 37 26
Simulator Screen Shot - iPhone X - 2019-03-29 at 22 37 48
Simulator Screen Shot - iPhone X - 2019-03-29 at 22 37 51

## Submitter Checklist:

## Reviewer Checklist:

  - [ ] `QA/(Yes|No)`
  - [ ] `bug` / `enhancement`


Shared/Extensions/StringExtensions.swift
```diff
@@ -109,4 +109,15 @@ public extension String {
         addWordsDescriptionBolded.setAttributes(attributes, range: nsRangeOfBoldedText)
         return addWordsDescriptionBolded
     }
+    
+    public func jsonObject() -> Any? {
+        do {
+            if let data = self.data(using: .utf8) {
+                return try JSONSerialization.jsonObject(with: data, options: .mutableContainers)
+            }
+        } catch {}
```
JonKaleidoscope: Since you are not using the catch you could refactor this function to use a `try?` If an error is thrown it will just return `nil`

```swift
guard let data = self.data(using: .utf8) else { return nil }
return try? JSONSerialization.jsonObject(with: data, options: .mutableContainers)
```

Shared/Extensions/StringExtensions.swift
```diff
@@ -109,4 +109,15 @@ public extension String {
         addWordsDescriptionBolded.setAttributes(attributes, range: nsRangeOfBoldedText)
         return addWordsDescriptionBolded
     }
+    
+    public func jsonObject() -> Any? {
+        do {
+            if let data = self.data(using: .utf8) {
+                return try JSONSerialization.jsonObject(with: data, options: .mutableContainers)
```
JonKaleidoscope: I think you will be fine without the `.mutableContainers` since you are calling this function in swift. The containers are mutable by default.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2592,52 +2593,51 @@ extension BrowserViewController: HistoryStateHelperDelegate {
...
-                make.width.equalTo(inputView.snp.height)
-                make.top.equalTo(nextButtonView.snp.top)
-                make.height.equalTo(inputView.snp.height)
+        self.openSearchLinkDict = nil
+        webView.evaluateJavaScript(OpenSearchEngine.fetchOpenSearchLinkScript) { (result, _) in
+            if let dict = (result as? String)?.jsonObject() as? [String: String] {
+                //check if the search engine has already been added.
+                self.openSearchLinkDict = dict
+                let shortName = dict["title"] ?? ""
+                let domain =
```
JonKaleidoscope: Is there no value for domain?

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2592,52 +2593,51 @@ extension BrowserViewController: HistoryStateHelperDelegate {
...
-            }
-            inputView.addSubview(self.customSearchEngineButton)
-            self.customSearchEngineButton.snp.remakeConstraints { make in
-                make.leading.equalTo(nextButtonView.snp.trailing).offset(20)
-                make.width.equalTo(inputView.snp.height)
-                make.top.equalTo(nextButtonView.snp.top)
-                make.height.equalTo(inputView.snp.height)
+        self.openSearchLinkDict = nil
+        webView.evaluateJavaScript(OpenSearchEngine.fetchOpenSearchLinkScript) { (result, _) in
+            if let dict = (result as? String)?.jsonObject() as? [String: String] {
```
JonKaleidoscope, line 2791: I think a `guard let` might be better served here. 

Also you might get away with using a JSON Decoder here since strings automatically conform to `Cobale`.
```swift
let decoder = JSONDecoder()
guard let dict = try? decoder.decode([String: String].self, from: result) else { return }
```

Shared/Extensions/StringExtensions.swift
danishjafri88: Hi, thanks reviewing this pr. Unfortunately this pr is still WIP and in draft, there are lot more changes that will be pushed out soon. 
In this case I have added a logger, hence we need to keep the implementation.

Shared/Extensions/StringExtensions.swift
danishjafri88: Updated thanks!

Client/Frontend/Browser/BrowserViewController.swift
danishjafri88: Hehe well it was committed in a hurry, this is history now.

Client/Frontend/Browser/BrowserViewController.swift
danishjafri88: The problem is `result` is of type `Any` while the expected type is String(the script returns a string). The Decoder requires a `Data` type. Hence the helper function. Since the function is in String Extension I cannot hardcode it to `[String: String]` type.

JonKaleidoscope: Fair enough.

Shared/Extensions/StringExtensions.swift
JonKaleidoscope: Gotcha

Client/Utils/FaviconFetcher.swift
```diff
@@ -147,26 +146,43 @@ open class FaviconFetcher: NSObject, XMLParserDelegate {
...
-
-                if let iconUrl = NSURL(string: href, relativeTo: url as URL), let absoluteString = iconUrl.absoluteString {
-                    let icon = Favicon(url: absoluteString)
-                    icons = [icon]
-                }
-
-                // If we haven't got any options icons, then use the default at the root of the domain.
-                if let url = NSURL(string: "/favicon.ico", relativeTo: url as URL), icons.isEmpty, let absoluteString = url.absoluteString {
+                
+                // SVG is not supported right not
```
danishjafri88, line 170: Are SVG images supported or am I missing something? Should we add support for svg icons. Maybe in a new PR?

jhreis:
When this is merged, please close https://github.com/brave/brave-ios/issues/1030

jhreis:
Closing, not deleting branch, will revisit when re-prioritized.

PR #1015, opened 2019-03-28
Ref #1010: Fix crash when moving Bookmark with sync from root level.

iccub:
## Submitter Checklist:

## Reviewer Checklist:

  - [ ] `QA/(Yes|No)`
  - [ ] `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Menu/BookmarksViewController.swift
+++ b/Client/Frontend/Menu/BookmarksViewController.swift
@@ -331,6 +331,12 @@ class BookmarksViewController: SiteTableViewController {
   
   fileprivate func configureCell(_ cell: UITableViewCell, atIndexPath indexPath: IndexPath) {
     
+    // Make sure Bookmark at index path exists,
+    // `frc.object(at:)` crashes otherwise, doesn't fail safely with nil
+    if let objectsCount = bookmarksFRC?.fetchedObjects?.count, indexPath.row >= objectsCount {
+        fatalError("Bookmarks FRC index out of bounds")
+    }
+    
     guard let item = bookmarksFRC?.object(at: indexPath) else { return }
     cell.tag = item.objectID.hashValue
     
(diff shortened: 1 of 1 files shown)


Client/Frontend/Menu/BookmarksViewController.swift
```diff
@@ -580,8 +582,14 @@ extension BookmarksViewController: NSFetchedResultsControllerDelegate {
     switch type {
     case .update:
         let update = { (path: IndexPath?) in
-            if let path = path, let cell = self.tableView.cellForRow(at: path) {
-                self.configureCell(cell, atIndexPath: path)
+            if let path = path, let cell = self.tableView.cellForRow(at: path),
+                let fetchedObjectsCount = self.bookmarksFRC?.fetchedObjects?.count {
```
iccub: I can place the out of bounds checker either here or in `configureCell()`.
Choose to place it here in case we want to redirect this update to `tableView.delete()` for example(not needed for now)

jhreis: I think here is good, I would also throw a `fatalError` inside of `configureCell` honestly. That should never be called, and should just fail aggressively IMO.

jhreis (changes requested):

Client/Frontend/Menu/BookmarksViewController.swift
```diff
@@ -580,8 +582,14 @@ extension BookmarksViewController: NSFetchedResultsControllerDelegate {
...
+                // When Bookmark is moved to another folder, it can be interpreted as update action
+                // (since the object is not deleted but updated to have a different parent Bookmark)
+                // Make sure we are not out of bounds here.
+                if path.row < fetchedObjectsCount {
```
jhreis: I think add this if-check when you unwrap `fetchedObjectsCount`:

```
let fetchedCount = self.bookmarksFRC?.fetchedCount?.count, path.row < fetchedObjectsCount
```

Client/Frontend/Menu/BookmarksViewController.swift
```diff
@@ -331,6 +331,12 @@ class BookmarksViewController: SiteTableViewController {
   
   fileprivate func configureCell(_ cell: UITableViewCell, atIndexPath indexPath: IndexPath) {
     
+    // Make sure Bookmark at index path exists,
+    // `frc.object(at:)` crashes otherwise, doesn't fail safely with nil
+    if let objectsCount = bookmarksFRC?.fetchedObjects?.count, indexPath.row >= objectsCount {
+        fatalError() // fail aggressively
```
jhreis: You can remove this comment, the fatalError should be enough, throw in a reason though:
`"Bookmarks FRC index out of bounds"`

jhreis (approved):

Issue #760, opened 2019-01-18
Do Not Create `Domain` Objects while in Private Mode


jhreis:
Currently `getOrCreateForUrl` is called while in private mode to help keep logic clean between modes. However, this creates a database entry for urls visited in private mode. No Domain objects should be written to the database while in Private Browsing.

The exception to this is user-requested saves, which are currently if the user saves a bookmark or new favorite. In these instances, it is fine to store the associated `Domain` object.

PR #1024, opened 2019-04-02
Fixed: #760. Added an InMemory Context to keep domain non persistent in PBM

danishjafri88:
Domain were saved to disk even in private mode. To solve this a new inMemory context is used to create domain in memory only. Doing this solves the issue and also changes the following
- Since Domains in normal and private mode have separate shields override, the `PrivateBrowsingShieldOverride ` is redundant now. 
- Bookmarks will be saved in Normal mode and opening it will trigger same check of private browsing and the domain will be recreated in inMemory.

## Submitter Checklist:

## Reviewer Checklist:

  - [x] `QA/(Yes|No)`
  - [x] `bug` / `enhancement`


Data/models/Domain.swift
```diff
@@ -25,8 +25,9 @@ public final class Domain: NSManagedObject, CRUD {
     
     // MARK: - Public interface
     
-    public class func getOrCreate(forUrl url: URL) -> Domain {
-        return getOrCreateInternal(url)
+    public class func getOrCreate(forUrl url: URL, persistent: Bool) -> Domain {
+        let context = persistent ? DataController.viewContext : DataController.viewContextInMemory
+        return getOrCreateInternal(url, context: context, save: true)
```
iccub, line 34: What will happen if we add a bookmark in PBM? The Bookmark will be saved to regular data store, but its children Domain will be in in-memory store, won't that crash or something?

danishjafri88: Nope I tried it out. When you add a bookmark, it is added in regular data store. But when you try to open it there is no such domain in memory so then everything switches to inMemeory context.

Data/models/CRUDProtocols.swift
```diff
@@ -35,7 +35,7 @@ extension Deletable where Self: NSManagedObject {
     }
     
     static func deleteAll(predicate: NSPredicate? = nil,
-                          context: WriteContext = .new,
+                          context: WriteContext = .new(false),
```
iccub: We should use a named parameter like `.new(inMemory: false)`, `false` alone doesn't say much

iccub (changes requested):

iccub (approved):

iccub (approved):

PR #1026, opened 2019-04-02
Fix #195  Added Migration code for sqlite files from Document to Application Support folder

### LINKED ISSUE #195: Brave.sqlite database should not be accessible from the Documents folder
Due to `UIFileSharingEnabled` being set to `YES` in the `info.plist` files stored in the Documents directory are accessible using iTunes (even if the device is locked) once the device has been connected to a desktop computer and the device has been trusted. Once a device is trusted the only way to remove that trust is via General > Reset > Reset Location & Privacy on the device.

[iOS Storage Best Practices](https://developer.apple.com/videos/play/tech-talks/204/) recommends storing user-visible content in the Documents directory and non user-visible content such as databases in the Application Support directory.

The Application Support directory can be hierarchical, is persistent, is included in iCloud and iTunes Backups unless files are marked with `NSURLIsExcludedFromBackupKey`, and disk space is reported in the Storage Settings UI under the "Documents & Data" total.

iOS does not delete the Application Support directory if the application is offloaded.

## Test Plan

1. Install Brave (or use an existing installation)
1. Visit a couple websites (just so there is history in the database)
1. Exit Brave
1. To into the Files app on device
1. See if there are any files named `Brave.sqlite` or `Database`, etc... anything that looks remotely suspicious. There may be a `downloads` folder (this is okay)

danishjafri88:
Sqlite file is now migrated to the new folder. There two questions that need to be answered/invetigated before merge.
1. How will this function when there is data model version change(ie DB migration).
2. Before we allow iTunesFileSharing, we need to make sure all contents of Documents Directory are allowed for user visibility.
Fore Example: `https-everywhere-data` folder and Locally saved cookies(can be moved to Application support).

## Submitter Checklist:

## Test Plan:
We have to test if the app upgrades correctly so:
- install previous app version, add few bookmarks, visit some websites so history is filled
- upgrade to version with this PR merged
- verify it doesn't crash, and Bookmarks/History are staying

### Screenshots:
Before:
Screenshot 2019-07-25 at 12 22 59 PM
After:
Screenshot 2019-07-25 at 12 25 03 PM

## Reviewer Checklist:

  - [ ] `QA/(Yes|No)`
  - [ ] `bug` / `enhancement`


Client/Application/AppDelegate.swift
```diff
@@ -64,6 +64,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
         
         HttpsEverywhereStats.shared.startLoading()
         
+        //Moves Coredata sqlite file from Documents dir to application support dir.
+        DataController.shared.migrateToNewPathIfNeeded()
```
jhreis: We should probably move this into the migration, and just let the migration call below handle this if possible. Is there a timing issue or something?

Best not to require different migrations to happen at different places.

jhreis (dismissed):

Data/models/DataController.swift
```diff
@@ -137,7 +163,7 @@ public class DataController: NSObject {
         return queue
     }()
     
-    private let storeURL: URL = {
+    private let oldStoreURL: URL = {
```
iccub, line 170: Can we add a `@deprecated` annotation here? Not sure if it's possible.
Something that this old store is deprecated, and please use new `storeUrl` or something

Data/models/DataController.swift
```diff
@@ -162,3 +198,10 @@ public class DataController: NSObject {
         return backgroundContext
     }
 }
+
+fileprivate extension URL {
+    
+    func sqliteFiles(dbName: String) throws -> [URL] {
+        return try FileManager.default.contentsOfDirectory(at: self, includingPropertiesForKeys: nil, options: []).filter({$0.lastPathComponent.hasPrefix(dbName)})
```
iccub: Won't that retrieve all files, not only with .sqlite extension?

danishjafri88: yes thats intended. I want to retrieve all files that have a name prefix `Brave.sqlite` like `Brave.sqlite-wal` etc. I think keeping this function in an extension may not be the best way, so moving it into the migration method itself.

Client/Application/AppDelegate.swift
danishjafri88: So I need the db migration to run before any db access. If I can enforce that no one will ever write anything DB before migration code then we can move it there. Any ideas we can do that?

Data/models/DataController.swift
danishjafri88: It will leave the code with a single warning ie when we are using it for migration. If that is ok I can push it

iccub: right.. i'm fine with just adding a comment here, without deprecation notice

Client/Application/AppDelegate.swift
jhreis: Unless there are issues right now, I think this is relatively safe to assume. Migrations should generally always happen before DB access, and hasn't been an issue this far (famous last words, right?). 

Also, it would only matter if we accessed the DB before someone had this migration take place, so after a couple months, this migration will likely not be very useful/important.

If we have genuine concerns about this, we should consider moving the entire migration process up.

Client/Application/AppDelegate.swift
```diff
@@ -64,8 +64,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
         
         HttpsEverywhereStats.shared.startLoading()
         
+
```
iccub: Remove

iccub (changes requested):
Looking good

iccub (approved):

Client/Application/AppDelegate.swift
danishjafri88: As discussed in DM, The placement of the db migration is curcial because of ` !DataController.shared.storeExists() ` check. Create a facade for the DB migration in Migration Code though.

PR #1058, opened 2019-04-17
Menu 2.0

iccub:
## Submitter Checklist:

## Reviewer Checklist:

  - [x] `QA/(Yes|No)`
  - [x] `bug` / `enhancement`


Client/Frontend/Browser/Toolbars/BottomToolbar/BottomToolbarView.swift
```diff
@@ -0,0 +1,137 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import UIKit
+import SnapKit
+import Shared
+import BraveShared
+
+class BottomToolbarView: UIView {
```
iccub, line 10: No changes here besides adding `menuButton`. Moved this class to its own file

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/ToolbarUrlActionsProtocol.swift
```diff
@@ -0,0 +1,70 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import UIKit
+import Shared
+
+protocol ToolbarUrlActionsProtocol where Self: UIViewController {
```
iccub, line 8: No changes here, moved it to its own file

Client/Frontend/Browser/Toolbars/ToolbarButton.swift
```diff
@@ -0,0 +1,53 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import UIKit
+
+class ToolbarButton: UIButton {
```
iccub, line 7: No changes here, moved it to its own file

Client/Frontend/Browser/Toolbars/ToolbarHelper.swift
```diff
@@ -0,0 +1,97 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import UIKit
+import Shared
+
+@objcMembers
+class ToolbarHelper: NSObject {
```
iccub, line 9: No changes here besides action for the new `menu`, moved it to its own file

Client/Frontend/Browser/Toolbars/ToolbarProtocols.swift
```diff
@@ -0,0 +1,35 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+
+protocol ToolbarProtocol: class {
```
iccub, line 7: No changes here besides actions for the new menu and removing params from `tabToolbarDidPressShare`

Client/Frontend/Browser/Toolbars/UrlBar/LocationContainerView.swift
```diff
@@ -0,0 +1,25 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import UIKit
+
+// We need a subclass so we can setup the shadows correctly
+// This subclass creates a strong shadow on the URLBar
+class LocationContainerView: UIView {
```
iccub, line 9: No changes here, moved to its own file

Client/Frontend/Browser/Toolbars/UrlBar/UrlBarTextField.swift
```diff
@@ -0,0 +1,85 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import UIKit
+import BraveShared
+
+class UrlBarTextField: AutocompleteTextField {
```
iccub, line 8: No changes here, moved to its own file

Data/models/Bookmark.swift
```diff
@@ -459,7 +523,9 @@ extension Bookmark {
             if isFolder {
                 removeFolderAndSendSyncRecords(uuid: syncUUID)
             } else {
-                Sync.shared.sendSyncRecords(action: .delete, records: [self])
+                DataController.perform(context: context) { context in
+                    Sync.shared.sendSyncRecords(action: .delete, records: [self], context: context)
```
iccub, line 518: A bit out of scope of this PR but this was throwing threading errors so fixed it

Client/Extensions/UIViewExtensions.swift
```diff
@@ -99,4 +99,15 @@ extension UIView {
         }
         return spacer
     }
+    
+    /// Returns a line with height of 1pt. Used to imitate a separator line in custom views.
+    static var separatorLine: UIView {
+        let view = UIView().then {
+        $0.backgroundColor = UIColor(white: 0.0, alpha: 0.2)
```
jhreis: Indent.

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/AddEditHeaderView.swift
```diff
@@ -0,0 +1,34 @@
...
+    let mainStackView = UIStackView().then {
+        $0.axis = .vertical
+        $0.spacing = UX.defaultSpacing
+    }
+    
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        
+        backgroundColor = .white
+        addSubview(mainStackView)
```
jhreis, line 27: Does this consume the entire UI (e.g. 100% width/height)?

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/BookmarkDetailsView.swift
```diff
@@ -0,0 +1,100 @@
...
+    
+    weak var delegate: BookmarkDetailsViewDelegate?
+    
+    let titleTextField = UITextField().then {
+        $0.placeholder = Strings.BookmarkTitlePlaceholderText
+        $0.clearButtonMode = .whileEditing
+        $0.translatesAutoresizingMaskIntoConstraints = false
+    }
+    
+    let urlTextField: UITextField? = UITextField().then {
```
jhreis, line 21: Don't think this needs to be optional, does it? We can then remove the [this check](https://github.com/brave/brave-ios/pull/1058/files?file-filters%5B%5D=.png&file-filters%5B%5D=.swift&hide-deleted-files=true#diff-47cd8652ccb1ab97a854f88732106494R57).

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/BookmarkDetailsView.swift
```diff
@@ -0,0 +1,100 @@
...
+        $0.placeholder = Strings.BookmarkUrlPlaceholderText
+        $0.keyboardType = .webSearch
```
jhreis: I think `.URL` would be better.

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/BookmarkDetailsView.swift
```diff
@@ -0,0 +1,100 @@
...
+    
+    // MARK: - Initialization
+    
+    convenience init(title: String?, url: String?) {
+        self.init(frame: .zero)
+        
+        guard let urlTextField = urlTextField else { fatalError("Url text field must be set up") }
+        
+        [UIView.separatorLine, contentStackView, UIView.separatorLine].forEach {
+            mainStackView.addArrangedSubview($0)
```
jhreis: Since we are literally just passing in the array items, we can just give the actual function itself:

```
[UIView.separatorLine, contentStackView, UIView.separatorLine].forEach(mainStackView.addArrangedSubview)
```
(no need to create an anonymous closure, to execute a function here)

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/BookmarkDetailsView.swift
```diff
@@ -0,0 +1,100 @@
...
+        }
+        
+        if let url = url, let favUrl = URL(string: url) {
+            faviconImageView.setIcon(nil, forURL: favUrl)
+        }
+        
+        titleTextField.text = title ?? Strings.NewBookmarkDefaultName
+        urlTextField.text = url ?? Strings.NewFolderDefaultName
+        
+        mainStackView.snp.makeConstraints {
```
jhreis: Shouldn't this be setup in the root-class VC? Or does this change depending on the subclass?

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/BookmarkFormFieldsProtocol.swift
```diff
@@ -0,0 +1,40 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import Shared
+
+protocol BookmarkFormFieldsProtocol where Self: UIView {
```
jhreis: I think this is a bit more clear ([and safer](https://developer.apple.com/documentation/xcode_release_notes/xcode_10_2_release_notes/swift_5_release_notes_for_xcode_10_2) apparently): 
```suggestion
protocol BookmarkFormFieldsProtocol: UIView {
```

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/BookmarkFormFieldsProtocol.swift
```diff
@@ -0,0 +1,40 @@
...
+    var titleTextField: UITextField { get }
+    /// Nil by default
```
jhreis: Comment not needed, this should be inferred by the `?` IMO, and also could depend on the subclass implementation.

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/BookmarkFormFieldsProtocol.swift
```diff
@@ -0,0 +1,40 @@
...
+    func validateFields() -> Bool
+}
+
+extension BookmarkFormFieldsProtocol {
+    var urlTextField: UITextField? { return nil }
+    
+    func validateFields() -> Bool {
+        // Only title field is implemented
+        if urlTextField == nil {
+            guard let titleText = titleTextField.text else { return false }
```
jhreis: Given the logic, this should probably be moved to the top of the function, as it is executed regardless of the `urlTextField` condition ([see](https://github.com/brave/brave-ios/pull/1058/files?file-filters%5B%5D=.png&file-filters%5B%5D=.swift&hide-deleted-files=true#diff-67509d7840a6d32baf837c30d41c78dfR28)).

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/BookmarkFormFieldsProtocol.swift
```diff
@@ -0,0 +1,40 @@
...
+            return validateTitle(titleText)
+        }
+        
+        guard let title = titleTextField.text, let url = urlTextField?.text else { return false }
```
jhreis: `guard let title = titleTextField.text` can be removed once pushed to top of function.

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/BookmarkFormFieldsProtocol.swift
```diff
@@ -0,0 +1,40 @@
...
+        
+        return validateTitle(title) && validateUrl(url)
+    }
+    
+    private func validateTitle(_ title: String) -> Bool {
```
jhreis: If this took an optional `title` (`title: String?)` it could make the logic from the caller a bit more clear, but up to you.

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/FolderDetailsView.swift
```diff
@@ -0,0 +1,44 @@
...
+        
+        let spacerView = UIView(frame: CGRect(x: 0, y: 0, width: 10, height: 16))
+        $0.leftViewMode = .always
+        $0.leftView = spacerView
+    }
+    
+    convenience init(title: String?, viewHeight: CGFloat) {
+        self.init(frame: .zero)
+        
+        [UIView.separatorLine, titleTextField, UIView.separatorLine].forEach {
```
jhreis:
```
    [UIView.separatorLine, titleTextField, UIView.separatorLine].forEach(mainStackView.addArrangedSubview)
```

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/IndentedImageTableViewCell.swift
```diff
@@ -0,0 +1,87 @@
...
+        $0.contentMode = .scaleAspectFit
+        $0.setContentCompressionResistancePriority(.defaultHigh, for: .horizontal)
+        $0.setContentHuggingPriority(.defaultHigh, for: .horizontal)
+    }
+    
+    let folderName = UILabel().then {
+        $0.textAlignment = .left
+    }
+    
+    var spacerLine: UIView {
```
jhreis: Use `then`?

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/IndentedImageTableViewCell.swift
```diff
@@ -0,0 +1,87 @@
...
+
+    override init(style: UITableViewCellStyle, reuseIdentifier: String?) {
+        super.init(style: style, reuseIdentifier: reuseIdentifier)
+
+        indentationWidth = 20
+        mainStackView.addArrangedSubview(customImage)
+        
+        let transparentLine = spacerLine
+        transparentLine.backgroundColor = .clear
+        folderNameStackView.addArrangedSubview(transparentLine)
```
jhreis: I think following your past convention here may be a bit nicer:

```
[transparentLine, folderName, spacerLine].forEach(addArrangedSubview)
```

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/MenuViewController.swift
```diff
@@ -0,0 +1,199 @@
...
+            switch self {
+            case .bookmarks: return #imageLiteral(resourceName: "menu_bookmarks")
+            case .history: return #imageLiteral(resourceName: "menu-history")
+            case .settings: return #imageLiteral(resourceName: "menu-settings")
+            case .add: return #imageLiteral(resourceName: "menu-add-bookmark")
+            case .share: return #imageLiteral(resourceName: "nav-share")
+            }
+        }
+        
+        // TODO: Remove when we can use Swift 4.2/`CaseIterable`
```
jhreis: We should be on Swift 4.2 now, so maybe try to implement this?

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/MenuViewController.swift
```diff
@@ -0,0 +1,199 @@
...
+    }
+    
+    // MARK: - Table view data source
+    
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        
+        guard let cell = tableView.cellForRow(at: indexPath) else { return }
+        
+        switch cell.tag {
+        case MenuButtons.bookmarks.rawValue: openBookmarks()
```
jhreis: Rather than using `rawValue` for all of these, wdyt about:

```
guard let button = MenuButtons(rawValue: cell.tag) { 
    assertionFailure("No cell with \(cell.tag) tag.")
}

switch button { ... }
```

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/MenuViewController.swift
```diff
@@ -0,0 +1,199 @@
...
+        open(vc, doneButton: DoneButton(style: .done, position: .right))
+    }
+    
+    private func openSettings() {
+        let vc = SettingsViewController(profile: bvc.profile, tabManager: bvc.tabManager)
+        open(vc, doneButton: DoneButton(style: .done, position: .right))
+    }
+    
+    private func openShareSheet() {
+        dismiss(animated: true)
```
jhreis: Maybe just call `dismissView()`?

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/MenuViewController.swift
```diff
@@ -0,0 +1,199 @@
...
+        nav.modalPresentationStyle = .formSheet
+        
+        let button = UIBarButtonItem(barButtonSystemItem: doneButton.style, target: nav, action: #selector(SettingsNavigationController.done))
+        
+        switch doneButton.position {
+        case .left: nav.navigationBar.topItem?.leftBarButtonItem = button
+        case .right: nav.navigationBar.topItem?.rightBarButtonItem = button
+        }
+        
+        dismiss(animated: true)
```
jhreis: Maybe just call `dismissView()`?

Client/Frontend/Browser/Toolbars/UrlBar/URLBarView.swift
```diff
@@ -368,7 +279,9 @@ class URLBarView: UIView {
         locationTextField.attributedPlaceholder = self.locationView.placeholder
         locationContainer.addSubview(locationTextField)
         locationTextField.snp.remakeConstraints { make in
-            make.edges.equalTo(self.locationView)
+            let insets = UIEdgeInsets(top: 0, left: URLBarViewUX.LocationLeftPadding,
+                                      bottom: 0, right: URLBarViewUX.LocationLeftPadding)
```
jhreis: I believe this is intentional, but setting `right` to `LeftPadding` looks kind of goofy. Maybe rename, or add a `RightPadding` that is just literally equal to `LeftPadding`.

Client/Frontend/Browser/Toolbars/UrlBar/URLBarView.swift
```diff
@@ -479,71 +392,36 @@ class URLBarView: UIView {
...
         
-        UIView.animate(withDuration: 0.3, delay: 0.0, usingSpringWithDamping: 0.85, initialSpringVelocity: 0.0, options: [], animations: {
-            self.transitionToOverlay(cancel)
-            self.setNeedsUpdateConstraints()
-            self.layoutIfNeeded()
-        }, completion: { _ in
-            self.updateViewsForOverlayModeAndToolbarChanges()
-        })
+        if inOverlayMode {
+            [progressBar, navigationStackView, menuButton, shieldsButton, tabsButton, locationView.contentView].forEach {
```
jhreis, line 413: This is pretty intense, is it possible to add these into a parent type view, and just set that hidden prop?

Data/models/Bookmark.swift
```diff
@@ -105,6 +105,31 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
...
+        var predicate: NSPredicate?
+        if let excludedFolder = excludedFolder {
+            predicate = NSPredicate(format: "isFolder == true AND SELF != %@", excludedFolder)
+        } else {
+            predicate = NSPredicate(format: "isFolder == true")
+        }
+        
+        fetchRequest.predicate = predicate
+        
+        //"parentFolder == %@"
```
jhreis: Remove unneeded comment.

Data/models/Bookmark.swift
```diff
@@ -105,6 +105,31 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
...
+        let context = DataController.viewContext
+        let fetchRequest = NSFetchRequest<Bookmark>()
+        
+        fetchRequest.entity = entity(context: context)
+        fetchRequest.fetchBatchSize = 20
+        
+        let createdSort = NSSortDescriptor(key: "created", ascending: false)
+        fetchRequest.sortDescriptors = [createdSort]
+        
+        var predicate: NSPredicate?
```
jhreis, line 118: Maybe I'm being too DRY here, but... thoughts?

```
        var predicateFormat = "isFolder == true"
        if let excludedFolder = excludedFolder {
            predicateFormat = predicateFormat.appendingFormat(" AND SELF != %@", excludedFolder)
        }
        
        fetchRequest.predicate = NSPredicate(format: predicateFormat)
```

Data/models/Bookmark.swift
```diff
@@ -127,6 +152,17 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
         updateInternal(customTitle: customTitle, url: url)
     }
     
+    enum SaveLocation {
+        case keep
+        case new(location: Bookmark?)
+    }
+    
+    public func updateWithNewLocation(customTitle: String?, url: String?, location: Bookmark?) {
+        let isTitleEmpty = customTitle?.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty == true
```
jhreis: If we flip this conditional, it will auto check for `nil` and `empty`:

```
let containsText = customTitle?.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty == false
if !containsText { return }
```

Data/models/Bookmark.swift
```diff
@@ -127,6 +152,17 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
         updateInternal(customTitle: customTitle, url: url)
     }
     
+    enum SaveLocation {
+        case keep
+        case new(location: Bookmark?)
+    }
+    
+    public func updateWithNewLocation(customTitle: String?, url: String?, location: Bookmark?) {
```
jhreis, line 156: This is almost identical to `update(` [above](https://github.com/brave/brave-ios/pull/1058/files?file-filters%5B%5D=.png&file-filters%5B%5D=.swift&hide-deleted-files=true#diff-01f57e7d7609dba0eec131782b05f57dR148). Should abstract implementation. This could honestly even be a string helper.
Something super generic would be:
```containsCharacters(in: CharacterSet) -> Bool```

and could call:
```myText.containsCharacters(in: .whitespacesAndNewlines.invert())```

Otherwise could do something more like:
`hasVisibleCharacters`

Lastly, since this logic is only really used in this class, could just create a private function 😆

File: Data/models/Bookmark.swift [hunk 33]
```diff
@@ -306,13 +342,14 @@ extension Bookmark {
...
                                 save: Bool = true, sendToSync: Bool = true,
+                                location: SaveLocation = .keep,
                                 context: WriteContext = .new) {
         
         DataController.perform(context: context) { context in
             guard let bookmarkToUpdate = context.object(with: self.objectID) as? Bookmark else { return }
             
             // See if there has been any change
-            if bookmarkToUpdate.customTitle == customTitle && bookmarkToUpdate.url == url && bookmarkToUpdate.syncOrder == newSyncOrder {
+            if bookmarkToUpdate.customTitle == customTitle && bookmarkToUpdate.url == url && bookmarkToUpdate.syncOrder == newSyncOrder, case .keep = location {
```
jhreis, line 353: I have never seen this:
```case .keep = location```
Assuming this checks to make sure `location` equals `.keep`?
(like an inline switch conditional or something?)

File: Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/AddEditBookmarkTableViewController.swift [hunk 34]
```diff
@@ -0,0 +1,433 @@
...
+            switch self {
+            case .currentSelection: self = .folderHierarchy
+            case .folderHierarchy: self = .currentSelection
+            }
+        }
+    }
+    
+    // MARK: - View setup
+    
+    private lazy var saveButton: UIBarButtonItem = {
```
jhreis, line 37: `then`?

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/AddEditBookmarkTableViewController.swift
```diff
@@ -0,0 +1,433 @@
...
+            return BookmarkDetailsView(title: bookmark.displayTitle, url: bookmark.url)
+        case .editFolder(let folder):
+            return FolderDetailsViewTableViewCell(title: folder.displayTitle, viewHeight: UX.cellHeight)
+        }
+    }()
+    
+    // MARK: - Special cells
+    
+    /// A non-regular folder cell. Raw value represents a tag value of the cell
+    enum SpecialCell: Int {
```
jhreis, line 63: `private`?

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/AddEditBookmarkTableViewController.swift
```diff
@@ -0,0 +1,433 @@
...
+        case favorites = 10
+        case rootLevel = 11
+        case addFolder = 12
+    }
+    
+    let folderCellTag = 13
```
jhreis: `private?`

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/AddEditBookmarkTableViewController.swift
```diff
@@ -0,0 +1,433 @@
...
+            $0.tag = SpecialCell.favorites.rawValue
+            if case .favorites = saveLocation, presentationMode == .folderHierarchy {
+                $0.accessoryType = .checkmark
+            }
+        }
+        
+        return cell
+    }
+    
+    private var addNewFolderCell: IndentedImageTableViewCell {
```
jhreis, line 103: `then`

(also, same problem, where this returns new instance each time)

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/AddEditBookmarkTableViewController.swift
```diff
@@ -0,0 +1,433 @@
...
+            $0.tag = SpecialCell.rootLevel.rawValue
+            if case .rootLevel = saveLocation, presentationMode == .folderHierarchy {
+                $0.accessoryType = .checkmark
+            }
+        }
+        
+        return cell
+    }
+    
+    private var favoritesCell: IndentedImageTableViewCell {
```
jhreis, line 91: This is really problematic, as it returns a new cell instance each time. Don't think you can use `then` or `let` constant here since it uses instance variables inside of it, but at least this should be `lazy` to avoid returning new instances.

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/AddEditBookmarkTableViewController.swift
```diff
@@ -0,0 +1,433 @@
...
+    
+    /// Returns a count of how many non-folder cells should be visible(depends on Mode state)
+    private var specialButtonsCount: Int {
+        switch mode {
+        case .addFolder(_), .editFolder(_): return 1
+        case .addBookmark(_, _), .editBookmark(_): return 3
+        }
+    }
+    
+    private var rootLevelFolderCell: IndentedImageTableViewCell {
```
jhreis, line 79: Ditto:

> This is really problematic, as it returns a new cell instance each time. Don't think you can use then or let constant here since it uses instance variables inside of it, but at least this should be lazy to avoid returning new instances.

jhreis (changes requested):
A lot of comments, but they are all pretty minor honestly. This PR is great @iccub, absolutely wonderful. Really looking forward to this, and keep it up!! 😁

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/AddEditHeaderView.swift
iccub: Kind of, we set frame in `AddEditBookmarkTableViewController` like this 
`header.frame = CGRect(x: 0, y: 0, width: tableView.bounds.width, height: tableView.bounds.height)`

but then we use `systemLayoutSizeFitting(UILayoutFittingCompressedSize)` to resize its height

So it uses full tableview width, plus system fitting height

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/BookmarkDetailsView.swift
iccub: This is optional by a `BookmarkFormFieldsProtocol` so I could abstract `validateFields()` for both Bookmark(url and title) and folder(title only)

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/BookmarkDetailsView.swift
iccub: Will change it, this text field setup was taken from previous implementation

Data/models/Bookmark.swift
iccub: Tested it and it crashes, apparently NSPredicate's does something more under the hood, probably converts the `NSObject` for `SELF` comparison correctly

Data/models/Bookmark.swift
iccub: Yes, this is a really weird way to check equality in enums with associated values.

so I can't do simply do `location == .keep` because one of the enum values has associated value(`case new(location: Bookmark?)`)

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/AddEditBookmarkTableViewController.swift
iccub: it's used in `BookmarkEditMode`

Client/Frontend/Browser/Toolbars/BottomToolbar/Menu/Bookmarks/AddEditBookmarkTableViewController.swift
iccub: They need to be recreated when switching between presentation modes, otherwise it causes problems, the cell disappears when toggling between folder hierarchy/current selection

Data/models/Bookmark.swift
iccub: Made a private function

iccub:
FYI updating folder location doesn't work with sync, investigating it.

Edit: should be fixed by https://github.com/brave/brave-ios/pull/1058/commits/76ab847d667abca3856a6e773a87c11a0e768750

jhreis (approved):

PR #1128, opened 2019-06-06
Fix #1127: Select tab if it's not selected to prevent fatalCrash

### LINKED ISSUE #1127: App is crashed with `fatalError` on push animation of BVC to Tray when selectedTab is nil.
For an unknown reason the selectedTab is nil when transitioning from `BrowserViewController` to `TabTray` using a custom push animation. It 's not yet known how and why the selectedTab is nil at that point of time, since the moment the you close the current selectedTab a new one is selected or if empty a new one is created and selected. 

### Steps to Reproduce 
No STR. Please update if found.
  1.
  2.
  3.

**Actual result:** 

**Expected result:**

**Brave Version:** 

**Device details:** 

**Website problems only:**

danishjafri88:
This fix should stop the crashes, but the cause of `nil` selectedTab is still unknown.

## Submitter Checklist:

## Reviewer Checklist:

  - [x] `QA/(Yes|No)`
  - [x] `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1426,10 +1426,16 @@ extension BrowserViewController {
 
 extension BrowserViewController: URLBarDelegate {
     func showTabTray() {
+        if tabManager.tabsForCurrentMode.isEmpty {
+            return
+        }
         updateFindInPageVisibility(visible: false)
         
         let tabTrayController = TabTrayController(tabManager: tabManager, profile: profile, tabTrayDelegate: self)
         
+        if tabManager.selectedTab == nil {
+            tabManager.selectTab(tabManager.tabsForCurrentMode.first)
+        }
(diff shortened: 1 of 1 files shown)


iccub (approved):
looks promising :)

as you said in the description, it's not easy to test so we just have to monitor crashes in Xcode organizer

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1426,10 +1426,16 @@ extension BrowserViewController {
 
 extension BrowserViewController: URLBarDelegate {
     func showTabTray() {
+        guard !tabManager.tabsForCurrentMode.isEmpty else {
```
jhreis: This is a double negative. Just use an `if`:

```suggestion
        if tabManager.tabsForCurrentMode.isEmpty {
```

jhreis (changes requested):

jhreis (approved):

PR #1341, opened 2019-08-07
Fix #1298: Text input support on AlertPopupView

### LINKED ISSUE #1298: Extend AlertPopupView to include an option to enter user text
### Description: 
`UserTextInputAlert` dialogs are spoof-able, the dialogs look very similar to `prompt()` alerts in javascript. We should update `AlertPopupView` to include an option for user to enter text to differentiate browser alerts from alerts from web pages.

jamesmudgett:
Adds text input support for AlertPopupView. Supports all input(keyboard) types and secure input as well. Example usage:
```
let popup = AlertPopupView(image: #imageLiteral(resourceName: "browser_lock_popup"), title: "PIN verification required", message: "", inputType: .asciiCapable, secureInput: true, inputPlaceholder: "Enter your Yubikey PIN")

        popup.addButton(title: "Cancel") { () -> PopupViewDismissType in
            return .flyDown
        }

        popup.addButton(title: "Verify", type: .primary) { [weak self] () -> PopupViewDismissType in
            guard let pin = popup.text else { return .deny }

            // ...more error handling.
            if pin.isEmpty {
                return .deny
            }

            print(pin)

            return .flyUp
        }
        browserLockPopup = popup
        popup.showWithType(showType: .flyUp)
```

## Submitter Checklist:

### Screenshots:

Simulator Screen Shot - iPhone Xs - 2019-08-07 at 15 56 41

## Reviewer Checklist:

  - [x] `QA/(Yes|No)`
  - [x] `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Popup/AlertPopupView.swift
+++ b/Client/Frontend/Popup/AlertPopupView.swift
@@ -6,16 +6,20 @@ import UIKit
 import BraveShared
 
 class AlertPopupView: PopupView {
-    
     fileprivate var dialogImage: UIImageView?
     fileprivate var titleLabel: UILabel!
     fileprivate var messageLabel: UILabel!
     fileprivate var containerView: UIView!
+    fileprivate var textField: UITextField?
+    
+    var text: String? {
+        return textField?.text
+    }
     
     fileprivate let kAlertPopupScreenFraction: CGFloat = 0.8
     fileprivate let kPadding: CGFloat = 20.0
     
-    init(image: UIImage?, title: String, message: String) {
+    init(image: UIImage?, title: String, message: String, inputType: UIKeyboardType? = nil, secureInput: Bool = false, inputPlaceholder: String? = nil) {
         super.init(frame: CGRect.zero)
         
         overlayDismisses = false
@@ -47,6 +51,23 @@ class AlertPopupView: PopupView {
         messageLabel.text = message
         messageLabel.numberOfLines = 0
         containerView.addSubview(messageLabel)
+        
+        if let inputType = inputType {
+            textField = UITextField(frame: CGRect.zero).then {
+                $0.keyboardType = inputType
+                $0.textColor = .black
+                $0.placeholder = inputPlaceholder ?? ""
+                $0.autocorrectionType = .no
+                $0.autocapitalizationType = .none
+                $0.layer.cornerRadius = 4
+                $0.layer.borderColor = UIColor(white: 0, alpha: 0.3).cgColor
+                $0.layer.borderWidth = 1
+                $0.delegate = self
+                $0.textAlignment = .center
+                $0.isSecureTextEntry = secureInput
+                containerView.addSubview($0)
+            }
+        }
         
         updateSubviews()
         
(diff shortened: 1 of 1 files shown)


Client/Frontend/Popup/AlertPopupView.swift
```diff
@@ -48,6 +52,22 @@ class AlertPopupView: PopupView {
         messageLabel.numberOfLines = 0
         containerView.addSubview(messageLabel)
         
+        if let inputType = inputType {
```
iccub, line 55: add `let textField = textField` here to avoid unwrappings

Client/Frontend/Popup/AlertPopupView.swift
```diff
@@ -48,6 +52,22 @@ class AlertPopupView: PopupView {
...
+            textField = UITextField(frame: CGRect.zero)
```
iccub: to make it a bit simpler replace all these textFields with `Then` library
```
textField.do {
  $0.keyBoardType = input type
  $0.textColor = .black
   ...
}

File: Client/Frontend/Popup/AlertPopupView.swift [hunk 3]
```diff
@@ -6,16 +6,20 @@ import UIKit
...
 
 class AlertPopupView: PopupView {
-    
     fileprivate var dialogImage: UIImageView?
     fileprivate var titleLabel: UILabel!
     fileprivate var messageLabel: UILabel!
     fileprivate var containerView: UIView!
+    fileprivate var textField: UITextField?
+    
+    public var text: String? {
```
iccub: Remove `public`, it's needed only in frameworks

Client/Frontend/Popup/AlertPopupView.swift
```diff
@@ -100,9 +120,20 @@ class AlertPopupView: PopupView {
         messageLabelFrame.origin.y = rint(titleLabelFrame.maxY + kPadding * 1.5 / 2.0 * resizePercentage)
         messageLabel.frame = messageLabelFrame
         
+        var textFieldFrame = textField?.frame ?? CGRect.zero
+        var maxY = !(messageLabel.text?.isEmpty ?? true) ? messageLabelFrame.maxY : titleLabelFrame.maxY
```
iccub: replace `!(messageLabel.text?.isEmpty ?? true)` with `messageLabel.text?.isEmpty == false`, it's a bit simpler to read

package-lock.json
```diff
@@ -104,10 +104,22 @@
       "version": "github:brave/ad-block#b933cf5168c5284dc159735d8bcefb415d0cfc00",
       "from": "github:brave/ad-block#b933cf5168c5284dc159735d8bcefb415d0cfc00",
       "requires": {
-        "bloom-filter-cpp": "^1.2.2",
+        "bloom-filter-cpp": "^1.2.0",
```
iccub: It would be best to not version this `package-lock.json` file at all for this PR

Client/U2FExtensions.swift
```diff
@@ -156,6 +156,29 @@ class U2FExtensions: NSObject {
         
         // Make sure the session is started
         YubiKitManager.shared.keySession.startSession()
+        
+//        // example popup with input
```
iccub: this still needed?

iccub (changes requested):
Few code style comments

Client/Frontend/Popup/AlertPopupView.swift
```diff
@@ -100,9 +120,20 @@ class AlertPopupView: PopupView {
...
+        if textField != nil {
```
jhreis: unwrap here `let textField = textField`.

Client/Frontend/Popup/AlertPopupView.swift
jhreis: I think even easier would be to reverse this:
```suggestion
        var maxY = messageLabel.text?.isEmpty == true ? titleLabelFrame.maxY : messageLabelFrame.maxY
```

Client/U2FExtensions.swift
jhreis: ... if not, then remove.

jhreis (changes requested):

Client/Frontend/Popup/AlertPopupView.swift
jamesmudgett: Not needed here with change to use Then framework.

jhreis (approved):

jhreis:
lgtm. @jamesmudgett in future please make sure to complete the PR template submitter checklist. 

@iccub feel free to merge if you approve it, please just verify the squashed message, as it may not be correct by default.

iccub (approved):

PR #1366, opened 2019-08-09
Fixed for iPad Fullscreen using Desktop UserAgent 

Brandon-T:
- Fixed for iPad Fullscreen using Desktop UserAgent and overall Fullscreen support for iOS-12 and iOS-13.
- Tested websites: Vimeo, DailyMotion, Youtube, CouchTuner, JWPlayer.
- Injects Javascript into the webpage to let it know that "Videos" support Fullscreen. This does NOT fix the fact that WebKit on iOS does NOT support HTML5 Fullscreen API! This means DIV tags and other tags CANNOT request fullscreen as it will never work anyway! This only fixes it for sites like Youtube where they attempt to full-screen the div-tag instead of the embedded video-tag. 

What it does: It hooks the full-screen API and returns true only if native-fullscreen is NOT supported and video tags support native-fs. When FS is requested on tags that are NOT video tags, it finds the first child video tag that is active and requests full-screen on that. This works for 100% of the video sites that I've tested.

If you have a different video site to test and this does NOT work for that site, please comment below or tag me or something. So far it's been a very solid work-around.

## Submitter Checklist:

## Reviewer Checklist:

  - [x] `QA/(Yes|No)`
  - [x] `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/UserContent/UserScripts/FullscreenHelper.js
+++ b/Client/Frontend/UserContent/UserScripts/FullscreenHelper.js
@@ -1,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+
+/*var $<possiblySupportedFunctions> = {
+    request: ["requestFullscreen", "webkitRequestFullscreen", "webkitRequestFullScreen", "mozRequestFullScreen", "msRequestFullscreen"],
+    exit: ["exitFullscreen", "webkitCancelFullScreen", "webkitExitFullscreen", "mozCancelFullScreen", "msExitFullscreen"],
+    enabled: ["fullscreenEnabled", "webkitFullscreenEnabled", "mozFullScreenEnabled", "msFullscreenEnabled"],
+    element: ["fullscreenElement", "webkitFullscreenElement", "webkitCurrentFullScreenElement", "mozFullScreenElement", "msFullscreenElement"],
+    change: ["fullscreenchange", "webkitfullscreenchange", "mozfullscreenchange", "MSFullscreenChange"],
+    error: ["fullscreenerror", "webkitfullscreenerror", "mozfullscreenerror", "MSFullscreenError"]
+}*/
+
+var $<isFullscreenSupportedNatively> = document.fullscreenEnabled ||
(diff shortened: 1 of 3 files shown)


Client/Frontend/UserContent/UserScripts/FullscreenHelper.js
```diff
@@ -0,0 +1,63 @@
...
+/*var $<possiblySupportedFunctions> = {
+    request: ["requestFullscreen", "webkitRequestFullscreen", "webkitRequestFullScreen", "mozRequestFullScreen", "msRequestFullscreen"],
+    exit: ["exitFullscreen", "webkitCancelFullScreen", "webkitExitFullscreen", "mozCancelFullScreen", "msExitFullscreen"],
+    enabled: ["fullscreenEnabled", "webkitFullscreenEnabled", "mozFullScreenEnabled", "msFullscreenEnabled"],
+    element: ["fullscreenElement", "webkitFullscreenElement", "webkitCurrentFullScreenElement", "mozFullScreenElement", "msFullscreenElement"],
+    change: ["fullscreenchange", "webkitfullscreenchange", "mozfullscreenchange", "MSFullscreenChange"],
+    error: ["fullscreenerror", "webkitfullscreenerror", "mozfullscreenerror", "MSFullscreenError"]
+}*/
+
+var $<isFullscreenSupportedNatively> = document.fullscreenEnabled ||
```
jhreis, line 15: `let` or `const` if unedited probably.

jhreis (changes requested):
Few questions about this:
1. it does not impact mobile UA users, correct?
2. this is passive, meaning it can always be activated, and we don't need to deal with pulling or pushing these scripts into web contexts? (i.e. we just always use them).

Client/Frontend/UserContent/UserScripts/FullscreenHelper.js
Brandon-T: For the first comment, yes it's passive and we don't need to keep pulling them in.
For the second comment, I think it would affect BOTH but atm, I don't see a difference. IE: I don't see it causing any problems as the behaviour is universal.

I tested in both Mobile UA and Desktop UA. Alternatively, I can just make it "active" ONLY if the user agent is Desktop. I think that might be best anyway so I'll update the PR to do that (IE: still passive but only active when using Desktop UA).

jhreis (approved):

jhreis:
Unblocking, as nothing to prevent it from being merged.

PR #1368, opened 2019-08-09
Fixes: #1290. iPad will now have Desktop UA as default in iOS 13 with a setting to disable it

### LINKED ISSUE #1290: User agent on iPad should default to a desktop one.
### Steps to Reproduce 
  1. Visit a website

**Actual result:** 

**Expected result:**

**Brave Version:** 

**Device details:** 

**Website problems only:**

danishjafri88:
Fixes: #1290, iPad will use Desktop UA by default.

To test please visit https://www.whatsmyua.info
and load the page in iPad/iPhone (iOS 12 & 13) and also try changing the mode from share option

On iOS 12 (iPhone & Pad)& iOS 13 (iPhone), there should be no change and should be same as older versions,
while on iOS 13 iPad the launch should be. in desktop and the setting will be available 
## Submitter Checklist:

## Reviewer Checklist:

  - [ ] `QA/(Yes|No)`
  - [ ] `bug` / `enhancement`

### DIFF
--- a/BraveShared/BraveStrings.swift
+++ b/BraveShared/BraveStrings.swift
@@ -393,6 +393,7 @@ extension Strings {
     public static let Privacy = NSLocalizedString("Privacy", tableName: "BraveShared", bundle: Bundle.braveShared, value: "Privacy", comment: "Settings privacy section title")
     public static let Security = NSLocalizedString("Security", tableName: "BraveShared", bundle: Bundle.braveShared, value: "Security", comment: "Settings security section title")
     public static let Save_Logins = NSLocalizedString("SaveLogins", tableName: "BraveShared", bundle: Bundle.braveShared, value: "Save Logins", comment: "Setting to enable the built-in password manager")
+    public static let AlwaysRequestDesktopSite = NSLocalizedString("AlwaysRequestDesktopSite", tableName: "BraveShared", bundle: Bundle.braveShared, value: "Always request desktop site", comment: "Setting to always request desktop site")
     public static let ShieldsAdStats = NSLocalizedString("AdsrBlocked", tableName: "BraveShared", bundle: Bundle.braveShared, value: "Ads \nBlocked", comment: "Shields Ads Stat")
     public static let ShieldsAdAndTrackerStats = NSLocalizedString("AdsAndTrackersrBlocked", tableName: "BraveShared", bundle: Bundle.braveShared, value: "Ads & Trackers \nBlocked", comment: "Shields Ads Stat")
     public static let ShieldsTrackerStats = NSLocalizedString("TrackersrBlocked", tableName: "BraveShared", bundle: Bundle.braveShared, value: "Trackers \nBlocked", comment: "Shields Trackers Stat")
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1277,8 +1277,9 @@ class BrowserViewController: UIViewController {
             }
         }
         
-        // Remember whether or not a desktop site was requested
-        tab.desktopSite = webView.customUserAgent?.isEmpty == false
+        // Remember whether or not a desktop/mobile site was requested
+        let isCustomUserAgentEmpty = webView.customUserAgent?.isEmpty == true
+        tab.desktopSite = isCustomUserAgentEmpty ? UIDevice.isIpad : !UIDevice.isIpad
     }
     
     // MARK: - Browser PIN Callout
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -202,6 +202,7 @@ class BrowserViewController: UIViewController {
         Preferences.Privacy.privateBrowsingOnly.observe(from: self)
         Preferences.General.tabBarVisibility.observe(from: self)
         Preferences.Shields.allShields.forEach { $0.observe(from: self) }
+        Preferences.General.alwaysRequestDesktopSite.observe(from: self)
         Preferences.Privacy.blockAllCookies.observe(from: self)
         // Lists need to be compiled before attempting tab restoration
         contentBlockListDeferred = ContentBlockerHelper.compileBundledLists()
@@ -1132,7 +1133,11 @@ class BrowserViewController: UIViewController {
     func restoreSpoofedUserAgentIfRequired(_ webView: WKWebView, newRequest: URLRequest) {
         // Restore any non-default UA from the request's header
         let ua = newRequest.value(forHTTPHeaderField: "User-Agent")
-        webView.customUserAgent = ua != UserAgent.defaultUserAgent() ? ua : nil
+        if #available(iOS 13.0, *) {
+            webView.customUserAgent = Preferences.General.alwaysRequestDesktopSite.value == UserAgent.isDesktopUA(uaString: ua ?? "") ? nil : ua
+        } else {
+            webView.customUserAgent = ua != UserAgent.defaultUserAgent() ? ua : nil
+        }
     }
 
     fileprivate func presentActivityViewController(_ url: URL, tab: Tab? = nil, sourceView: UIView?, sourceRect: CGRect, arrowDirection: UIPopoverArrowDirection) {
@@ -1278,8 +1283,15 @@ class BrowserViewController: UIViewController {
         }
         
         // Remember whether or not a desktop/mobile site was requested
-        let isCustomUserAgentEmpty = webView.customUserAgent?.isEmpty == true
-        tab.desktopSite = isCustomUserAgentEmpty ? UIDevice.isIpad : !UIDevice.isIpad
+        if #available(iOS 13.0, *) {
+            if let customUA = webView.customUserAgent, customUA.isEmpty == false {
+                tab.desktopSite = UserAgent.isDesktopUA(uaString: customUA)
+            } else {
+                tab.desktopSite = Preferences.General.alwaysRequestDesktopSite.value
+            }
+        } else {
+            tab.desktopSite = webView.customUserAgent?.isEmpty == false
+        }
     }
     
     // MARK: - Browser PIN Callout
@@ -3055,6 +3067,15 @@ extension BrowserViewController: PreferencesObserver {
         switch key {
         case Preferences.General.tabBarVisibility.key:
             updateTabsBarVisibility()
+        case Preferences.General.alwaysRequestDesktopSite.key:
+            tabManager.reset()
+            self.tabManager.reloadSelectedTab()
+            for tab in self.tabManager.allTabs where tab != self.tabManager.selectedTab {
+                tab.createWebview()
+                if let url = tab.webView?.url {
+                    tab.loadRequest(PrivilegedRequest(url: url) as URLRequest)
+                }
+            }
         case Preferences.Privacy.privateBrowsingOnly.key:
             let isPrivate = Preferences.Privacy.privateBrowsingOnly.value
             switchToPrivacyMode(isPrivate: isPrivate)
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1284,7 +1284,7 @@ class BrowserViewController: UIViewController {
         
         // Remember whether or not a desktop/mobile site was requested
         if #available(iOS 13.0, *) {
-            if let customUA = webView.customUserAgent, customUA.isEmpty == false {
+            if let customUA = webView.customUserAgent, !customUA.isEmpty {
                 tab.desktopSite = UserAgent.isDesktopUA(uaString: customUA)
             } else {
                 tab.desktopSite = Preferences.General.alwaysRequestDesktopSite.value
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
@@ -64,6 +64,16 @@ extension BrowserViewController: WKNavigationDelegate {
         return false
     }
 
+    // This is the iOS 13 delegate callback which is similar to the delegate below it
+    // except that a WKWebpagePreferences is provided, Setting the type of contentMode loads the site accordingly.
+    @available(iOS 13.0, *)
+    func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, preferences: WKWebpagePreferences, decisionHandler: @escaping (WKNavigationActionPolicy, WKWebpagePreferences) -> Void) {
+        self.webView(webView, decidePolicyFor: navigationAction) {
+            preferences.preferredContentMode = Preferences.General.alwaysRequestDesktopSite.value ? .desktop :. mobile
+            decisionHandler($0, preferences)
+        }
+    }
+    
     // This is the place where we decide what to do with a new navigation action. There are a number of special schemes
     // and http(s) urls that need to be handled in a different way. All the logic for that is inside this delegate
     // method.
--- a/Client/Frontend/Browser/Tab.swift
+++ b/Client/Frontend/Browser/Tab.swift
@@ -123,7 +123,7 @@ class Tab: NSObject {
 
     /// Whether or not the desktop site was requested with the last request, reload or navigation. Note that this property needs to
     /// be managed by the web view's navigation delegate.
-    var desktopSite: Bool = false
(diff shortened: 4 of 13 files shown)


kylehickinson (commented):
Do we want to use desktop UA even when the user is using Brave in split-screen/multitasking and the horizontal size class becomes `compact`?

Brandon-T:
Running your code on Youtube still has this one problem from WebKit (also exists in Firefox iOS):

Screen Shot 2019-08-09 at 8 50 33 PM

We actually need BOTH of these PRs for iPad. Danish's PR sets the User-Agent to default.. and mine fixes the Full-Screen issue: https://github.com/brave/brave-ios/pull/1366

Brandon-T:
For iOS 13:

```
if #available(iOS 13.0, *) {
    configuration.defaultWebpagePreferences.preferredContentMode = .recommended
}
```


We can use `.recommended`:

```swift
/*
WKContentModeRecommended -  The recommended content mode for the current platform
WKContentModeMobile  -     Represents content targeting mobile browsers
WKContentModeDesktop  -     Represents content targeting desktop browsers
WKContentModeRecommended - behaves like WKContentModeMobile on iPhone and iPad mini
     and WKContentModeDesktop on other iPad models as well as Mac.
*/
```

Still same FS problem though but that's about it. It'll take care of Desktop UA itself.

Client/Frontend/Browser/Tab.swift
```diff
@@ -123,7 +123,7 @@ class Tab: NSObject {
 
     /// Whether or not the desktop site was requested with the last request, reload or navigation. Note that this property needs to
     /// be managed by the web view's navigation delegate.
-    var desktopSite: Bool = false
+    var desktopSite: Bool = UIDevice.isIpad ? true : false
```
iccub: This should be a `Preference` that can be set in settings. Turned on by default on iPads.

I've started some initial work on it here if you want to take a look https://github.com/brave/brave-ios/compare/development...iccub:bugfix/1290?expand=1

iccub (changes requested):

Client/Frontend/Browser/Tab.swift
```diff
@@ -123,7 +123,7 @@ class Tab: NSObject {
 
     /// Whether or not the desktop site was requested with the last request, reload or navigation. Note that this property needs to
     /// be managed by the web view's navigation delegate.
-    var desktopSite: Bool = false
+    var desktopSite: Bool = Preferences.General.alwaysRequestDesktopSite.value ? true : false
```
jhreis: `Preferences.General.alwaysRequestDesktopSite.value` is a `Bool`, no need for the ternary operator here.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1132,7 +1133,11 @@ class BrowserViewController: UIViewController {
     func restoreSpoofedUserAgentIfRequired(_ webView: WKWebView, newRequest: URLRequest) {
         // Restore any non-default UA from the request's header
         let ua = newRequest.value(forHTTPHeaderField: "User-Agent")
-        webView.customUserAgent = ua != UserAgent.defaultUserAgent() ? ua : nil
+        if #available(iOS 13.0, *) {
```
Brandon-T: We use if-available here but above we is `islessthan13`

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1277,8 +1282,16 @@ class BrowserViewController: UIViewController {
             }
         }
         
-        // Remember whether or not a desktop site was requested
-        tab.desktopSite = webView.customUserAgent?.isEmpty == false
+        // Remember whether or not a desktop/mobile site was requested
+        if #available(iOS 13.0, *) {
+            if let customUA = webView.customUserAgent, customUA.isEmpty == false {
```
Brandon-T: `customUA.isEmpty == false` can be changed to `!customUA.isEmpty`

Client/Frontend/Browser/Tab.swift
Brandon-T: No need for branch statement `? true : false` because the `alwaysRequestDesktopSite.value` already results in a boolean.

I think we can just do:

`var desktopSite = Preferences.General.alwaysRequestDesktopSite.value`

Client/Frontend/Browser/TabManager.swift
```diff
@@ -165,6 +165,11 @@ class TabManager: NSObject {
     private class func getNewConfiguration() -> WKWebViewConfiguration {
         let configuration = WKWebViewConfiguration()
         configuration.processPool = WKProcessPool()
+        // Setting the Unique applicationName for the user agent to identify Brave.
+        if #available(iOS 13.0, *) {
+            configuration.applicationNameForUserAgent = Preferences.General.alwaysRequestDesktopSite.value ?
+            UserAgent.desktopUAApplicationName() : UserAgent.mobileUAApplicationName()
```
Brandon-T, line 171: In both cases won't this value be: `FxiOS`?

Shared/SwiftAttributes.swift
```diff
@@ -0,0 +1,12 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+
+public var lessThaniOS13: Bool {
```
Brandon-T, line 7: Used only in one place, can we make it a lambda/location function for that one scenario instead of a global variable?

Brandon-T (changes requested):

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -64,6 +64,16 @@ extension BrowserViewController: WKNavigationDelegate {
         return false
     }
 
+    // This is the iOS 13 delegate callback which is similar to the delegate below it
+    // except that a WKWebpagePreferences is provided, Setting the type of contentMode loads the site accordingly.
+    @available(iOS 13.0, *)
+    func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, preferences: WKWebpagePreferences, decisionHandler: @escaping (WKNavigationActionPolicy, WKWebpagePreferences) -> Void) {
+        self.webView(webView, decidePolicyFor: navigationAction) {
+            preferences.preferredContentMode = Preferences.General.alwaysRequestDesktopSite.value ? .desktop :. mobile
+            decisionHandler($0, preferences)
```
Brandon-T, line 73: This breaks `requestMobileSite` from the Share Menu because it only obeys your "settings" from `Preferences` and not your site's preference..

IE: Have AlwaysRequestDesktop site on.. Then go to youtube and go to share menu and choose `RequestMobileSite`, it will still show Desktop, but if you go back to that same share menu, it will show `RequestDesktopSite` even though you're already on desktop UA..

In other words, the text will change for the share menu button but it does nothing in the end.

Brandon-T (changes requested):

Client/Frontend/Browser/TabManager.swift
danishjafri88: No, in mobile it will be FxiOS and for desktop it will be Safari

Client/Frontend/Browser/Tab.swift
danishjafri88: Lol, bad copy

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
danishjafri88: Its not because of the settings but somehow YouTube is setting the url param as app=desktop. If you instead try  https://www.whatsmyua.info or anyother site it shows it works fine.

danishjafri88: let me see if I can fix it on our end

danishjafri88: I found that the reason is because we are trying to open a mobile site ie `m.youtube.com` in Desktop. Youtube then redirects it to `https://www.youtube.com/?app=desktop`. When we change the UA with the param app=desktop no change is done. 

I propose that we also change our favourites URL from mobile to desktop in this case? @Brandon-T  @iccub @jhreis

danishjafri88: It happens on Desktop as well

Brandon-T (approved):

iccub (changes requested):
Please rebase to build with iOS 13

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
iccub: It doesn't work for me.
I open https://www.whatsmyua.info, request mobile site, and doesn't switch to mobile

I agree we can remove `mobile` part from favorites, youtube and twitter

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -64,6 +64,16 @@ extension BrowserViewController: WKNavigationDelegate {
         return false
     }
 
+    // This is the iOS 13 delegate callback which is similar to the delegate below it
+    // except that a WKWebpagePreferences is provided, Setting the type of contentMode loads the site accordingly.
+    @available(iOS 13.0, *)
+    func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, preferences: WKWebpagePreferences, decisionHandler: @escaping (WKNavigationActionPolicy, WKWebpagePreferences) -> Void) {
+        self.webView(webView, decidePolicyFor: navigationAction) {
+            preferences.preferredContentMode = Preferences.General.alwaysRequestDesktopSite.value ? .desktop :. mobile
```
iccub, line 72: `.desktop :. mobile` add space after `:`

ClientTests/ClientTests.swift
```diff
@@ -14,13 +14,12 @@ import Alamofire
 class ClientTests: XCTestCase {
 
     // Simple test to make sure the WKWebView UA matches the expected FxiOS pattern.
-    func testUserAgent() {
+    func testMobileUserAgent() {
```
iccub, line 17: This test doesn't work for me, breaks on CI too

Shared/UserAgent.swift
```diff
@@ -45,59 +45,51 @@ open class UserAgent {
                 return firefoxUA
             }
         }
-
         return nil
     }
 
+    private static var systemDefaultUA: String = ""
```
iccub, line 51: `String` can be removed

iccub (changes requested):
One unit test is failing

`Request mobile` site doesn't seem to be working for me(XCode beta 6)

BraveShared/BraveStrings.swift
```diff
@@ -393,6 +393,7 @@ extension Strings {
     public static let Privacy = NSLocalizedString("Privacy", tableName: "BraveShared", bundle: Bundle.braveShared, value: "Privacy", comment: "Settings privacy section title")
     public static let Security = NSLocalizedString("Security", tableName: "BraveShared", bundle: Bundle.braveShared, value: "Security", comment: "Settings security section title")
     public static let Save_Logins = NSLocalizedString("SaveLogins", tableName: "BraveShared", bundle: Bundle.braveShared, value: "Save Logins", comment: "Setting to enable the built-in password manager")
+    public static let AlwaysRequestDesktopSite = NSLocalizedString("AlwaysRequestDesktopSite", tableName: "BraveShared", bundle: Bundle.braveShared, value: "Always request desktop site", comment: "Setting to always request desktop site")
```
jhreis, line 396: > Setting to always use the desktop version of a site

```suggestion
    public static let AlwaysRequestDesktopSite = NSLocalizedString("AlwaysRequestDesktopSite", tableName: "BraveShared", bundle: Bundle.braveShared, value: "Always request desktop site", comment: "Setting to always use the desktop version of a site")
```

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -3054,6 +3067,15 @@ extension BrowserViewController: PreferencesObserver {
         switch key {
         case Preferences.General.tabBarVisibility.key:
             updateTabsBarVisibility()
+        case Preferences.General.alwaysRequestDesktopSite.key:
+            tabManager.reset()
+            self.tabManager.reloadSelectedTab()
+            for tab in self.tabManager.allTabs where tab != self.tabManager.selectedTab {
```
jhreis, line 3073: I think this is overkill. I think `tabManager.reset()` should be enough. We definitely _do not) want to reload _all_ tabs, this could create hundreds of webviews.

Client/Frontend/Browser/Tab.swift
```diff
@@ -195,6 +195,11 @@ class Tab: NSObject {
             configuration!.preferences = WKPreferences()
             configuration!.preferences.javaScriptCanOpenWindowsAutomatically = false
             configuration!.allowsInlineMediaPlayback = true
+            if #available(iOS 13.0, *) {
+                configuration!.defaultWebpagePreferences = WKWebpagePreferences().then {
+                    $0.preferredContentMode = Preferences.General.alwaysRequestDesktopSite.value ? .desktop : .mobile
```
jhreis, line 200: Not a big fan of the duplication:

https://github.com/brave/brave-ios/pull/1368/files#diff-c0ee444fbdc7d184b084fb4cc3e534fcR72

But may not easy way around this?

Client/Frontend/Browser/Tab.swift
```diff
@@ -399,7 +404,16 @@ class Tab: NSObject {
     }
 
     func reload() {
-        let userAgent: String? = desktopSite ? UserAgent.desktopUserAgent() : nil
+        
+        var userAgent: String?
```
jhreis, line 408: Rather than specifying the `userAgent` here, we should be specifying the `mobileUA`:

```
var mobileUA: String? = nil

if #available(iOS 13.0, *) {
    mobileUA = UserAgent.mobileUserAgent()
}

let userAgent: String? = desktopSite ? UserAgent.desktopUserAgent() : mobileUA
```

This is _much_ cleaner IMO.

@danishjafri88 when seeing code duplication like this (`userAgent = desktopSite ? UserAgent.desktopUserAgent() : `), it is best to try to identify how we can abstract the duplicated portions.

Client/Frontend/Settings/SettingsViewController.swift
```diff
@@ -166,7 +166,10 @@ class SettingsViewController: TableViewController {
             }
             general.rows.append(row)
         }
-        
+        if #available(iOS 13.0, *), UIDevice.isIpad {
```
jhreis: Let's remove this iPad check.

Shared/SwiftAttributes.swift
jhreis, line 7: Let's just fully delete this.

Shared/UserAgent.swift
```diff
@@ -45,59 +45,51 @@ open class UserAgent {
...
     public static func defaultUserAgent() -> String {
         assert(Thread.current.isMainThread, "This method must be called on the main thread.")
-
         if let firefoxUA = UserAgent.cachedUserAgent(checkiOSVersion: true) {
             return firefoxUA
         }
-
-        let webView = UIWebView()
-
+        // Reset UA
```
jhreis, line 61: All of these changes look a bit suspicious. We should pull down FF's latest `Shared/UserAgent.swift` code, and use that. If that is what this is, great, if not, let's get rid of it.

Client/Frontend/Browser/TabManager.swift
jhreis: This does not seem correct to me. Why is this? Is it something from FF upstream?

jhreis (changes requested):

Shared/UserAgentExtension.swift
```diff
@@ -0,0 +1,35 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
```
jhreis, line 5: Ditto: https://github.com/brave/brave-ios/pull/1368/files#r316382234

jhreis:
Closing in favor of: https://github.com/brave/brave-ios/pull/1425

PR #1401, opened 2019-08-15
Fixed #1397 by creating TabMO when opening popups.

### LINKED ISSUE #1397: tabs opened using "window.open" are lost when restarting brave
### Description: 

It seems like tabs that are opened via `target="_blank"` are lost when restarting brave.

### Steps to Reproduce 

**`Case #1:`**

  1. visit https://www.startpage.com
  2. search for `brave software` and click on a few results (these should open in new tabs)
  3. restart brave and you'll notice that the tabs opened via `startpage.com` have been lost

**`Case #2:`**

  1. visit https://www.gmail.com and login into an account
  2. open a few links from several emails (these should open in new tabs)
  3. restart brave and you'll notice that the tabs opened via `gmail.com` have been lost

**Actual result:** 

Example of the issue occurring --> https://youtu.be/TaI8-1fKMF4

**Expected result:**

Opened tabs should never be lost when restarting the browser

**Brave Version:** 

* Using `1.11 (19.08.01.15)` - Reproduced
* Using `1.11.1 (19.08.13.17)` - Reproduced

**Device details:** 

* Using `iPhone 6s+ iOS 12.4`

**Website problems only:**

### Additional Information

@LaurenWags mind going through to see if you can reproduce the issue with your `iPad`? CCing @jhreis @anthonypkeane

danishjafri88:
Fixes #1397 (Only in normal mode. TabMO will still not be created in private mode.)
TabMO was not created for popups. Hence when relaunching the app, the popups were not restored.

STR: Same as in the ticket #1397 

## Submitter Checklist:

## Reviewer Checklist:

  - [ ] `QA/(Yes|No)`
  - [ ] `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -298,6 +298,12 @@ class TabManager: NSObject {
 
     func addPopupForParentTab(_ parentTab: Tab, configuration: WKWebViewConfiguration) -> Tab {
         let popup = Tab(configuration: configuration, type: parentTab.type)
+        if parentTab.isPrivate {
+            // Creating random tab id for private mode, as we don't want to save to database.
+            popup.id = UUID().uuidString
+        } else {
+            popup.id = TabMO.create()
+        }
         configureTab(popup, request: nil, afterTab: parentTab, flushToDisk: true, zombie: false, isPopup: true)
 
         // Wait momentarily before selecting the new tab, otherwise the parent tab


Brandon-T (approved):

iccub (approved):

Client/Frontend/Browser/TabManager.swift
```diff
@@ -298,6 +298,12 @@ class TabManager: NSObject {
 
     func addPopupForParentTab(_ parentTab: Tab, configuration: WKWebViewConfiguration) -> Tab {
         let popup = Tab(configuration: configuration, type: parentTab.type)
+        if parentTab.isPrivate {
```
jhreis, line 301: This works, but is not a great solution, as it is just copy and pasting code. We should look at abstracting this, potentially, why not place inside Tab init?

jhreis (changes requested):

PR #1403, opened 2019-08-16
Revert "Fixed: #917 Updated the checks to set shields."

jhreis:
Broke shields.


iccub (approved):
I ran the branch on my device and ad blocking is working again

PR #1425, opened 2019-08-22
Added Desktop UA for iPad toggle.

### LINKED ISSUE #1290: User agent on iPad should default to a desktop one.

Brandon-T:
Fixes: #1290 .
Uses same UA as Firefox (on mobile it will contain: "FxiOS/Version Mobile/OSVersion".. with Desktop UA, it will be the same but without that string included (IE: Same as Safari which is what FF does).
Note: I do not do any UA modifications (The only one I did was to change from 10_11 to 10_15 which is what Safari on iOS 13 does.. but leaving it at 10_11 is the same anyway.. neither have any effects).

## Summary of Changes

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -421,7 +421,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
         // This only needs to be done once per runtime. Note that we use defaults here that are
         // readable from extensions, so they can just use the cached identifier.
         
-        if #available(iOS 13.0, *), UIDevice.isIpad {
+        if #available(iOS 13.0, *) {
             //iOS 13 iPad UA is the same as Safari's..
         } else {
             let defaults = UserDefaults(suiteName: AppInfo.sharedContainerIdentifier)!
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -420,8 +420,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
         // Set the UA for WKWebView (via defaults), the favicon fetcher, and the image loader.
         // This only needs to be done once per runtime. Note that we use defaults here that are
         // readable from extensions, so they can just use the cached identifier.
-        let defaults = UserDefaults(suiteName: AppInfo.sharedContainerIdentifier)!
-        defaults.register(defaults: ["UserAgent": firefoxUA])
+        
+        if #available(iOS 13.0, *), UIDevice.isIpad {
+            //iOS 13 iPad UA is the same as Safari's..
+        } else {
+            let defaults = UserDefaults(suiteName: AppInfo.sharedContainerIdentifier)!
+            defaults.register(defaults: ["UserAgent": firefoxUA])
+        }
 
         SDWebImageDownloader.shared().setValue(firefoxUA, forHTTPHeaderField: "User-Agent")
 
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -421,12 +421,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
         // This only needs to be done once per runtime. Note that we use defaults here that are
         // readable from extensions, so they can just use the cached identifier.
         
-        if #available(iOS 13.0, *) {
-            //iOS 13 iPad UA is the same as Safari's..
-        } else {
-            let defaults = UserDefaults(suiteName: AppInfo.sharedContainerIdentifier)!
-            defaults.register(defaults: ["UserAgent": firefoxUA])
-        }
+        let defaults = UserDefaults(suiteName: AppInfo.sharedContainerIdentifier)!
+        defaults.register(defaults: ["UserAgent": firefoxUA])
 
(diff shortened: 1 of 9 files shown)


Shared/UserAgent.swift
```diff
@@ -116,4 +116,26 @@ open class UserAgent {
 
         return String(userAgent)
     }
+    
+    public static func mobileApplicationName() -> String? {
```
iccub: this isn't used

Shared/UserAgent.swift
```diff
@@ -116,4 +116,26 @@ open class UserAgent {
...
+        return "FxiOS"
+    }
+    
+    public static func desktopApplicationName() -> String? {
```
iccub: this isn't used

iccub (changes requested):
Found one bug on iPhone:

- When you visit a website for the first time and tap on share menu it shows `request MOBILE site` not `desktop`. After you toggle to mobile-desktop then it shows correct item

iccub (approved):

Client/Application/AppDelegate.swift
```diff
@@ -424,8 +424,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
         // Set the UA for WKWebView (via defaults), the favicon fetcher, and the image loader.
         // This only needs to be done once per runtime. Note that we use defaults here that are
         // readable from extensions, so they can just use the cached identifier.
-        let defaults = UserDefaults(suiteName: AppInfo.sharedContainerIdentifier)!
-        defaults.register(defaults: ["UserAgent": firefoxUA])
+        
+        if #available(iOS 13.0, *), UIDevice.isIpad {
+            //iOS 13 iPad UA is the same as Safari's..
+        } else {
+            let defaults = UserDefaults(suiteName: AppInfo.sharedContainerIdentifier)!
```
jhreis: Is this registration needed for iOS 13 at all though? I didn't think iOS 13 used `UserDefaults` to assign UA.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1145,7 +1146,12 @@ class BrowserViewController: UIViewController {
     func restoreSpoofedUserAgentIfRequired(_ webView: WKWebView, newRequest: URLRequest) {
         // Restore any non-default UA from the request's header
         let ua = newRequest.value(forHTTPHeaderField: "User-Agent")
-        webView.customUserAgent = ua != UserAgent.defaultUserAgent() ? ua : nil
+        
+        if #available(iOS 13.0, *) {
+            webView.customUserAgent = Preferences.General.alwaysRequestDesktopSite.value == UserAgent.isDesktopUA(ua) ? nil : ua
+        } else {
+            webView.customUserAgent = ua != UserAgent.defaultUserAgent() ? ua : nil
```
jhreis: I know this was copied from what existed before, but this looks confusing with reversed logic from above, let's modify this so the conditionals look more identical:
```suggestion
            webView.customUserAgent = ua == UserAgent.defaultUserAgent() ? nil : ua
```

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -3074,6 +3086,15 @@ extension BrowserViewController: PreferencesObserver {
             setupTabs()
             updateTabsBarVisibility()
             updateApplicationShortcuts()
+        case Preferences.General.alwaysRequestDesktopSite.key:
+            tabManager.reset()
+            self.tabManager.reloadSelectedTab()
+            for tab in self.tabManager.allTabs where tab != self.tabManager.selectedTab {
```
jhreis: See: https://github.com/brave/brave-ios/pull/1368#discussion_r316375423

I think all of this logic is overkill. Let's run some tests without all of this looping logic (e.g. tab manager reset, and reloading selected tab). The consequence to this _should_ be lazy loaded alternative pages (e.g. when a user clicks on a new tab, it will be freshly loaded).

Not all tabs have webviews, so this could be a pretty bloated operation. Notice how `reset()` only resets tabs that have active webviews?

jhreis (changes requested):

Client/Application/AppDelegate.swift
Brandon-T: It's not needed for iOS 13 at all.

Brandon-T: Fixed.

Client/Frontend/Browser/BrowserViewController.swift

Client/Frontend/Browser/BrowserViewController.swift
Brandon-T: Fixed. Tested it and it works.

Behaviour: Current tab is reloaded no matter what. When the user taps on another tab, only that tab is reloaded. If they tap on another tab, that other tab is also reloaded (lazy loading).. so on and so forth. So it works the way you suggested :)

jhreis (approved):

jhreis:
@Brandon-T Perfect, just CI failing I think.

iccub:
Unit tests need a fix

PR #1460, opened 2019-08-30
Fix #1459: Replace UIWebView user agent creation with manual UA builder.

### LINKED ISSUE #1459: UIWebView deprecation notice
UIWebView is deprecated and may result in rejecting the app in App Store.
We have to remove it.
Currently we use it in UserAgent creation and KIF framework uses it.

iccub:
## Summary of Changes

The logic is based on Chromium's implementation https://chromium.googlesource.com/chromium/src.git/+/master/ios/web/common/user_agent.mm

This pull request fixes issue #1459 

## Submitter Checklist:

## Test Plan:

Verify new and old app versions have the same user agent.

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Cartfile
```diff
@@ -3,7 +3,6 @@ github "garvankeeley/Deferred"		"9.9.2"
 github "SnapKit/SnapKit"		"5.0.1"
 github "rs/SDWebImage"			"4.4.6"
 github "swisspol/GCDWebServer"		"3.5.2"
-github "kif-framework/KIF"		"v3.7.8"
```
iccub, line 6: We don't use it, and it contains UIWebView

Client.xcodeproj/project.pbxproj
```diff
@@ -6558,7 +6471,11 @@
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
 				INFOPLIST_FILE = UnitTests/Info.plist;
-				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
```
iccub, line 6637: XCode 11 did this automatically, apprently the format for search paths is a bit different now

SharedTests/UserAgentBuilderTests.swift
```diff
@@ -0,0 +1,30 @@
...
+        let webView = WKWebView(frame: CGRect(x: 0, y: 0, width: 100, height: 100))
+        
+        webView.evaluateJavaScript("navigator.userAgent") { ua, error in
+            wkWebViewUA = ua as! String
+            webViewUAExpectation.fulfill()
+        }
+        
+        wait(for: [webViewUAExpectation], timeout: 10)
+        
+        XCTAssertEqual(builderUA, wkWebViewUA)
```
iccub, line 26: This should give us enough confidence that our User Agent builder always match WKWebView's UA

Shared/UserAgentBuilder.swift
```diff
@@ -0,0 +1,80 @@
...
+        
+        let platform = UIDevice.current.userInterfaceIdiom == .pad ? "OS" : "iPhone OS"
+        
+        return "\(currentDevice); CPU \(platform) \(osVersion) like Mac OS X"
+    }
+    
+    var webkitVersion: UAVersions {
+        let _12_0 = OSVersionMap(majorVersion: 12, minorVersion: 0,
+                              uaVersions: UAVersions(safariVersion: "605.1",
+                                                     webkitVersion: "605.1.15"))
```
kylehickinson, line 54: Are these going to be important to manually update you think?

iccub: We have to.
We will see when it breaks thanks to the unit test

Chromium does it manually too I think https://chromium.googlesource.com/chromium/src.git/+/master/ios/web/common/user_agent.mm#46

iccub:
In general I tried to utilize WKWebView first, but it's asynchronous, and can't be easily synchronized.
If Chromium uses the manual process for UA creation we should be fine imo

Old vs new UA
Zrzut ekranu 2019-08-30 o 18 06 30

iccub:
And evaluating WKWebView's UA by javascript takes 2-3 seconds which is bad If we would like to synchronize it

kylehickinson:
Yeah that sucks. Is there a specific spot those version numbers can be read from easily aside from watching for Chromium changes?

iccub:
Could be done if we could subscribe to this commit history log, maybe there's github's mirror somewhere and we could subscribe to it when this file changes, no idea really

https://chromium.googlesource.com/chromium/src.git/+log/master/ios/web/common/user_agent.mm

kylehickinson:
What if we hard-code values we know (and update based on Chromium) and overwrite with WKWebView's value async whenever it finishes successfully?

iccub:
We can do that if we want to be extra safe. I think the unit test againt WKWebView should be enough though, and it's a bit less complex

@jhreis opinions on that welcome

kylehickinson (approved):
LGTM. Still wanna find a way to automate the versions somehow or at least notify us when they change rather than wait until CI tests catch it

jhreis (approved):

iccub:
There's issue for up to date UA already, linking it for reference https://github.com/brave/brave-ios/issues/838

iccub (commented):
I can't approve my PR, Joel's commit to add path version works fine on iOS 12.4.1

Issue #1478, opened 2019-09-03
Internal Issue


anthonypkeane:
https://github.com/brave/internal/issues/648

iccub:
Closed via https://github.com/brave/brave-ios/pull/1479

PR #1479, opened 2019-09-03
Custom client for ddg.

iccub:
Ref: internal#648

## Summary of Changes

This pull request fixes issue #1478 

## Submitter Checklist:

## Test Plan:

- Verify iOS device is in the United States region
- Launch Brave and use DDG as the default search engine (can go into private mode to make this easier)
- Perform a search
- Verify the URL contains `t=brave` in the URL bar (may need to long-press "copy" and paste somewhere else to see full URL)

- Change iOS device region to New Zealand
- Back in Brave perform another DDG search
- verify URL contains `t=braveed` (noticed additional `ed`)

- That's it. Cheers 🌮 

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/Search/OpenSearch.swift
+++ b/Client/Frontend/Browser/Search/OpenSearch.swift
@@ -4,6 +4,7 @@
 
 import UIKit
 import Shared
+import BraveShared
 import Fuzi
 
 private let TypeSearch = "text/html"
@@ -17,6 +18,8 @@ class OpenSearchEngine: NSObject, NSSecureCoding {
         static let qwant = "Qwant"
     }
     
+    static let defaultSearchClientName = "brave"
+    
     let shortName: String
     let engineID: String?
     let image: UIImage
@@ -26,10 +29,12 @@ class OpenSearchEngine: NSObject, NSSecureCoding {
 
     fileprivate let SearchTermComponent = "{searchTerms}"
     fileprivate let LocaleTermComponent = "{moz:locale}"
+    fileprivate let RegionalClientComponent = "{customClient}"
 
     fileprivate lazy var searchQueryComponentKey: String? = self.getQueryArgFromTemplate()
 
-    init(engineID: String?, shortName: String, image: UIImage, searchTemplate: String, suggestTemplate: String?, isCustomEngine: Bool) {
+    init(engineID: String?, shortName: String, image: UIImage, searchTemplate: String,
+         suggestTemplate: String?, isCustomEngine: Bool) {
         self.shortName = shortName
         self.image = image
         self.searchTemplate = searchTemplate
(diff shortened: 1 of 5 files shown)


jhreis:
I think some of the behaviors here aren't quite right, but that is mostly due to product requirements not being clear. I'm blocking this for now, and bumping the ticket out of the current milestone.

jhreis:
Will leave more formal review once we identify the exact, expected behavior.

iccub:
This should be ready for review @jhreis

Client/Application/AppDelegate.swift
```diff
@@ -273,6 +272,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
         }
         
         AdblockResourceDownloader.shared.startLoading()
+        Preferences.General.isFirstLaunch.value = false
```
jhreis: I don't think we should move this, if we ran into some weird issue, it could end up duplicating all default favorites. I think that is the only real thing that uses the `isFirstLaunch` bool.

Client/Frontend/Browser/Search/SearchEngines.swift
```diff
@@ -62,9 +62,14 @@ class SearchEngines {
         self.orderedEngines = getOrderedEngines()
     }
     
-    func setupDefaultRegionalSearchEngines() {
-        guard let region = Locale.current.regionCode,
-            let searchEngine = SearchEngines.defaultRegionSearchEngines[region] else { return }
+    func regionalSearchEngineSetup(for locale: Locale = Locale.current) {
```
jhreis, line 65: This isn't needed, as we never pass in `locale`, so maybe just use `setupDefaultRegionalSearchEngines` and internalize the local parameter?

Client/Frontend/Browser/Search/OpenSearch.swift
```diff
@@ -121,32 +128,47 @@ class OpenSearchEngine: NSObject, NSSecureCoding {
     /**
      * Returns the search suggestion URL for the given query.
      */
-    func suggestURLForQuery(_ query: String) -> URL? {
+    func suggestURLForQuery(_ query: String, locale: Locale = Locale.current) -> URL? {
```
jhreis, line 131: I'm not sure adding a new parameter is necessary here, now, since we never pass anything in.

jhreis (changes requested):

Client/Frontend/Browser/Search/SearchEngines.swift
iccub: I can remove it but this is good if we would want to add unit tests for it, much easier to test regional url params

iccub: It's easier to add unit tests with this param

Client/Frontend/Browser/Search/OpenSearch.swift
iccub: As with previous comment, this is good for unit test purposes.
Atm we don't test this method but in the future the query params can be different between search suggestion and a regular query

jhreis (approved):
Very nice! 😄

jumde:
lgtm

PR #1556, opened 2019-09-18
Fixed Rewards Onboarding UI #159

### LINKED ISSUE #159: Rewards Onboarding UI
## Description
Add Brave Rewards to the onboarding experience for iOS.
Product spec: https://docs.google.com/document/d/15ouI4hsKOduE3IrB-RkvPaPww_iqnc6MLmvSqoCu6Os/edit#

#### iPhone
The Shields screen now has a "Continue" button that takes the user to the next screen in onboarding.

ios 2
ios 3

The "Agree" button is disabled until the user checks the Terms checkbox.

ios 4

Tapping on the link to Terms displays our terms page in webview.

ios 5 - terms

The "Agree" button is enabled once the checkbox is checked.

ios 5

After the user taps "Agree", trigger ads onboarding with a count down timer.
Gradient values:
start #4F30AB
middle #FF1893
end #FA7250

ios 6

Show our custom notification (product spec: https://docs.google.com/document/d/1l_jXRt5o_RKBFiv3y3RHivrt5xGxSpzua-0OjTPc5I4/edit?ts=5d5d7079)

ios 7

If the user taps the notification, show https://brave.com/my-first-ad/

ios 10

If the user ignores the notification, show a start browsing screen.

ios 8

#### iPad
iPad designs are identical to iPhone, but displayed as a card on top of a gradient background.

12 9″ iPad Pro - screen 2
12 9″ iPad Pro - screen 3
12 9″ iPad Pro - screen 4
12 9″ iPad Pro - screen 5
12 9″ iPad Pro - screen 6
12 9″ iPad Pro - screen 8

#### Dark Mode
dark mode 1
dark mode 3
dark mode 4
dark mode 5

#### Animated transitions between screens
When the white panel changes heights between screens, the height change and content should be animated for a smooth transition, for example (this is on Android; in iOS we would maintain the rounded corners of the white panel):
welcome transition out

slide transition

higher res video versions:
[general transition.zip](https://github.com/brave/browser-android-tabs/files/3384480/general.transition.zip)

**Animation breakdown:** the content in the white panel fades out: start Opacity: 1, end Opacity: 1, duration 0.2s. The white bg resizes starting at T=0.1s to the appropriate height for the next screen, duration 0.3s.  The illustration and new text content fades in starting at T=0.3s: start Opacity: 0, end Opacity: 1, duration: 0.4s.

### New user, In ad region
Show search engine selection, Brave Shields explanation, Brave Rewards opt-in with Terms of Service, and ad onboarding.

user flow - ios onboarding ads available

### New user, Out of ad region
Show search engine selection, Brave Shields explanation, and Brave Rewards opt-in with Terms of Service.

user flow - ios onboarding ads unavailable

For this user segment, the text on the Brave Rewards should not mention ads: 

> Support your favorite websites and creators based on your attention.

ios ads unavailable

### Existing user, Brave Rewards OFF
Show Brave Rewards opt-in with Terms of Service, and ad onboarding.

user flow - ios onboarding BR off

### Existing user, Brave Rewards ON
Show ads opt-in with ad onboarding.

user flow - ios onboarding BR on

ios br on

## Assets
[png assets.zip](https://github.com/brave/brave-rewards-ios/files/3607975/png.assets.zip)

Lottie assets:

- Rewards: https://assets10.lottiefiles.com/packages/lf20_uX7EQp.json
Preview: https://lottiefiles.com/share/uX7EQp
- Ad notification: https://assets5.lottiefiles.com/packages/lf20_CawP3d.json
Preview: https://lottiefiles.com/share/CawP3d

Abstract link for more CSS and plain text: https://share.goabstract.com/382f3231-c578-468e-8be6-fca295b7fd79?sha=65d9fd8a0b273c12a47465c32e75c865c914954b

Brandon-T:
Fixes https://github.com/brave/brave-rewards-ios/issues/159

Completes the on-boarding flow for new and existing users.

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Client/Frontend/Browser/Onboarding/OnboardingWebViewController.swift
```diff
@@ -0,0 +1,279 @@
...
+            updateWebPageDomain()
+        }
+        
+        toolbar.forwardButton.isEnabled = webView.canGoForward
+        toolbar.forwardButton.tintColor = webView.canGoForward ? UX.buttonEnabledColor : UX.buttonDisabledColor
+    }
+    
+    @objc
+    private func onAction() {
+        print("Access Button Pressed..")
```
kylehickinson: What's this supposed to do?

Client/Frontend/Browser/Onboarding/OnboardingViewController.swift
```diff
@@ -18,6 +18,10 @@ class OnboardingViewController: UIViewController {
         return theme.isDark
     }
     
+    var themeColour: UIColor {
+        return theme.isDark ? UIColor(rgb: 0x343A40) : UIColor(rgb: 0xFFFFFF)
```
kylehickinson, line 24: Should these colors be defined somewhere?

Client/Frontend/Browser/Onboarding/OnboardingViewController.swift
```diff
@@ -18,6 +18,10 @@ class OnboardingViewController: UIViewController {
         return theme.isDark
     }
     
+    var themeColour: UIColor {
```
kylehickinson: This doesn't seem to update if dark mode is changed during the onboarding

danishjafri88:
No way to exit the T&C screen. A glitch after tapping start browsing and the the countdown looks a little blurred.

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
```diff
@@ -0,0 +1,71 @@
...
+    
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        
+        contentView.animate(from: 0.0, to: 1.0, duration: 3.0) { [weak self] in
+            self?.generateNotification()
+        }
+    }
+    
+    private func generateNotification() {
```
kylehickinson: This and all the UserNotifications related stuff should be removed and use the real my first ad stuff, which is already in `development`

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownView.swift
```diff
@@ -0,0 +1,370 @@
...
+        static let gradientPinkDark = UIColor(rgb: 0xFF1893)
+        static let gradientOrangeDark = UIColor(rgb: 0xFA7250)
+    }
+    
+    fileprivate let countdownLayer = CenteredTextLayer().then {
+        $0.font = UIFont.systemFont(ofSize: 54.0) as CTFont
+        $0.alignmentMode = .center
+    }
+    
+    private let gradientLayer = { () -> CAGradientLayer in
```
kylehickinson: Nit/Consistency: Switch all these to use `.then` i.e. 

```
private let gradientLayer = CAGradientLayer().then {
  $0.type = .conic
  // etc
}
```

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownView.swift
```diff
@@ -0,0 +1,370 @@
...
+            case adConfirmation
+        }
+        
+        var countdownText: String? {
+            get {
+                countdownView.countdownLayer.string as? String
+            }
+            
+            set {
+                countdownView.countdownLayer.string = newValue as NSString?
```
kylehickinson: Is the cast really necessary?

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownView.swift
```diff
@@ -0,0 +1,370 @@
...
+                $0.trailing.equalTo(self.safeArea.trailing)
+                $0.bottom.equalTo(self.safeArea.bottom)
+            }
+            
+            descriptionView.addSubview(descriptionStackView)
+            descriptionStackView.snp.makeConstraints {
+                $0.edges.equalToSuperview().inset(UX.descriptionContentInset)
+            }
+            
+            [descriptionView].forEach(mainStackView.addArrangedSubview(_:))
```
kylehickinson: No need to loop

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownView.swift
```diff
@@ -0,0 +1,370 @@
...
+        }
+    }
+    
+    @available(*, unavailable)
+    required init(coder: NSCoder) { fatalError() }
+    
+    func animate(from startOffset: CGFloat, to endOffset: CGFloat, duration: CFTimeInterval, completion: (() -> Void)? = nil) {
+        CATransaction.begin()
+        
+        countdownLayer.string = String(Int(duration)) as NSString
```
kylehickinson: Is the cast necessary here?

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownView.swift
```diff
@@ -0,0 +1,370 @@
...
+        shapeLayer.add(backgroundAnimation, forKey: "backgroundAnimation")
+        strokeLayer.add(strokeAnimation, forKey: "strokeAnimation")
+        strokeBallLayer.add(ballAnimation, forKey: "ballAnimation")
+        countdownLayer.add(countdownAnimation, forKey: "countdownAnimation")
+        
+        CATransaction.commit()
+    }
+}
+
+private class CenteredTextLayer: CATextLayer {
```
kylehickinson, line 386: Why do we need to use a CATextLayer instead of just a regular UILabel?

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+    
+    func userNotificationCenter(_ center: UNUserNotificationCenter, willPresent notification: UNNotification, withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
+        
+        defer { center.delegate = nil }
+        completionHandler([.alert])
+        
+        DispatchQueue.main.asyncAfter(deadline: .now() + 1.0) {
+            self.contentView.setState(.adConfirmation)
```
kylehickinson: When the user clicks the "My First Ad" it should open the my-first-ad URL as it does right now in the BVC. Are we going to be able to show this URL here? Or should we show it to them when they finally finish onboarding?

Client/Frontend/Browser/Onboarding/OnboardingAdsFinishedView.swift
```diff
@@ -0,0 +1,110 @@
...
+                $0.trailing.equalTo(self.safeArea.trailing)
+                $0.bottom.equalTo(self.safeArea.bottom)
+            }
+            
+            descriptionView.addSubview(descriptionStackView)
+            descriptionStackView.snp.makeConstraints {
+                $0.edges.equalToSuperview().inset(UX.descriptionContentInset)
+            }
+            
+            [descriptionView].forEach(mainStackView.addArrangedSubview(_:))
```
kylehickinson: No need for loop

Client/Frontend/Browser/Onboarding/OnboardingAdsFinishedView.swift
```diff
@@ -0,0 +1,110 @@
...
+
+            [UIView.spacer(.horizontal, amount: 0),
```
kylehickinson: I don't understand why we are using spacer views? Just use UIStackView's `setCustomSpacing`

Client/Frontend/Browser/Onboarding/OnboardingNavigationController.swift
```diff
@@ -127,3 +175,140 @@ extension OnboardingNavigationController {
...
+            tDetailsContent?.layer.pop_add($0, forKey: "alpha")
+        }
+        
+        DispatchQueue.main.asyncAfter(deadline: .now() + self.transitionDuration(using: transitionContext)) {
+            transitionContext.completeTransition(!transitionContext.transitionWasCancelled)
+        }
+    }
+    
+    func transitionDuration(using transitionContext: UIViewControllerContextTransitioning?) -> TimeInterval {
+
```
kylehickinson: whitespace

Client/Frontend/Browser/Onboarding/OnboardingRewardsAgreementView.swift
```diff
@@ -0,0 +1,223 @@
...
+                $0.trailing.equalTo(self.safeArea.trailing)
+                $0.bottom.equalTo(self.safeArea.bottom)
+            }
+            
+            descriptionView.addSubview(descriptionStackView)
+            descriptionStackView.snp.makeConstraints {
+                $0.edges.equalToSuperview().inset(UX.descriptionContentInset)
+            }
+            
+            [descriptionView].forEach(mainStackView.addArrangedSubview(_:))
```
kylehickinson: No need to loop

Client/Frontend/Browser/Onboarding/OnboardingRewardsAgreementView.swift
```diff
@@ -0,0 +1,223 @@
...
+
+            [skipButton, agreeButton, UIView.spacer(.horizontal, amount: 0)]
```
kylehickinson: Same comment about spacer views as above → use `UIStackView.setCustomSpacing(_:after:)`

Client/Frontend/Browser/Onboarding/OnboardingRewardsAgreementViewController.swift
```diff
@@ -0,0 +1,28 @@
...
+
+import Foundation
+
+class OnboardingRewardsAgreementViewController: OnboardingViewController {
+    private var contentView: View {
+        return view as! View // swiftlint:disable:this force_cast
+    }
+    
+    override func loadView() {
+        view = View(theme: theme, themeColour: themeColour)
```
kylehickinson: I would sync up with @iccub on the way we're supposed to do theming for onboarding

Client/Frontend/Browser/Onboarding/OnboardingRewardsView.swift
```diff
@@ -0,0 +1,132 @@
...
+                $0.trailing.equalTo(self.safeArea.trailing)
+                $0.bottom.equalTo(self.safeArea.bottom)
+            }
+            
+            descriptionView.addSubview(descriptionStackView)
+            descriptionStackView.snp.makeConstraints {
+                $0.edges.equalToSuperview().inset(UX.descriptionContentInset)
+            }
+            
+            [descriptionView].forEach(mainStackView.addArrangedSubview(_:))
```
kylehickinson: No need for loop

Client/Frontend/Browser/Onboarding/OnboardingRewardsView.swift
```diff
@@ -0,0 +1,132 @@
...
+
+            [skipButton, joinButton, UIView.spacer(.horizontal, amount: 0)]
```
kylehickinson, line 95: No spacers, same as above

Client/Frontend/Browser/Onboarding/OnboardingShieldsView.swift
```diff
@@ -67,38 +67,44 @@ extension OnboardingShieldsViewController {
...
                 $0.bottom.equalTo(self.safeArea.bottom)
-                $0.top.equalTo(self) // extend the view undeneath the safe area/notch
             }
             
+            descriptionView.addSubview(descriptionStackView)
             descriptionStackView.snp.makeConstraints {
                 $0.edges.equalToSuperview().inset(UX.descriptionContentInset)
             }
+            
+            [descriptionView].forEach(mainStackView.addArrangedSubview(_:))
```
kylehickinson: No need to loop

Client/Frontend/Browser/Onboarding/OnboardingShieldsView.swift
```diff
@@ -67,38 +67,44 @@ extension OnboardingShieldsViewController {
...
+            
+            [skipButton, continueButton, UIView.spacer(.horizontal, amount: 0)]
```
kylehickinson, line 99: No spacers

Client/Frontend/Browser/Onboarding/OnboardingWebViewController.swift
```diff
@@ -0,0 +1,279 @@
...
+                toolbar.urlLabel.textColor = UX.insecureWebPageColor
+            }
+        } else {
+            toolbar.secureIcon.isHidden = true
+            toolbar.urlLabel.textColor = UX.unknownWebPageColor
+        }
+    }
+    
+    private func updateBackForwardUI() {
+        toolbar.backButton.isEnabled = webView.canGoBack
```
kylehickinson, line 130: Remove all the delegate methods below that are implemented simply to update UI and just subscribe to relevant KVO as BrowserViewController does

kylehickinson (changes requested):

danishjafri88 (commented):
@kylehickinson beat me to the review :). Just one question, should the onboarding not show if the `NO_REWARDS` flag is set?

kylehickinson:
@danishjafri88 You're right, it probably shouldn't since you can't create a BraveRewards to do it.

@Brandon-T  Made me realize that this doesn't actually enable rewards (create a wallet) when you accept the terms like it should... So add that to the list 😛

You will also need to add a loader to the button which creates the wallet (check the rewards panel)

danishjafri88:
Another issue (Not very crucial imo): If the user backgrounds the app while the counter is running and he comes back after 3 seconds the timer will not update. This is because the CALayer animations pause on background. Maybe pause and resume, or put a timer to update ui as well?
ezgif com-resize

Client/Frontend/Browser/Onboarding/OnboardingWebViewController.swift
Brandon-T: No idea.. there is a button on the top right of the web-view that I have no information about.

This terms shows in its own web-view.

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownView.swift
Brandon-T: Can't animate UILabel's text property so I can't animate the countdown. If I use `CATextLayer`, I can animate the string property

Client/Frontend/Browser/Onboarding/OnboardingWebViewController.swift
kylehickinson: Oh, so this was a designed screen? I just thought you made it up 😆 I'd ping the designer who did it (Karen or James)

Client/Frontend/Browser/Onboarding/OnboardingViewController.swift
Brandon-T: They were already defined on development and I moved them to here instead.

Client/Frontend/Browser/Onboarding/OnboardingViewController.swift
Brandon-T: None of the Brave UI seems to update when I switch from dark to light mode and vice-versa. I've been killing the browser to refresh it. This is true for development right now.

I only "moved" the code that already exists on development.

kylehickinson: That's definitely not the case. I just launched the live version, and changed my phone to dark mode and it switches everything to dark right away

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownView.swift
Brandon-T: Fixed.

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
Brandon-T: > When the user clicks the "My First Ad" it should open the my-first-ad URL as it does right now in the BVC. Are we going to be able to show this URL here? Or should we show it to them when they finally finish onboarding?

The ad is supposed to show during the onboarding and not after. Not sure why all of the code is in BrowserViewController.

We have to ask the user if they saw an ad on the screen right after showing the ad.

Client/Frontend/Browser/Onboarding/OnboardingWebViewController.swift
Brandon-T: > No way to exit the T&C screen. A glitch after tapping start browsing and the the countdown looks a little blurred.
> ...
> ...

You can exit the terms and conditions by pressing the X button at the top left.. or by swiping down on the screen.

image

danishjafri88: @Brandon-T as you can see in the screenshot I don't have the nav bar, nor is it modal. it shows full screen

danishjafri88: ezgif com-resize-2
iPhone XS iOS 12  (simulator & device)

Client/Frontend/Browser/Onboarding/OnboardingRewardsView.swift
Brandon-T: All of these spacers are currently in dev. I just copied all of them to keep the code consistent and keep the UI the exact same as what's already there: https://github.com/brave/brave-ios/search?utf8=%E2%9C%93&q=UIView.spacer%28.vertical&type=

Even on the old on-boarding code it is there.

Removing them completely breaks the UI as the stackView is centered and the spacer has lower content hugging priority than other views so it takes up the remaining space allowing the desired views to be centered properly and spaced equally.

Client/Frontend/Browser/Onboarding/OnboardingAdsFinishedView.swift
Brandon-T: This doesn't work. You can't set custom spacing "before". It also depends on the layout and since this layout requires "centering", a blank view with lower hugging before allows it to achieve equal centering and proper spacing. Without it, the UI is completely broken.

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
Brandon-T: Fixed.

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
Brandon-T: The UX shows that the ad shows on top of the onboarding screens and when you tap, the onboarding is complete and the BVC shows. If you don't tap, you get the `adConfirmation` screen to tell you why you might have missed the ad.

Client/Frontend/Browser/Onboarding/OnboardingWebViewController.swift
Brandon-T: Fixed.

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownView.swift
kylehickinson: What are we animating? I don't see any animation of the actual text in simulator, it just goes from 3 to 2 instantly? And its blurry like @danishjafri88 mentioned probably because it's a CATextLayer.  Unless I'm missing something, lets just replace it with a UILabel

Client/Frontend/Browser/Onboarding/OnboardingAdsFinishedView.swift
kylehickinson: Not resolved for this one (accidental resolve or just missed it when pushing fixes?)

kylehickinson (changes requested):
As mentioned: Rewards has to actually be enabled and turned on before the user can go to the "waiting for first ad" screen

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
kylehickinson: I'm still seeing all the UserNotifications stuff?

Client/Frontend/Browser/Onboarding/OnboardingNavigationController.swift
```diff
@@ -53,13 +65,25 @@ class OnboardingNavigationController: UINavigationController {
...
             }
         }
         
         var type: OnboardingViewController.Type {
             switch self {
             case .searchEnginePicker: return OnboardingSearchEnginesViewController.self
             case .shieldsInfo: return OnboardingShieldsViewController.self
+            case .rewardsInfo: return OnboardingRewardsViewController.self
+            case .rewardsAgreement: return OnboardingRewardsAgreementViewController.self
+            case .adsCountdown: return OnboardingAdsCountdownViewController.self
```
kylehickinson, line 95: I think as per spec we shouldn't be showing the ads stuff unless the user can actually enable ads (`BraveAds.isSupportedRegion(Locale.current.identifier)` returning true)

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -245,7 +245,7 @@ class BrowserViewController: UIViewController {
     }
     
     // Display first ad when the user gets back to this controller if they havent seen one before
-    func displayMyFirstAdIfAvailable() {
+    func displayMyFirstAdIfAvailable(_ completion: (() -> Void)? = nil) {
```
kylehickinson: This change doesn't do anything? `completion` isn't called.

Either way, lets remove this change and simply copy/paste similar code into onboarding (barring figuring out how to show the my-first-ad URL).

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
```diff
@@ -0,0 +1,46 @@
...
+        contentView.invalidButton.addTarget(self, action: #selector(continueTapped), for: .touchDown)
+    }
+    
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        
+        contentView.animate(from: 0.0, to: 1.0, duration: 3.0) { [weak self] in
+            guard let self = self else { return }
+            
+            (UIApplication.shared.delegate as? AppDelegate)?.browserViewController.displayMyFirstAdIfAvailable({
```
kylehickinson: This won't actually show up on onboarding because its presented on the BVC (as talked about above)

kylehickinson (changes requested):

Client/Frontend/Browser/Onboarding/OnboardingViewController.swift
Brandon-T: I tested it a million times. It does NOT switch for me on simulator. In any case, I copied the same code used to theme the rest of the application into the onboarding. Hopefully it works. On simulator it does NOT show for me. On iPhone-X it does.

I was wondering why this was left out in the original on-boarding code in the first place..

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownView.swift
Brandon-T: Blurriness is fixed by setting the scale on the layer.

To keep things in sync with the rest of animations, it must be a CATextLayer. A UILabel's text cannot be animated so it won't be able to countdown with the rest of the animation, unless I add a timer and then manually change the text every tick.

Plus when timers go in the background they get expired so it's even worse than the current solution which works fine.

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
Brandon-T: Fixed. Max copy paste duplicate code..

Client/Frontend/Browser/BrowserViewController.swift

Client/Frontend/Browser/Onboarding/OnboardingNavigationController.swift
Brandon-T: Fixed.

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
```diff
@@ -0,0 +1,128 @@
...
+
+extension OnboardingAdsCountdownViewController {
+    
+    private func setupBraveRewards() {
+        #if NO_REWARDS
+        rewards = nil
+        rewardsObserver = nil
+        #else
+        RewardsHelper.configureRewardsLogs()
+        rewards = BraveRewards(configuration: .default)
```
kylehickinson: Hmmm, none of this is needed at all in onboarding. All we need is 1. to check that BraveAds is supported in this region, which doesn't require us make a new one, and 2. the `AdsViewController.displayFirstAd` call

In fact, the only place we need an actual rewards instance is when clicking "accept" or w/e the button is where we need to create the wallet and enable rewards (which I think this PR is still missing). This should be passed _into_ the onboarding controller since the BVC already makes that object (which you copied here)

Is there a reason you added all this and the rewards observers and everything?

kylehickinson (changes requested):

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
Brandon-T: Lol.. why does it seem like this is more complicated than it needs to be? I just need to show a notification and know when it has shown and if the user tapped it.

I already have the check for if ads are supported in the NavigationController which determines what screens will be shown. So that's done.

Anyway, it is all there because I need to know if the notification has shown, and if the user tapped it, through the use of `AdsNotificationHandler(ads: rewards.ads, presentingController: self)` which has a rewards.ads parameter.. The code is so tightly coupled I can't have one without the other. To get notified, I have to create a BraveRewards and then a NotificationHandler and then display the notification. But now I need a wallet before all of this which can only be done through BVC? OR does the callback for `displayFirstAd` notify me of the user's actions? Did I understand this correct?

If this is correct, I will add it now but it isn't clear. 

BVC is going to pass the onboarding navigation controller a `BraveRewards` instance 
 and it will then pass that to subsequent child onboarding controllers to generate a `notification` and do all the same logic that BVC already does?

kylehickinson: The `AdsNotificationHandler` is completely separate from the "My First Ad".  You know it showed up based on whether or not it was dismissed/timed out/was clicked/etc from the AdsViewController displayFirstAd method. Also the `notificationAdded` observer method isn't for ads, it's for the wallet notifications (insufficient funds, grant added, etc.)

The loading part, means the whole point of this onboarding is to get users to opt-in to Brave Rewards. To opt-in you have to create a wallet (like the panel does). Simply showing them a first ad is nothing but a special UI.

The spec for this is in Drive: `User Onboarding iOS - Phase 2` under **Cards**, point 3 & 4

kylehickinson: The code related to creating a wallet is in brave-rewards-ios: 
https://github.com/brave/brave-rewards-ios/blob/0f598b794003adc7231e6059b6f075ca798314cc/BraveRewardsUI/Extensions/BraveLedgerExtensions.swift#L89

This may have to become public though, otherwise its the same functionality

Brandon-T: This can't be right. I need to know if the ad shown as well.. not just if the user tapped it.
So that I can show the "Did you see an ad screen" or take them directly to the browser.. but it's fine I'll just remove that and show it always then.

Creating the wallet part is fine.. It will be in the next commit.

kylehickinson: If you need additional stuff from the My First Ad display the code for it is here and could be modified to support more than just tapping:

https://github.com/brave/brave-rewards-ios/blob/5fe6a61bd8735308bd6f079b21938a038fe22301/BraveRewardsUI/Ads/AdsViewController.swift#L308

There's nothing really to notify about "this was shown". We just assume it was shown because we're adding it to the view hierarchy ourselves. Maybe we could add a callback for the "animate out" part and if the user didn't tap it before it disappeared then we show the did you see it part, but just adding more complexity

kylehickinson: Just basically assume "My first ad" has nothing to do with real ads or the setup/logic around ads, because it doesn't function like it at all other than my first ad showing the same UI

Brandon-T: Done. Code is added to create a wallet. It is commented out for now until the interface is public.

Added loader when the user agrees. Removed all the rewards observer code and kept only ad creation code.

kylehickinson: Okay so make the changes in brave-rewards-ios and update the Carthage dep here (Cartfile.resolved update commited) when it's done and uncomment and will continue review then

kylehickinson: If you need additional updates for the first ad stuff probably best to do them at the same time first in brave-rewards-ios and then you can fix both when the dep is updated in this PR

Brandon-T: Done.

Client/Frontend/Browser/Onboarding/OnboardingRewardsAgreementViewController.swift
```diff
@@ -0,0 +1,56 @@
...
+    private func onAgreed() {
+        let titleColour = contentView.agreeButton.titleColor(for: .normal)
+        contentView.agreeButton.setTitleColor(.clear, for: .normal)
+        contentView.agreeButton.addSubview(loadingView)
+        loadingView.snp.makeConstraints {
+            $0.center.equalToSuperview()
+        }
+        
+        loadingView.startAnimating()
+        rewards?.ledger.createWallet { [weak self] error in
```
kylehickinson: Should be `createWalletAndFetchDetails`

kylehickinson (changes requested):
Needs updated brave-rewards-ios dev (Cartfile.resolved file)

Client/Frontend/Browser/Onboarding/OnboardingRewardsAgreementViewController.swift
kylehickinson: Oh, and we should guard this from being able to be tapped multiple times (Check whether or not loading view is already animating, or a bool flag for instance)

Client/Frontend/Browser/Onboarding/OnboardingRewardsAgreementViewController.swift
```diff
@@ -0,0 +1,61 @@
...
+        }
+        
+        loadingView.startAnimating()
+        rewards?.ledger.createWalletAndFetchDetails { [weak self] _ in
+            guard let self = self else { return }
+
+            self.loadingView.stopAnimating()
+            self.loadingView.removeFromSuperview()
+            self.contentView.agreeButton.setTitleColor(titleColour, for: .normal)
+            self.continueTapped()
```
kylehickinson: Shouldn't we check for an error here before automatically moving them on?

kylehickinson (changes requested):
Found some bugs while runtime testing:
1. After creating a wallet within onboarding, the Brave Rewards icon doesn't work. Tapping it yields no panel popup. Restarting app fixes it
2. As per Slack, selecting "I didn't see an ad" points to a screen which doesn't make sense in our case
3. The circle for "your ad will appear in…" is blurry and could use the `scale` treatment that the text layer got
4. The ••• icon in the web view still doesnt do anything. (Edit: as per Slack, to be removed)
5. The first ad appears incorrectly on iPad: IMG_0106

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownView.swift
```diff
@@ -0,0 +1,390 @@
...
+        }
+        
+        func setState(_ state: State) {
+            switch state {
+            case .countdown:
+                titleLabel.isHidden = false
+                buttonsStackView.isHidden = true
+                countdownView.isHidden = false
+                titleLabel.text = Strings.OBAdsTitle
+                break
```
iccub: Remove break

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownView.swift
```diff
@@ -0,0 +1,390 @@
...
+                
+            case .finished:
+                titleLabel.isHidden = false
+                buttonsStackView.isHidden = false
+                countdownView.isHidden = true
+                titleLabel.text = Strings.OBCompleteTitle
+                break
```
iccub: remove break

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownView.swift
```diff
@@ -0,0 +1,390 @@
...
+
+            [UIView.spacer(.vertical, amount: 0),
+             finishedButton,
+             UIView.spacer(.vertical, amount: 20)]
+                .forEach(buttonsStackView.addArrangedSubview(_:))
+            
+            [titleLabel, countdownView, buttonsStackView].forEach(descriptionStackView.addArrangedSubview(_:))
+            
+            countdownView.snp.makeConstraints {
+                $0.width.equalTo(156.0)
```
iccub: I believe you can use snapkit's `size` property to set both width and height if they are the same.
This constant could also go in `UX` struct

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownView.swift
```diff
@@ -0,0 +1,390 @@
...
+                $0.height.equalTo(156.0)
+            }
+        }
+        
+        func applyTheme(_ theme: Theme) {
+            descriptionView.backgroundColor = OnboardingViewController.colourForTheme(theme)
+            titleLabel.textColor = theme.isDark ? .white : .black
```
iccub: Change `textColor` to `appearanceTextColor` for proper theming on iOS 12.

Also replace `.white` and `.black` with `theme.colors.tints.home`, and please test if it works correctly(UI looks good)

kylehickinson (changes requested):
We should disable the "skip" button while it is creating a wallet during the Brave Rewards step

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownView.swift
```diff
@@ -0,0 +1,390 @@
...
+    }
+    
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+    }
+    
+    override func layoutSubviews() {
+        super.layoutSubviews()
+
+        //Layout shapeLayer
```
iccub, line 255: In general, we add a space after a comment block

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+        
+        //Countdown timer
+        contentView.countdownText = "3"
+        contentView.finishedButton.addTarget(self, action: #selector(continueTapped), for: .touchDown)
+    }
+    
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        
+        if let time = timeSinceAnimationStarted, Date().timeIntervalSince(time) >= 3.0 {
```
iccub: I think we should extract this `3` number to a integer constant and then cast it to string or double, as we use it in few places here

kylehickinson:
Also the animation circle thing is still blurry despite the content scale lines being added

Client/Frontend/Browser/Onboarding/OnboardingRewardsAgreementViewController.swift
```diff
@@ -0,0 +1,69 @@
...
+    
+    override func loadView() {
+        view = View(theme: theme)
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        contentView.agreeButton.addTarget(self, action: #selector(onAgreed), for: .touchDown)
+        contentView.skipButton.addTarget(self, action: #selector(skipTapped), for: .touchDown)
```
kylehickinson: These should be `touchUpInside` instead of `touchDown`

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+    override func loadView() {
+        view = View(theme: theme)
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        //Countdown timer
+        contentView.countdownText = "3"
+        contentView.finishedButton.addTarget(self, action: #selector(continueTapped), for: .touchDown)
```
kylehickinson: These should be `touchUpInside` instead of `touchDown`

Client/Frontend/Browser/Onboarding/OnboardingNavigationController.swift
```diff
@@ -19,6 +22,12 @@ protocol OnboardingControllerDelegate: class {
     func onboardingCompleted(_ onboardingController: OnboardingNavigationController)
 }
 
+enum OnboardingViewAnimationID: Int {
+    case background = 1
+    case details = 2
```
iccub: you don't need to declare `2,3...` if you have assigned `1` to first case, it's inferred
You can leave it if you find it more readable though

kylehickinson:
> After creating a wallet within onboarding, the Brave Rewards icon doesn't work. Tapping it yields no panel popup. Restarting app fixes it

This bug is also still happening.

Client/Frontend/Browser/Onboarding/OnboardingRewardsAgreementView.swift
```diff
@@ -0,0 +1,231 @@
...
+            
+            $0.textAlignment = .center
+        }
+        
+        private lazy var textStackView = UIStackView().then { stackView in
+            stackView.axis = .vertical
+            stackView.spacing = 8
+            let descriptionStackView =  UIStackView(arrangedSubviews: [descriptionLabel, descriptionCheckbox]).then {
+                $0.alignment = .center
+                $0.spacing = 65.0
```
iccub, line 98: Will this look good on all devices?
I guess the goal here is to have description aligned to left and checkbox aligned to right?

Client/Frontend/Browser/Onboarding/OnboardingRewardsAgreementView.swift
```diff
@@ -0,0 +1,231 @@
...
+        
+        private let buttonsStackView = UIStackView().then {
+            $0.distribution = .equalCentering
+        }
+        
+        private func updateDescriptionLabel() {
+            descriptionLabel.attributedText = {
+                let text = NSMutableAttributedString(string: Strings.OBRewardsAgreementDetail, attributes: [
+                    .font: UIFont.systemFont(ofSize: 16, weight: UIFont.Weight.regular),
+                    .foregroundColor: titleLabel.textColor!
```
iccub: we could unwrap it at the beginning of the method, it's used in few places here

or make something with a fallback like `let color = titleLabel.textColor : .black`

Client/Frontend/Browser/Onboarding/OnboardingRewardsAgreementView.swift
```diff
@@ -0,0 +1,231 @@
...
+                .forEach(buttonsStackView.addArrangedSubview(_:))
+            
+            [textStackView, buttonsStackView].forEach(descriptionStackView.addArrangedSubview(_:))
+            
+            descriptionCheckbox.addTarget(self, action: #selector(onTermsAccepted(_:)), for: .touchUpInside)
+        }
+        
+        func applyTheme(_ theme: Theme) {
+            descriptionView.backgroundColor = OnboardingViewController.colourForTheme(theme)
+            titleLabel.textColor = theme.isDark ? .white : .black
```
iccub: same as other comment, use `appearanceTextColor` and `theme.color.tints.home`

Client/Frontend/Browser/Onboarding/OnboardingRewardsAgreementView.swift
```diff
@@ -0,0 +1,231 @@
...
+            updateDescriptionLabel()
+            if theme.isDark {
+                descriptionCheckbox.setImage(#imageLiteral(resourceName: "checkbox_off_dark"), for: .normal)
```
iccub: can be simplified to `let image = theme.isDark ? img1 : img2`
and then set the image

Brandon-T:
> > After creating a wallet within onboarding, the Brave Rewards icon doesn't work. Tapping it yields no panel popup. Restarting app fixes it
> 
> This bug is also still happening.


https://i.imgur.com/O0ND3um.mp4

Works for me 100% of the time..
That button is also not part of the onboarding flow, so I can't see how it affects it in any way.

For the blur part, I added rasterization scaling but I'm not seeing any blur on my end so if it doesn't fix it, I'm out of ideas. However, according to S/O this is the solution.

Client/Frontend/Browser/Onboarding/OnboardingRewardsAgreementView.swift
```diff
@@ -0,0 +1,231 @@
...
+        @objc
+        private func onTermsAccepted(_ button: UIButton) {
+            button.isSelected.toggle()
+            
+            agreeButton.backgroundColor = button.isSelected ? BraveUX.BraveOrange : BraveUX.BraveOrange.withAlphaComponent(0.7)
+            agreeButton.isEnabled = button.isSelected
+        }
+        
+        @objc
+        func textView(_ textView: UITextView, shouldInteractWith URL: URL, in characterRange: NSRange, interaction: UITextItemInteraction) -> Bool {
```
iccub: Could we move it and `UITextViewDelegate` to an extension below?

Client/Frontend/Browser/Onboarding/OnboardingRewardsView.swift
```diff
@@ -0,0 +1,137 @@
...
+                .forEach(buttonsStackView.addArrangedSubview(_:))
+            
+            [textStackView, buttonsStackView].forEach(descriptionStackView.addArrangedSubview(_:))
+        }
+        
+        func applyTheme(_ theme: Theme) {
+            descriptionView.backgroundColor = OnboardingViewController.colourForTheme(theme)
+            titleLabel.textColor = theme.isDark ? .white : .black
```
iccub: see my other comments about this

Client/Frontend/Browser/Onboarding/OnboardingRewardsView.swift
```diff
@@ -0,0 +1,137 @@
...
+            mainStackView.addArrangedSubview(descriptionView)
+
+            [skipButton, joinButton, UIView.spacer(.horizontal, amount: 0)]
+                .forEach(buttonsStackView.addArrangedSubview(_:))
+            
+            [textStackView, buttonsStackView].forEach(descriptionStackView.addArrangedSubview(_:))
+        }
+        
+        func applyTheme(_ theme: Theme) {
+            descriptionView.backgroundColor = OnboardingViewController.colourForTheme(theme)
```
iccub: rename `colourForTheme` to `colorForTheme`

Client/Frontend/Browser/Onboarding/OnboardingSearchEnginesView.swift
```diff
@@ -120,6 +130,14 @@ extension OnboardingSearchEnginesViewController {
         @available(*, unavailable)
         required init(coder: NSCoder) { fatalError() }
         
+        func applyTheme(_ theme: Theme) {
+            containerView.backgroundColor = OnboardingViewController.colourForTheme(theme)
+            searchEnginesTable.backgroundView?.backgroundColor = OnboardingViewController.colourForTheme(theme)
+            titlePrimary.textColor = theme.isDark ? .white : .black
```
iccub: see other comment about theming this

Client/Frontend/Browser/Onboarding/OnboardingShieldsView.swift
```diff
@@ -67,38 +73,52 @@ extension OnboardingShieldsViewController {
...
+            .forEach(buttonsStackView.addArrangedSubview(_:))
+            
+            [textStackView, buttonsStackView].forEach(descriptionStackView.addArrangedSubview(_:))
+        }
+        
+        func applyTheme(_ theme: Theme) {
+            descriptionView.backgroundColor = OnboardingViewController.colourForTheme(theme)
+            textStackView.arrangedSubviews.forEach({
+                if let label = $0 as? UILabel {
+                    label.textColor = theme.isDark ? .white : .black
```
iccub: see other comment about theming this

Client/Frontend/Browser/Onboarding/OnboardingViewController.swift
```diff
@@ -5,21 +5,37 @@
...
+    var theme: Theme
     
     /// Whether the on-boarding is dark or not, based solely on passed in theme
     ///  Added explicitly to make removal easier in the future if just `theme` is used
     var dark: Bool {
         return theme.isDark
     }
     
-    init(profile: Profile, theme: Theme) {
+    static func colourForTheme(_ theme: Theme) -> UIColor {
```
iccub: rename to `color` without `u`

iccub (changes requested):
Just leaving some smaller feedback and code style concerns.
Kyle is doing good job reviewing this PR

Cartfile.resolved
```diff
@@ -3,8 +3,8 @@ github "DaveWoodCom/XCGLogger" "7.0.0"
 github "Sidetalker/onepassword-app-extension" "68a9d932a6373ca045984fbf4fb9a25aee43284c"
 github "SnapKit/SnapKit" "5.0.1"
 github "SwiftyJSON/SwiftyJSON" "5.0.0"
-github "airbnb/lottie-ios" "531eb22d47a9439a2344fdaf96832189f2c6e925"
-github "brave/brave-rewards-ios" "092cd4e52d9d821c7186de941adc6be8126baf23"
+github "airbnb/lottie-ios" "860124f1e8fe2163a87c225a931132fc57324b26"
```
kylehickinson: Is this change to Lottie library intentional?

Brandon-T: When I updated carthage, it also updated. If we don't want it, I can remove it but it works fine.

kylehickinson: Probably best to double check what changed in the update on `airbnb/lottie-ios` side.

Also you can update a single lib by passing the name into carthage update like: `carthage update brave-rewards-ios <other options>`

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+
+extension OnboardingAdsCountdownViewController {
+    
+    private func displayMyFirstAdIfAvailable(_ completion: ((AdsNotificationHandler.Action) -> Void)? = nil) {
+        if Preferences.Rewards.myFirstAdShown.value { return }
+
+        if BraveAds.isSupportedRegion(Locale.current.identifier) {
+            DispatchQueue.main.asyncAfter(deadline: .now() + 3) {
```
kylehickinson: This 3 second delay is not needed, since you already wait until the animation completes which is 3 seconds long

kylehickinson (changes requested):

kylehickinson:
Missing flows from `Existing User Onboarding iOS` spec in Drive:

- User has already seen all onboarding, rewards are off and ads become available in users region → User should be shown the onboarding for just rewards opt-in and my first ad
- User has already seen regular onboarding, and ads are available in users region → User should be shown the onboarding for just rewards opt-in and my first ad

Brandon-T:
> * rewards are off and ads become available in users region

This is already there:

```swift
case .existingUser: 
    return BraveAds.isSupportedRegion(Locale.current.identifier) ? [.rewardsInfo, .rewardsAgreement, .adsCountdown] : []
```

Existing user: Once rewards is supported in their area via `BraveAds.isSupportedRegion(Locale.current.identifier)`, show `rewardsInfo`, `agreement`, and `adsCountdown`.

kylehickinson:
Well I tested it after changing my region back to US and it didn't open onboarding

Brandon-T:
> Well I tested it after changing my region back to US and it didn't open onboarding

I just noticed.. the code had a "TBD" in it from the previous developer. I added the extra flows.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -679,10 +679,11 @@ class BrowserViewController: UIViewController {
             onboarding.onboardingDelegate = self
             present(onboarding, animated: true)
         } else if BraveAds.isSupportedRegion(Locale.current.identifier) {
+            let isAdsEnabled = rewards?.ads.isEnabled ?? false
```
kylehickinson: Is this to check if rewards is off? If so, I'd do `rewards?.ledger.isEnabled == false`

kylehickinson (changes requested):

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
kylehickinson: @Brandon-T You marked this resolved but I don't see the change… unresolving for now

Brandon-T: Hmm for some reason it didn't go through when I committed via Xcode. Should be there this time.

kylehickinson:
Just have to get verification on what to do when the user does have Rewards enabled already and then ads becomes available in their region. At the moment enabling Rewards enables ads (just as it did in Desktop where I copied the code from), but the spec says the onboarding should come up again, which means we'd need to change it to only enable ads by default if their region is supported. And then this second onboarding would just turn on ads, not enable rewards, which doesn't require ToS or anything (cc @anthonypkeane) 

Probably will have further discussion

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -667,17 +669,81 @@ class BrowserViewController: UIViewController {
...
+        // 1. Existing user.
+        // 2. The user skipped onboarding before.
+        // 3. 60 days have passed since they last saw onboarding.
+        if Preferences.General.basicOnboardingCompleted.value == OnboardingState.skipped.rawValue {
+
+            guard let daysUntilNextPrompt = Preferences.General.basicOnboardingNextOnboardingPrompt.value else {
+                return
+            }
+            
+            // 60 days has passed since the user last saw the onboarding.. it's time to show the onboarding again..
```
kylehickinson, line 685: Is the 60 days for basic onboarding as well?

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -667,17 +669,81 @@ class BrowserViewController: UIViewController {
...
+                
+                Preferences.General.basicOnboardingNextOnboardingPrompt.value = Date(timeIntervalSinceNow: BrowserViewController.onboardingDaysInterval * 24 * 60 * 60)
+            }
+            
+            return
+        }
+        
+        // 1. Rewards are on/off (existing user)
+        // 2. User hasn't seen the rewards part of the onboarding yet.
+        if Preferences.General.basicOnboardingCompleted.value == OnboardingState.completed.rawValue && Preferences.General.basicOnboardingProgress.value == OnboardingProgress.searchEngine.rawValue {
```
kylehickinson: Lol these if statements are getting long, can we move each condition onto a new line 😂

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -667,17 +669,81 @@ class BrowserViewController: UIViewController {
...
+                ) else { return }
+            
+            onboarding.onboardingDelegate = self
+            present(onboarding, animated: true)
+        }
+        
+        // 1. Rewards are on/off (existing user)
+        // 2. Ads are now available
+        // 3. User hasn't seen the ads part of onboarding yet
+        if BraveAds.isSupportedRegion(Locale.current.identifier) && Preferences.General.basicOnboardingCompleted.value == OnboardingState.completed.rawValue && Preferences.General.basicOnboardingProgress.value != OnboardingProgress.ads.rawValue {
```
kylehickinson: Same as above

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -667,17 +669,81 @@ class BrowserViewController: UIViewController {
...
+                    profile: profile,
+                    onboardingType: isRewardsEnabled ? .existingUserRewardsOn : .existingUserRewardsOff,
+                    rewards: rewards,
+                    theme: Theme.of(tabManager.selectedTab)
+                    ) else { return }
+                
+                onboarding.onboardingDelegate = self
+                present(onboarding, animated: true)
+                
+                Preferences.General.basicOnboardingNextOnboardingPrompt.value = Date(timeIntervalSinceNow: BrowserViewController.onboardingDaysInterval * 24 * 60 * 60)
```
kylehickinson: I think there's some shared code that lets us do like `60.toDays` or something, can you double check that exists

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -3104,9 +3170,19 @@ extension BrowserViewController: OnboardingControllerDelegate {
...
+            Preferences.General.basicOnboardingNextOnboardingPrompt.value = nil
         default: break
         }
         
         onboardingController.dismiss(animated: true)
     }
+    
+    func onboardingSkipped(_ onboardingController: OnboardingNavigationController) {
+        Preferences.General.basicOnboardingCompleted.value = OnboardingState.skipped.rawValue
+        Preferences.General.basicOnboardingNextOnboardingPrompt.value = Date(timeIntervalSinceNow: BrowserViewController.onboardingDaysInterval * 24 * 60 * 60)
```
kylehickinson: Same as above

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -3104,9 +3170,19 @@ extension BrowserViewController: OnboardingControllerDelegate {
...
+        onboardingController.dismiss(animated: true)
+    }
+    
+    // 60 days until the next time the user sees the onboarding..
+    static let onboardingDaysInterval: Double = 60.0
```
kylehickinson: Or maybe even just turn this into `60.days` or w/e it is and don't do any conversion above

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
```diff
@@ -0,0 +1,97 @@
...
+
+import Foundation
+import BraveRewards
+import BraveRewardsUI
+import BraveShared
+
+class OnboardingAdsCountdownViewController: OnboardingViewController, UNUserNotificationCenterDelegate {
+    
+    private struct UX {
+        static let animationTime = 5.0
```
kylehickinson: As mentioned in Slack, this should be 3 here, and we'll update ad automatic dismissal duration elsewhere

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
```diff
@@ -0,0 +1,97 @@
...
+                } else {
+                    //User saw the ad.. they interacted with it.. onboarding is finished.
+                    self.continueTapped()
+                }
+            }
+        }
+    }
+    
+    override func continueTapped() {
+        Preferences.Rewards.myFirstAdShown.value = true
```
kylehickinson: We should probably set this after displaying the view, incase the user for some reason force quits the app before clicking continue or something stupid like that and then it'd show again if they ever disabled/reenabled Brave Rewards

Client/Frontend/Browser/Onboarding/OnboardingNavigationController.swift
```diff
@@ -30,59 +42,90 @@ class OnboardingNavigationController: UINavigationController {
...
         modalPresentationStyle = UIDevice.current.userInterfaceIdiom == .phone ? .fullScreen : .formSheet
         
         if #available(iOS 13.0, *) {
             // Prevent dismissing the modal by swipe
             isModalInPresentation = true
         }
         preferredContentSize = UX.preferredModalSize
+        
+        let backgroundView = UIView().then {
+            $0.backgroundColor = #colorLiteral(red: 0.1176470588, green: 0.1254901961, blue: 0.1607843137, alpha: 1)
```
kylehickinson, line 122: This color usage needed? Not sure if we replace it based on theme?

Client/Frontend/Browser/Onboarding/OnboardingViewController.swift
```diff
@@ -87,4 +108,17 @@ class OnboardingViewController: UIViewController {
...
     }
+    
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        if #available(iOS 13.0, *) {
+            if UITraitCollection.current.userInterfaceStyle != previousTraitCollection?.userInterfaceStyle {
+                theme = OnboardingViewController.getUpdatedTheme()
+                applyTheme(theme)
+            }
+        }
+    }
```
kylehickinson, line 115: @iccub / @jhreis This OK? I'm not super familiar with the theming work

Client/Frontend/Browser/Onboarding/OnboardingViewController.swift
```diff
@@ -5,21 +5,37 @@
...
+    
+    static func getUpdatedTheme() -> Theme {
+        let fallback = Theme.DefaultTheme.light.rawValue
+        if #available(iOS 13.0, *) {
+            let isDark = UITraitCollection.current.userInterfaceStyle == .dark
+            return Theme.from(id: isDark ? Theme.DefaultTheme.dark.rawValue : fallback)
+        }
+        return Theme.from(id: fallback)
+    }
```
kylehickinson: @iccub @jhreis Same here

kylehickinson (changes requested):

Client/Frontend/Browser/BrowserViewController.swift
Brandon-T: It's actually not clarified if it's for basic or not but it does say:

"If the user SKIPS" the on-boarding, we show them again in 60 days. I'm not sure if this applies to search engine but I can't see why it won't since it's also onboarding. I'll ask.

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
Brandon-T: This part is weird. If I don't do this and they force quit before completing the onboarding, onboarding is technically no longer completed.. So it will popup the next time they launch in order for them to complete it.. However, they will NEVER see the popup because of this boolean.. Hence why I put it only when they press continue.

Client/Frontend/Browser/Onboarding/OnboardingNavigationController.swift
Brandon-T: It's needed. It was there before my time on Onboarding. I just "moved" it out of each controller into just one parent controller.

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
kylehickinson: Ah I see… so if they force-quit in the middle of onboarding they'd see onboarding again?

Client/Frontend/Browser/Onboarding/OnboardingViewController.swift
Brandon-T: This is how it's done everywhere else if you search for `applyTheme`

Client/Frontend/Browser/Onboarding/OnboardingViewController.swift
Brandon-T: This code is copy pasted into onboarding from 

```
static func of(_ tab: Tab?) -> Theme {
```

in order to get the most recent theme info.. This is done because `Theme` is a struct (copy) and does NOT update automatically (class reference) if the user goes and switches the system to light/dark mode.

**EDIT**: I changed it to `Theme.of` which does the exact same thing so I don't duplicate code..

Client/Frontend/Browser/BrowserViewController.swift

Client/Frontend/Browser/BrowserViewController.swift

Client/Frontend/Browser/Onboarding/OnboardingAdsCountdownViewController.swift
Brandon-T: Yeah. This is what was originally written before I took over the PR. Since they force quit, the onboarding state was never updated so they need to actually finish it or skip it if they don't want to see it.

```
/// If user kills the app before completing onboarding, it should be treated as unfinished.
static let basicOnboardingCompleted = Option<Int>(key: ...)
```

kylehickinson: Gotcha, ok this is fine then

Client/Frontend/Browser/Onboarding/OnboardingNavigationController.swift
```diff
@@ -30,59 +42,90 @@ class OnboardingNavigationController: UINavigationController {
...
         
         /// Returns a list of onboarding screens for given type.
         /// Screens should be sorted in order of which they are presented to the user.
         fileprivate var screens: [Screens] {
+            #if NO_REWARDS
             switch self {
-            case .newUser: return [.searchEnginePicker, .shieldsInfo, /* .rewardsInfo, .adsInfo */]
-            case .existingUser: return [/* .rewardsInfo, .adsInfo */]
+            case .newUser: return [.searchEnginePicker, .shieldsInfo]
+            case .existingUser: return []
```
kylehickinson: FYI: This line does not compile

`Type 'OnboardingNavigationController.OnboardingType' has no member 'existingUser'`

kylehickinson (changes requested):

kylehickinson:
Can we increase the hit box on the checkbox when accepting ToS for Rewards? It's a bit hard to hit.

anthonypkeane:
> Can we increase the hit box on the checkbox when accepting ToS for Rewards? It's a bit hard to hit.

HIG is 44px, any hit space around the box from 32-44 should be fine.

kylehickinson (changes requested):
Hit this weird state, not sure if its because I've had it installed in the past or what, but it went to the "ads available" flow, and when clicked get started it went here but no ad shows up. Maybe we should just ignore check for "my first ad" when going through this countdown page? If the user somehow does end up on it I'd rather then see something. Also in this state there's no way to get out, since no ad appears, therefore you have no automatic dismissal to trigger the next step

Simulator Screen Shot - iPhone 11 Pro - 2019-10-02 at 16 40 28

kylehickinson:
~Also this should say "Get started" instead of "Join" (because you aren't joining anything). I know the flow says join but I believe we said otherwise on the call~ Sticking with Join as below

anthonypkeane:
> Also this should say "Get started" instead of "Join" (because you aren't joining anything). I know the flow says join but I believe we said otherwise on the call
> 


Lets leave this as-is for now and test on the device. I know @karenkliu had good reason for that word choice.

Client.xcodeproj/project.pbxproj
```diff
@@ -1692,13 +1701,22 @@
...
 		5E34780F22D7A1D200B0D5F8 /* ResourceDownloadManager.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ResourceDownloadManager.swift; sourceTree = "<group>"; };
 		5E4845BF22DE381200372022 /* WindowRenderHelper.js */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.javascript; path = WindowRenderHelper.js; sourceTree = "<group>"; };
 		5E4845C122DE3DF800372022 /* WindowRenderHelperScript.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = WindowRenderHelperScript.swift; sourceTree = "<group>"; };
 		5E9288C922DF864C007BE7A6 /* TabSessionTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TabSessionTests.swift; sourceTree = "<group>"; };
 		5EAB745B230456F0006055F5 /* BookmarkValidation.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BookmarkValidation.swift; sourceTree = "<group>"; };
-		5EB57D0122FDC0CB00A07325 /* FullscreenHelper.js */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.javascript; path = FullscreenHelper.js; sourceTree = "<group>"; };
```
iccub, line 1701: This file keeps getting removed in all PRs for some reason, should it be removed here?

Client/Frontend/Browser/Onboarding/OnboardingNavigationController.swift
```diff
@@ -127,3 +197,156 @@ extension OnboardingNavigationController {
...
+        fBackground?.alpha = 1.0
+        fDetails?.alpha = 1.0
+        fDetails?.layer.maskedCorners = [.layerMinXMinYCorner, .layerMaxXMinYCorner]
+        fDetailsContent?.alpha = 1.0
+        
+        tBackground?.alpha = 0.0
+        tDetails?.alpha = 0.0
+        tDetailsContent?.alpha = 0.0
+        
+        //guard let tDetailsFrame = tDetails?.superview?.convert(tDetails?.frame ?? .zero, to: container) else { return }
```
iccub: Remove comment if not needed

Client/Frontend/Browser/Onboarding/OnboardingNavigationController.swift
```diff
@@ -127,3 +197,156 @@ extension OnboardingNavigationController {
...
+        let container = transitionContext.containerView
+        
+        guard let fromView = transitionContext.view(forKey: UITransitionContextViewKey.from) else { return }
+        
+        guard let toView = transitionContext.view(forKey: UITransitionContextViewKey.to) else { return }
+        
+        //Setup
+        fromView.frame = container.bounds
+        toView.frame = container.bounds
+        //isPresenting ? container.addSubview(toView) : container.insertSubview(toView, belowSubview: fromView)
```
iccub: Remove comment if not needed

iccub (changes requested):
Small review + CI is still failing

Client.xcodeproj/project.pbxproj
Brandon-T: Not sure why.. but if I ever rebase on development, it automatically does this. I have reverted it.

iccub (approved):

Client/Frontend/Browser/BrowserViewController.swift
kylehickinson: This matches Android, and we will be sticking with that for now

kylehickinson (approved):

Client/Frontend/Browser/Onboarding/OnboardingRewardsAgreementView.swift
```diff
@@ -0,0 +1,249 @@
...
+        
+        @objc
+        private func onTermsAccepted(_ button: UIButton) {
+            button.isSelected.toggle()
+            
+            agreeButton.backgroundColor = button.isSelected ? BraveUX.BraveOrange : BraveUX.BraveOrange.withAlphaComponent(0.7)
+            agreeButton.isEnabled = button.isSelected
+        }
+        
+        override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
```
kylehickinson, line 206: Why not just override `pointInside` instead of hit test:

```swift
override func point(inside point: CGPoint, with event: UIEvent?) -> Bool {
  if UIEdgeInsetsInsetRect(bounds, UX.checkboxInsets)).contains(point) {
    return true
  }
  return super.point(inside: point, with: event)
}
```

Edit: Just realized this is hitTest on the main view, not the checkbox, so point inside won't work

Brandon-T: In case we are wondering. I HAVE to add it on the main view and not the button itself because the stackView is legit blocking the input to the button on the top and bottom (it clips it).. but NOT the sides due to the label next to it lol.

kylehickinson: Yeah stack view's a bitch for not accepting touches outside its bounds :P

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -255,9 +255,10 @@ class BrowserViewController: UIViewController {
             DispatchQueue.main.asyncAfter(deadline: .now() + 3) {
                 if Preferences.Rewards.myFirstAdShown.value { return }
                 Preferences.Rewards.myFirstAdShown.value = true
-                AdsViewController.displayFirstAd(on: self) { [weak self] url in
-                    let request = URLRequest(url: url)
-                    self?.tabManager.addTabAndSelect(request, isPrivate: PrivateBrowsingManager.shared.isPrivateBrowsing)
```
kylehickinson, line 260: This change is on purpose, coming from #1597, looks like was an accident on rebase?

kylehickinson (changes requested):

Client/Frontend/Browser/BrowserViewController.swift
Brandon-T: It's on purpose. That other code is missing an argument for `displayFirstAd`

kylehickinson: Yeah sorry I mean the part where it uses `addTabAndSelect` instead of `openNewTab`

kylehickinson (approved):

karenkliu:
@Brandon-T @kylehickinson @anthonypkeane Caught up with Kyle about the "Join" button on the screen for Existing user BR on, Ads become available case. When I originally came up with the design, I thought that the user had to turn on ads when ads became available in their region. However, since ads are turned on by default when the user check BR ToS, we should change the button to "Show Me". That screen is purely to introduce users to ads; there is nothing to "turn on" or "join."

PR #1568, opened 2019-09-20
Fix #1554: Adds fallback logic for failed database migration.

### LINKED ISSUE #1554: Data loss on 1.12 upgrade for some users.

jhreis:
## Summary of Changes

This pull request fixes issue #1554

*Items remaining*
- Added migration flags
- Add migration re-try on subsequent version upgrades

## Submitter Checklist:

### New Users
1. Install 1.12.1 Brave
1. Save a bookmark 
1. Open Files app > Brave
1. See if there are any files stored here (there should not be)

### 1.12 Users, Good State -> standard upgrade testing
1. Install latest 1.12
1. Save bookmarks, open tabs
1. Install 1.12.1
1. Verify everything looks good

### 1.12 Users, Bad State
- Since we are not attempting to recover any data, we do not actually need to do anything special here, same as above.

### 1.11.4 Users, successful migration
- Since we can't repro a failed migration, this will be standard upgrade testing
- It would be best if testers had device plugged into Xcode logger in case they do run into any issues, then we can capture the error logs
1. Install 1.11.4 Brave
1. Save bookmarks, tabs
1. Upgrade to 1.12.1
1. Verify everything looks good
1. Go into Files app > Brave
1. See if there are any files stored here (there should not be)

### 1.11.4 Users, failed migration
- I will release a special beta version for this (will provide specific build number). This special build will force a migration failure, so this is how it should look / function.
1. Install 1.11.4 Brave
1. Save bookmarks, tabs
1. Upgrade to 1.12.1 <Migration Failure Version>
1. Verify everything looks good (since we fallback, there should be no apparent data lose)
1. Go into Files app > Brave
1. See if there are any files stored here (*there _should_ be ~3 "Brave" files in here this time*)

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Application/Migration.swift
+++ b/Client/Application/Migration.swift
@@ -27,15 +27,24 @@ class Migration {
     
     static func moveDatabaseToApplicationDirectory() {
         if Preferences.Database.DocumentToSupportDirectoryMigration.completed.value {
-            // Migration has been done in some regard, so drop out
+            // Migration has been done in some regard, so drop out.
             return
         }
+        
+        if Preferences.Database.DocumentToSupportDirectoryMigration.previousAttemptedVersion.value == AppInfo.appVersion {
+            // Migration has already been attempted for this version.
+            return
+        }
+        
         // Moves Coredata sqlite file from documents dir to application support dir.
         do {
             try DataController.shared.migrateToNewPathIfNeeded()
         } catch {
             log.error(error)
         }
+        
+        // Regardless of what happened, we attemtped a migration and document it:
+        Preferences.Database.DocumentToSupportDirectoryMigration.previousAttemptedVersion.value = AppInfo.appVersion
     }
     
     /// Adblock files don't have to be moved, they now have a new directory and will be downloaded there.
--- a/Client/Application/Migration.swift
+++ b/Client/Application/Migration.swift
@@ -26,7 +26,7 @@ class Migration {
     }
     
     static func moveDatabaseToApplicationDirectory() {
-        if Preferences.Database.documentToSupportDirectoryMigrationComplete.value {
+        if Preferences.Database.DocumentToSupportDirectoryMigration.completed.value {
             // Migration has been done in some regard, so drop out
             return
         }
--- a/Client/Application/Migration.swift
+++ b/Client/Application/Migration.swift
@@ -26,8 +26,16 @@ class Migration {
     }
     
     static func moveDatabaseToApplicationDirectory() {
+        if Preferences.Database.documentToSupportDirectoryMigrationComplete.value {
+            // Migration has been done in some regard, so drop out
+            return
+        }
         // Moves Coredata sqlite file from documents dir to application support dir.
-        DataController.shared.migrateToNewPathIfNeeded()
+        do {
+            try DataController.shared.migrateToNewPathIfNeeded()
+        } catch {
+            log.error(error)
+        }
     }
     
     /// Adblock files don't have to be moved, they now have a new directory and will be downloaded there.
--- a/Client/Application/Migration.swift
+++ b/Client/Application/Migration.swift
@@ -27,12 +27,7 @@ class Migration {
     
     static func moveDatabaseToApplicationDirectory() {
         // Moves Coredata sqlite file from documents dir to application support dir.
-        
-        do {
-            try DataController.shared.migrateToNewPathIfNeeded()
-        } catch {
-            log.error("Document -> Support database migration failed: \(error)")
-        }
+        DataController.shared.migrateToNewPathIfNeeded()
     }
     
     /// Adblock files don't have to be moved, they now have a new directory and will be downloaded there.
--- a/Client/Application/Migration.swift
+++ b/Client/Application/Migration.swift
@@ -26,8 +26,13 @@ class Migration {
     }
     
     static func moveDatabaseToApplicationDirectory() {
-        //Moves Coredata sqlite file from Documents dir to application support dir.
-        DataController.shared.migrateToNewPathIfNeeded()
+        // Moves Coredata sqlite file from documents dir to application support dir.
+        
+        do {
+            try DataController.shared.migrateToNewPathIfNeeded()
+        } catch {
+            log.error("Document -> Support database migration failed: \(error)")
+        }
     }
     
     /// Adblock files don't have to be moved, they now have a new directory and will be downloaded there.
--- a/Data/DataPreferences.swift
+++ b/Data/DataPreferences.swift
@@ -5,7 +5,7 @@
 import Foundation
 import BraveShared
 
-extension Preferences {
+public extension Preferences {
     final class Sync {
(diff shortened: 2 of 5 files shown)


Brandon-T (approved):

Client/Application/Migration.swift
```diff
@@ -26,8 +26,25 @@ class Migration {
     }
     
     static func moveDatabaseToApplicationDirectory() {
-        //Moves Coredata sqlite file from Documents dir to application support dir.
-        DataController.shared.migrateToNewPathIfNeeded()
+        if Preferences.Database.DocumentToSupportDirectoryMigration.completed.value {
```
danishjafri88, line 29: Can we merge the two if's and the code below. and keep a common comment for it.

danishjafri88 (approved):

Client/Application/Migration.swift
```diff
@@ -26,8 +26,25 @@ class Migration {
...
+            // Migration has been done in some regard, so drop out.
+            return
+        }
+        
+        if Preferences.Database.DocumentToSupportDirectoryMigration.previousAttemptedVersion.value == AppInfo.appVersion {
+            // Migration has already been attempted for this version.
```
iccub, line 35: We could convert some of these comments to debug logs

Data/DataPreferences.swift
```diff
@@ -22,4 +22,21 @@ extension Preferences {
         /// See Sync.syncSeed for more details.
         static let seedName = Option<Bool>(key: "sync.is-sync-seed-set", default: false)
     }
+    
+    final class Database {
+        
+        public final class DocumentToSupportDirectoryMigration {
+            /// This is the filepath used for the coredata database
```
iccub: Is this comment relevant? the preference is a bool, not a filepath string

Data/models/DataController.swift
```diff
@@ -30,22 +31,75 @@ public class DataController: NSObject {
...
+            throw MigrationError.MigrationFailed("Document -> Support database migration failed: \(error)")
+            // Migration failed somehow, and old store is present. Flag not being updated 😭
         }
+        
+        // Regardless of cleanup logic, the actual migration was successful, so we're just going for it 🙀😎
+        Preferences.Database.DocumentToSupportDirectoryMigration.completed.value = true
+        
+        // Cleanup time 🧹
+        do {
+            try coordinator.destroyPersistentStore(at: oldDocumentStoreURL, ofType: NSSQLiteStoreType, options: nil)
```
iccub, line 86: Don't we need `NSPersistentStoreFileProtectionKey` option here as well?

Data/models/DataController.swift
```diff
@@ -30,22 +31,75 @@ public class DataController: NSObject {
...
+        let coordinator = migrationContainer.persistentStoreCoordinator
+
+        guard let oldStore = coordinator.persistentStore(for: oldDocumentStoreURL) else {
+            throw MigrationError.OldStoreMissing("Old store unavailable")
+        }
+        
+        // Attempting actual database migration Document -> Support 🤞
+        do {
+            let migrationOptions = [
+                NSPersistentStoreFileProtectionKey: true
```
iccub, line 73: Is this the key that will most likely help with this migration bug in 1.12?
If so I would add a comment here

iccub (changes requested):

Data/models/DataController.swift
jhreis: Based on my testing, I'm not even entirely sure that's a valid key for `destroy`. We considered just excluding this entirely, as we're deleting the files right after, so could do that.

Client/Application/Migration.swift
jhreis: Yes, true, I'll place some logs in. We need to massively cleanup our logs in general too, lol.

Data/models/DataController.swift
jhreis: I don't think it is actually necessary. The modification to `addPersistentStore` should be enough. I just added this in case it was useful for adding file protection for the new store. @Brandon-T is submitting a new PR to handle some of the encryption concerns too at some point.

iccub (approved):

Client/Application/Migration.swift
jhreis: Going to leave separate to leave comment clarity. Otherwise comment will be something like:
// this 
// or
// something else

Which is okay, but given how sensitive this is, I think this is probably fine. Appreciate the feedback though.

Data/DataPreferences.swift
jhreis: Good catch, fixed. I originally had this just pointing to filepath, but switched to bool. Thanks 😄

kjozwiak:
Results for `iPhone 6s+` using `iOS 12.4.1`:

#### `New Users` - **PASSED** (x2) <-- Using `Production` TF

* installed `1.12.1 (19.9.20.20.23)` & saved a bookmark
* opened `Files.app` and ensured that there weren't any files being stored under `Brave`
* restarted `Brave` a few times and ensured that there weren't any files being created under `Brave`

#### `1.12 Users, Good State -> standard upgrade testing` - **PASSED** (x3) <-- Using `Production` TF

* installed `1.12 (19.9.12.18)` and saved several bookmarks and opened ~10 different tabs
* updated to `1.12.1 (19.9.20.20.23)` and ensured that all the bookmarks and tabs were still present after updating
* opened `Files.app` and ensured that there weren't any files being stored under `Brave`

#### `1.11.4 Users, successful migration` - **PASSED** (x3) <-- Using `Production` TF

* installed `1.11.4 (19.8.29.21)` and saved several bookmarks and opened ~10 different tabs
* updated to `1.12.1 (19.9.20.20.23)` and ensured that all the bookmarks and tabs were still present after updating
* opened `Files.app` and ensured that there weren't any files being stored under `Brave`

#### `1.11.4 Users, failed migration` - (x3) **Needs conformation from Developer** <-- Using `BETA` TF

* installed `1.11.4 (19.8.30.19)` and saved several bookmarks and opened ~10 different tabs
* updated to `1.12.1 (19.9.20.23.9)` (corrupted DB build)

As mentioned in Slack earlier, the following occurs:

* onboarding appears twice, once for `1.11.4 (19.8.30.19)` and once for `1.12.1 (19.9.20.23.9)` after upgrading
* all NTP stats are being cleared
* Setting are being cleared
* `profile.profile` folder is being left behind which includes `login` & `TabManagerScreenshot` files

Issue #1585, opened 2019-09-26
WebAuthn: no prompt during assertion when userPresence is explicitly to required or preferred


MaximeNdutiye:
### Description: 

Whenever `PublicKeyCredentialRequestOptions.userVerification` is explicitly set to `"preferred"` or `"required"` Brave behaves as though `userVerification="discouraged"`. 

The behaviour is out of spec and always results in the browser not prompting the user to touch their authenticator.

### Steps to Reproduce 

1. [perform a create request](https://eparkko-lab.github.io/webauthn-playground/?requestType=create&webauthnRequest=eyJjaGFsbGVuZ2UiOiJhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhIiwicnAiOnsiaWQiOiJlcGFya2tvLWxhYi5naXRodWIuaW8iLCJuYW1lIjoiZXBhcmtrby1sYWIuZ2l0aHViLmlvIn0sImF1dGhlbnRpY2F0b3JTZWxlY3Rpb24iOnsidXNlclZlcmlmaWNhdGlvbiI6InByZWZlcnJlZCJ9LCJ0aW1lb3V0IjoxNTAwMCwicHViS2V5Q3JlZFBhcmFtcyI6W3sidHlwZSI6InB1YmxpYy1rZXkiLCJhbGciOi03fV0sInVzZXIiOnsiaWQiOiJiYmJiYmJiIiwibmFtZSI6IkJvYmJ5QGxvY2FsaG9zdCIsImRpc3BsYXlOYW1lIjoiYmJiYmJiYiJ9fQ%3D%3D&submitRequest=true)

[...]

PR #1608, opened 2019-10-07
Fixes YubiKey presentation for "required" and "preferred"

Brandon-T:
## Summary of Changes

This pull request fixes issue https://github.com/brave/brave-ios/issues/1585

The boolean was flipped the wrong direction. This could have possibly been an enum instead but it doesn't matter as we only care if it's set to true or not.

## Submitter Checklist:

## Test Plan:

see https://github.com/brave/brave-ios/issues/1585

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/WebAuthN/WebAuthnAuthenticateRequest.swift
+++ b/Client/WebAuthN/WebAuthnAuthenticateRequest.swift
@@ -37,7 +37,7 @@ extension WebAuthnAuthenticateRequest: Decodable {
         
         // userPresence is the inverse of userVerification, UP by default is true
         let userVerifcationString = try publicKeyDictionary.decodeIfPresent(String.self, forKey: .userVerification) ?? "discouraged"
-        userPresence = userVerifcationString == "discouraged"
+        userPresence = userVerifcationString != "discouraged"
         
         let allowCredentialsArray = try publicKeyDictionary.decode([AllowCredentials].self, forKey: .allowCredentials)
     
--- a/Client/WebAuthN/WebAuthnAuthenticateRequest.swift
+++ b/Client/WebAuthN/WebAuthnAuthenticateRequest.swift
@@ -36,7 +36,7 @@ extension WebAuthnAuthenticateRequest: Decodable {
         challenge = try publicKeyDictionary.decode(String.self, forKey: .challenge)
         
         // userPresence is the inverse of userVerification, UP by default is true
-        let userVerifcationString = try publicKeyDictionary.decodeIfPresent(String.self, forKey: .userVerification) ?? "discouraged"
+        let userVerifcationString = try publicKeyDictionary.decodeIfPresent(String.self, forKey: .userVerification) ?? "required"
         userPresence = userVerifcationString != "discouraged"
         
         let allowCredentialsArray = try publicKeyDictionary.decode([AllowCredentials].self, forKey: .allowCredentials)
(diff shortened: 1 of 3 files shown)


iccub (changes requested):
Tests need to be fixed

iccub (approved):
Approving and adding blocked label until @jumde reviews it.

Client/WebAuthN/WebAuthnAuthenticateRequest.swift
```diff
@@ -36,8 +36,8 @@ extension WebAuthnAuthenticateRequest: Decodable {
         challenge = try publicKeyDictionary.decode(String.self, forKey: .challenge)
         
         // userPresence is the inverse of userVerification, UP by default is true
-        let userVerifcationString = try publicKeyDictionary.decodeIfPresent(String.self, forKey: .userVerification) ?? "discouraged"
-        userPresence = userVerifcationString == "discouraged"
+        let userVerifcationString = try publicKeyDictionary.decodeIfPresent(String.self, forKey: .userVerification) ?? "required"
```
jumde: The default UV value if not present is being updated to `discouraged`. See here: https://github.com/w3c/webauthn/pull/1259/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR2437

Brandon-T: https://chromium.googlesource.com/chromium/src/+/master/content/browser/webauth/uv_preferred.md

It is still "preferred".. Google Chrome is now throwing a warning in the console for Chrome 77 when a value is not explicitly provided.

Also UP is not always the inverse of UV.. https://github.com/w3c/webauthn/issues/1306 and https://github.com/w3c/webauthn/issues/1123

So now.. I have to go re-read everything.. This was opened just 18 days ago. Or does this not apply to us?

jumde: Looks like the change from `preferred` to `discouraged` was discarded. Good to know. 

```
We have changed Chrome so that, with Chrome 77, any WebAuthn call that doesn't set an explicit value for userVerification will trigger a warning in the Javascript console pointing to this page. This can be silenced by specifying any explicit value. We hope this to point RPs towards considering whether preferred is their desired behaviour.
```

Given the change that was merged here: https://github.com/w3c/webauthn/pull/1140/files - we should set UP to true always.

Client/WebAuthN/WebAuthnAuthenticateRequest.swift
```diff
@@ -35,9 +42,12 @@ extension WebAuthnAuthenticateRequest: Decodable {
...
         
-        // userPresence is the inverse of userVerification, UP by default is true
-        let userVerifcationString = try publicKeyDictionary.decodeIfPresent(String.self, forKey: .userVerification) ?? "discouraged"
-        userPresence = userVerifcationString == "discouraged"
+        // As of the latest spec changes, this valid will always be true!
+        // https://github.com/w3c/webauthn/pull/1140/files
+        userPresence = true
+        
+        // As a result of the above, we need to ensure that the default value is preferred
+        userVerification = try publicKeyDictionary.decodeIfPresent(WebAuthnUserVerification.self, forKey: .userVerification) ?? .preferred
```
jumde, line 50: Are we using this anywhere?

Brandon-T: No it's not used atm but I left it there in case in the future something changes. I does come back as part of the WebAuthnRequest in the response json so I parsed it anyway even though it is currently not used anywhere except in unit tests to make sure we have a valid default value.

jumde (approved):
lgtm

iccub:
Tests failing

PR #1625, opened 2019-10-10
Ref #1499: Add link preview to long press context menu.

iccub:
## Submitter Checklist:

## Screenshots:

Zrzut ekranu 2019-10-10 o 18 56 27

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -2505,17 +2505,29 @@ extension BrowserViewController: WKUIDelegate {
                     
                     actions.append(shareAction)
                 }
+                
+                let linkPreview = Preferences.General.enableLinkPreview.value
+                
+                let linkPreviewTitle = linkPreview ?
+                    Strings.HideLinkPreviewsActionTitle : Strings.ShowLinkPreviewsActionTitle
+                let linkPreviewAction = UIAction(title: linkPreviewTitle, image: UIImage(systemName: "eye.fill")) { _ in
+                    Preferences.General.enableLinkPreview.value.toggle()
+                }
             
+                actions.append(linkPreviewAction)
             }
             
-            return UIMenu(title: url.absoluteString, children: actions)
+            let menuTitle = Preferences.General.enableLinkPreview.value ? "" : url.absoluteString
+            return UIMenu(title: menuTitle, children: actions)
         }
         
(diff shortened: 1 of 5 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2511,11 +2511,21 @@ extension BrowserViewController: WKUIDelegate {
             return UIMenu(title: url.absoluteString, children: actions)
         }
         
-        let config = UIContextMenuConfiguration(identifier: nil, previewProvider: nil, actionProvider: actionProvider)
+        let linkPreview: UIContextMenuContentPreviewProvider = {
+            return SFSafariViewController(url: url)
```
kylehickinson: Is it possible to make our own preview provider that uses BVC in some way (so shields are loaded in and such)

kylehickinson (changes requested):
Also, Safari provides a way to disable these previews, we may want to give the same option to users

kylehickinson:
Sorry that review meant to be `comment`, not `request changes`

Client/Frontend/Browser/BrowserViewController.swift
iccub: UIContextMenuContentPreviewProvider accepts any `UIViewController` so we could try to add some custom view with content blocker lists

iccub:
I think I should add a preference to hide preview links, similar to what Safari does at the moment

iccub:
Added content blocking to the link preview and an option to hide link previews

Zrzut ekranu 2019-10-11 o 14 53 41
Zrzut ekranu 2019-10-11 o 14 53 29

kylehickinson (changes requested):
Somehow after disabling link previews, I no longer got the enable link previews option… It seems like something is supposed to be there (you can see the menu has a final separator down there but no text/image.

IMG_C03DD67E96AA-1

kylehickinson:
Ahhh, never mind, that list scrolls, and wasn't visible because of the super long URL… Not sure if we should truncate it?

iccub:
Added a 100char truncate to it, Safari truncates it too btw

Zrzut ekranu 2019-10-11 o 15 59 53

kylehickinson (approved):
Works for me

Brandon-T (approved):

PR #1628, opened 2019-10-11
Ref #1533 - Set location view text field to scheme-less domain url

chickdan:
## Summary of Changes

This pull request fixes issue #1533 

This change sets the `urlTextField` using the `domainURL`. Setting it here allows for the full url to still be accessible when the user taps into the location view.

## Submitter Checklist:

## Screenshots:

Shortened domainURL
Simulator Screen Shot - iPhone Xs - 2019-10-10 at 23 49 05

Full url when location view tapped
Simulator Screen Shot - iPhone Xs - 2019-10-10 at 23 49 12

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


iccub (dismissed):
Thanks for the contribution, the code works well :) 
I'm adding `blocked` label, waiting on final decision from the product team

Client/Frontend/Browser/Toolbars/UrlBar/TabLocationView.swift
```diff
@@ -321,7 +321,7 @@ class TabLocationView: UIView {
 
     fileprivate func updateTextWithURL() {
         (urlTextField as? DisplayTextField)?.hostString = url?.host ?? ""
-        urlTextField.text = url?.schemelessAbsoluteString.trim("/")
+        urlTextField.text = url?.domainURL.schemelessAbsoluteString.trim("/")
```
iccub: This `domainURL` strips `m.` and `mobile.` schemes as well.
I talked to security team and they want to hide scheme only with `www.` case.

So one additional helper method needs to be added to `URLExtensions.swift` to remove only the `www` part

iccub (changes requested):

chickdan:
@iccub I've pushed an update for the requested change. Please let me know if you have any suggestions for improvement.

Shared/Extensions/URLExtensions.swift
```diff
@@ -237,6 +237,39 @@ extension URL {
         return publicSuffixFromHost(host, withAdditionalParts: 1)
     }
 
+    /**
+    * Returns a possible mobile host and the domain, but with the same scheme, and a trailing '/'.
+    *
+    * E.g., https://m.foo.com/bar/baz?noo=abc#123  => https://m.foo.com/
+    * Or https://mobile.foo.com/bar/baz?noo=abc#123 => https://mobile.foo.com/
+    */
+    public var mobileDomainURL: URL {
```
iccub: All of this code is duplicated and the only difference is between `"^(www)\\."`, `"^(www|mobile|m)\\."` regexes
What do you think about some DRY refactor?

- change `domainURL` to a function `func domainURL(stripWWWSubdomainOnly: Bool = false) -> URL`
- same for `normalizedHost`
- in `normalizedHost` add something like `let regex = stripWWWSubdomainOnly ? "^(www)\\." : "^(www|mobile|m)\\."`
- update documentation comment in `domainURL`
- update `testdomainURL` unit test

iccub (changes requested):

Shared/Extensions/URLExtensions.swift
chickdan: Great suggestions! I've made the requested changes and added a new unit test to account for the `stripWWWSubdomainOnly` parameter.

iccub (approved):
Great stuff, thanks :)

iccub:
Just needs a rebase and we are good

chickdan:
conflicts resolved :)

iccub:
CI still failing with 
```
/Users/travis/build/brave/brave-ios/BraveShared/Analytics/UrpService.swift:28:44: cannot call value of non-function type 'URL?'
        guard let hostUrl = try? host.asURL(), let normalizedHost = hostUrl.normalizedHost() else { return nil }
```

Issue #1655, opened 2019-04-15
Implement DeviceCheck


jhreis:
@anthonypkeane commented on [Fri Apr 12 2019](https://github.com/brave/brave-ios/issues/1052)

### Description: 
Implement DeviceCheck for upcoming Brave Rewards on iOS
https://developer.apple.com/documentation/devicecheck

nvonpentz:
Server changes are in review here https://github.com/brave-intl/bat-captcha/pull/45, but the client side should not need to wait for the server to start implementing @tmancey

PR #1777, opened 2019-10-23
Fix: #1655 - Added working Device Check API that is compliant with the DeviceCheck Server

Brandon-T:
- Security Review: https://github.com/brave/security/issues/13
- Spec: https://github.com/brave-intl/bat-captcha/pull/45
- Added working Device Check API that is compliant with the DeviceCheck Server
- Added PEM formatting (ASN.1 DER - ANSI X9.62) for Elliptic Curve Keys following the RFC spec.
- Removed biometric requirements from the Keys.
- Added signing and verifying of challenge nonce with the key as per Brave spec.
- Added example flow for Device Check from Brave spec.
- Made URLs compliant with Brave server instead of RFC spec.

**Hello World:**
```swift
//Sample Flow:
//generate Token -> generateEnrollment -> registerDevice
//getAttestation -> setAttestation (verify it)
// Verify if we failed or not.. and do something with the result..

class DeviceCheckExample {
  private let client = DeviceCheckClient()
  private var token = ""
  private let paymentId = UUID().uuidString //This comes from Brave Rewards..
  private var completion: ((_ error: Error?) -> Void)?
  
  // Starts the device check flow with the server
  func startHandshake(_ completion: @escaping (Error?) -> Void) {
    self.completion = completion
    self.generateToken()
  }

  deinit { self.completion = nil }
  
  // Generates a device check token using the iOS Apple APIs.
  // I do not store this token anywhere atm.
  private func generateToken() {
    client.generateToken { [weak self] token, error in
      
      if let error = error {
        self.completion?(error)
        return
      }
      
      self.token = token
      self.generateEnrollment()
    }
  }
  
  // Generates an enrollment structure using the token and making a call to the server.
  private func generateEnrollment() {
    if token.isEmpty {
      self.completion?("Invalid Device Check Token")
      return
    }
    
    client.generateEnrollment(paymentId: paymentId, token: token) { [weak self] registration, error in
      
      if let error = error {
        self.completion?(error)
        return
      }
      
      self.registerDevice(registration)
    }
  }
  
  // Registers the device with the server
  private func registerDevice(_ registration: DeviceCheckRegistration?) {
    guard let registration = registration else {
      self.completion?("Error registering device - Unknown Reason")
      return
    }
    
    client.registerDevice(enrollment: registration) { [weak self] error in
      
      if let error = error {
        self.completion?(error)
        return
      }
      
      self.getAttestation()
    }
  }
  
  // Retrieves a challenge-nonce from the server
  private func getAttestation() {
    client.getAttestation(paymentId: paymentId) { [weak self] attestation, error in
      
      if let error = error {
        self.completion?(error)
        return
      }
      
      self.setAttestation(attestation)
    }
  }
  
  // Responds to the challenge and sends it to the server.
  private func setAttestation(_ attestation: AttestationBlob?) {
    guard let attestation = attestation else {
      self.completion?("Error retrieving attestation - Unknown Reason")
      return
    }
    
    client.setAttestation(nonce: attestation.nonce) { [weak self] error in
      
      if let error = error {
        self.completion?(error)
        return
      }
      
      self.completion?(nil)
    }
  }
}
```

**Hello World Usage:**
```swift
DeviceCheckExample().startHandshake { error in
  if let error = error {
    log.error(error) //Failed to retrieve attestation for this device.. Do something..
    return
  }
  
  //Attestation has been successful so do something else now.. claim the grant I guess?..
}
```

## Summary of Changes

This pull request fixes issue https://github.com/brave/brave-ios/issues/1655

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Client/DeviceCheck/Cryptography.swift
```diff
@@ -0,0 +1,240 @@
...
+      kSecAttrApplicationTag: id.data(using: .utf8)!,
+      kSecMatchLimit: kSecMatchLimitOne,
+      kSecReturnRef: kCFBooleanFalse as Any,
+      kSecReturnAttributes: kCFBooleanTrue as Any
+    ]
+    
+    var result: CFTypeRef?
+    let error = SecItemCopyMatching(query as CFDictionary, &result)
+    if error == errSecSuccess || error == errSecInteractionNotAllowed {
+      if let result = result as? [String: Any], let accessControl = result[kSecAttrAccessControl as String] as! SecAccessControl? { //swiftlint:disable:this force_cast
```
iccub: Do you think it's better to crash the app here instead of unwrapping this and return false?

Client/DeviceCheck/Cryptography.swift
```diff
@@ -0,0 +1,240 @@
...
+  private let key: SecKey
+  private let keyId: String
+  
+  public init(key: SecKey, keyId: String) {
+    self.key = key
+    self.keyId = keyId
+  }
+  
+  /// Returns the private key
+  public func getPrivateKey() -> SecKey {
```
jumde: Do you expect the private key to be used outside this file? If not we can make it private and only expose the utility functions like `sign` and `getPublicKey`.

Brandon-T: True. I will fix it. The private key is not needed outside this file but also, the private key can't be exported as well so it doesn't hurt :)

But yeah, good point. It will be fixed.

Client/DeviceCheck/DeviceCheck.swift
```diff
@@ -0,0 +1,367 @@
...
+      completion(registration, nil)
+    } catch {
+      completion(nil, error)
+    }
+  }
+}
+
+private extension DeviceCheckClient {
+  // The base URL of the server
+  private static let baseURL = URL(string: "http://reputation-alb-1044480833.us-west-2.elb.amazonaws.com")
```
jumde: https

Brandon-T: I got the url from slack:

image

If I use `https`, I get an error atm. I'm assuming it's because it's dev which is why I never mentioned it and I'm thinking production and staging will be different than dev.

Client/DeviceCheck/Cryptography.swift
```diff
@@ -0,0 +1,240 @@
...
+  //The error domain
+  public let domain: String
+  
+  //The error code
+  public let code: Int32
+  
+  //A description of the error
+  public let description: String?
+  
+  init(code: Int32, description: String? = nil) {
```
iccub: All `CryptographyError` return code = -1. I think we could hardcode it, or at least add as a default init parameter

Client/DeviceCheck/Cryptography.swift
```diff
@@ -0,0 +1,240 @@
...
+    
+    return CryptographyError(code: error)
+  }
+  
+  /// Signs a "message" with the key and returns the signature
+  public func sign(message: String) throws -> Data {
+    var error: Unmanaged<CFError>?
+    let signature = SecKeyCreateSignature(key,
+                                          .ecdsaSignatureMessageX962SHA256,
+                                          message.data(using: .utf8)! as CFData,
```
iccub: Unwrap and throw error if can't be unwrapped

Client/DeviceCheck/Cryptography.swift
```diff
@@ -0,0 +1,240 @@
...
+
+/// A class used for generating cryptographic keys
+public class Cryptography {
+  
+  /// The access control flags for any keys generated
+  public static let accessControlFlags = SecAccessControlCreateWithFlags(kCFAllocatorDefault, kSecAttrAccessibleWhenUnlockedThisDeviceOnly, [.privateKeyUsage], nil) //.biometryAny
+  
+  /// Determines if a key exists in the keychain without retrieving it
+  public class func keyExists(id: String) -> Bool {
+    let query: [CFString: Any] = [
```
iccub: This query is the same as in `isKeyRequiringBiometrics` perhaps we could extract it?

Client/DeviceCheck/Cryptography.swift
```diff
@@ -0,0 +1,240 @@
...
+      kSecClass: kSecClassKey,
+      kSecAttrApplicationTag: id.data(using: .utf8)!,
```
iccub: Remove force cast

Client/DeviceCheck/Cryptography.swift
```diff
@@ -0,0 +1,240 @@
...
+      kSecMatchLimit: kSecMatchLimitOne,
+      kSecReturnRef: kCFBooleanFalse as Any,
+      kSecReturnAttributes: kCFBooleanTrue as Any
+    ]
+    
+    var result: CFTypeRef?
+    let error = SecItemCopyMatching(query as CFDictionary, &result)
+    return error == errSecSuccess || error == errSecInteractionNotAllowed
+      && result != nil && CFDictionaryGetCount((result! as! CFDictionary)) > 0 //swiftlint:disable:this force_cast
```
iccub: Remove force cast if it's better to fail safely

Client/DeviceCheck/Cryptography.swift
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    return false
+  }
+  
+  /// Retrieves an existing key from the secure-enclave
+  public class func getExistingKey(id: String) throws -> CryptographicKey? {
+    let query: [CFString: Any] = [
+      kSecClass: kSecClassKey,
+      kSecAttrApplicationTag: id.data(using: .utf8)!,
```
iccub: Remove force unwrapping

Client/DeviceCheck/Cryptography.swift
```diff
@@ -0,0 +1,240 @@
...
+      kSecMatchLimit: kSecMatchLimitOne,
+      kSecReturnRef: kCFBooleanTrue as Any
+    ]
+    
+    var result: CFTypeRef?
+    let error = SecItemCopyMatching(query as CFDictionary, &result)
+    if error == errSecSuccess || error == errSecDuplicateItem || error == errSecInteractionNotAllowed {
+      if let res = result {
+        return CryptographicKey(key: res as! SecKey, keyId: id) //swiftlint:disable:this force_cast
```
iccub: remove force cast

iccub (changes requested):
Mostly some concerns about force casting and force unwrapping, I don't think we should use so many force casts in this code.
If there is a reason for it please let me know

Client/DeviceCheck/Cryptography.swift
Brandon-T: image

Screen Shot 2019-10-24 at 11 51 58 AM

This is a bug in Xcode and it can't be done any other way. There's no way to unwrap a `CFTypeRef?` as another `CF` type. You will always end up with a force somewhere :(

I've tried to `if let`, I have tried `guard let`, I have tried optional cast and ternary unwrap.. None of them work and will give the above errors. That's why I have to force cast it to an optional type and then unwrap that optional type :(

iccub: Lol XCode really doesn't like you :D
Would be good to put a comment about it somewhere in code

Brandon-T: I fixed it by first checking what type was returned via its TypeId and then force-casting. This should provide the safety we're looking for.

Client/DeviceCheck/DeviceCheck.swift
```diff
@@ -0,0 +1,371 @@
...
+    } catch {
+      completion(error)
+    }
+  }
+  
+  // Retrieves existing attestations for this device and returns a nonce if any
+  public func getAttestation(paymentId: String, _ completion: @escaping (AttestationBlob?, Error?) -> Void) {
+    do {
+      guard let privateKey = try Cryptography.getExistingKey(id: DeviceCheckClient.privateKeyId) else {
+        throw CryptographyError(code: -1, description: "Unable to retrieve existing private key")
```
iccub: Remove `code`

Client/DeviceCheck/DeviceCheck.swift
```diff
@@ -0,0 +1,371 @@
...
+      }
+      
+      guard let publicKey = try privateKey.getPublicKeySha256FingerPrint() else {
+        throw CryptographyError(code: -1, description: "Unable to retrieve public key")
```
iccub: Remove `code`

Client/DeviceCheck/DeviceCheck.swift
```diff
@@ -0,0 +1,371 @@
...
+    } catch {
+      completion(nil, error)
+    }
+  }
+  
+  // Sends the attestation to the server along with the nonce and the challenge signature
+  public func setAttestation(nonce: String, _ completion: @escaping (Error?) -> Void) {
+    do {
+      guard let privateKey = try Cryptography.getExistingKey(id: DeviceCheckClient.privateKeyId) else {
+        throw CryptographyError(code: -1, description: "Unable to retrieve existing private key")
```
iccub: Remove `code`

Client/DeviceCheck/DeviceCheck.swift
```diff
@@ -0,0 +1,371 @@
...
+          
+          completion(deviceCheckToken, nil)
+    }
+  }
+  
+  // Generates an enrollment structure to be used with `registerDevice`
+  public func generateEnrollment(paymentId: String, token: String, _ completion: (DeviceCheckRegistration?, Error?) -> Void) {
+    do {
+      guard let privateKey = try Cryptography.generateKey(id: DeviceCheckClient.privateKeyId) else {
+        throw CryptographyError(code: -1, description: "Unable to generate private key")
```
iccub: Remove `code`

Client/DeviceCheck/DeviceCheck.swift
```diff
@@ -0,0 +1,371 @@
...
+      }
+      
+      guard let publicKey = try privateKey.getPublicAsPEM() else {
+        throw CryptographyError(code: -1, description: "Unable to retrieve public key")
```
iccub: Remove `code`

Client/DeviceCheck/DeviceCheck.swift
```diff
@@ -0,0 +1,371 @@
...
+    let query = parameters.map({
+      "\($0.key)=\($0.value.addingPercentEncoding(withAllowedCharacters: CharacterSet(charactersIn: "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~")) ?? $0.value)"
+    }).joined(separator: "&")
+    
+    var urlComponents = URLComponents()
+    urlComponents.scheme = url.scheme
+    urlComponents.host = url.host
+    urlComponents.path = url.path
+    urlComponents.percentEncodedQuery = query
+    //urlComponents.queryItems = parameters.map({
```
iccub: Is this code still needed?

Client/DeviceCheck/DeviceCheck.swift
```diff
@@ -0,0 +1,371 @@
...
+  }
+  
+  enum CodingKeys: String, CodingKey {
+    case enrollmentBlob
+    case signature
+  }
+}
+
+public struct DeviceCheckEnrollment: Codable {
+  // Not sure yet
```
iccub: Still not sure or can this comment be updated?

Client/DeviceCheck/Cryptography.swift
```diff
@@ -0,0 +1,279 @@
...
+  
+  /// Determines if a key requires biometrics to access
+  public class func isKeyRequiringBiometrics(id: String) -> Bool {
+    let properties = getKeyProperties(id: id)
+    if properties.status == errSecSuccess || properties.status == errSecInteractionNotAllowed {
+      if let result = properties.result as? [String: Any],
+        let item = result[kSecAttrAccessControl as String] as CFTypeRef?,
+        CFGetTypeID(item) == SecAccessControlGetTypeID() {
+        
+        //cbio, pbio (.currentBioSet, .presentBioSet)
```
iccub: remove if not needed

iccub (changes requested):
Looks good in general, need some cleanup - removing commented out code

iccub (approved):
Goood, do we wait with this PR until server side is complete?

Brandon-T:
> Goood, do we wait with this PR until server side is complete?

The server side works with this PR right now. We're currently waiting on grants to be completed but I'm working on integrating this into the Wallet creation atm (in a local branch) & waiting on Security team to give the ok for this PR. Once it's merged, we can completely integrate it into wallet and make a PR for that. Then we wait again for the grant stuff to be complete & continue integration.

jseidl (approved):
Reviewed the code for common general programming issues but @riastradh-brave is a better suited person for attesting the cryptography-related code.

All looks good for me. Good job folks!

jumde (approved):
lgtm

PR #1790, opened 2019-10-24
Fix 1714: overrides for settings view

quangDecember:
## Summary of Changes

This pull request fixes issue #1714 

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Settings/SettingsNavigationController.swift
+++ b/Client/Frontend/Settings/SettingsNavigationController.swift
@@ -16,6 +16,13 @@ class SettingsNavigationController: UINavigationController {
     }
 
     override var preferredStatusBarStyle: UIStatusBarStyle {
+        if #available(iOS 13, *) {
+            if self.view.appearanceOverrideUserInterfaceStyle == .light || self.overrideUserInterfaceStyle == .light {
+                return .darkContent
+            } else {
+                return .lightContent
+            }
+        }
         return .default
     }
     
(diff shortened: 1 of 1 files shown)


quangDecember:
Hi I made a fix for #1714 
the problem was Settings page status bar seems to represent the OS, not the current app appearance, so I overrides it again

Client/Frontend/Settings/SettingsNavigationController.swift
```diff
@@ -16,6 +16,13 @@ class SettingsNavigationController: UINavigationController {
     }
 
     override var preferredStatusBarStyle: UIStatusBarStyle {
+        if #available(iOS 13, *) {
+            if self.view.appearanceOverrideUserInterfaceStyle == .light || self.overrideUserInterfaceStyle == .light {
+                return .darkContent
+            } else {
```
Brandon-T: This is so minor imo, but it's worth mentioning just for future reference.

```swift
if condition {
    return a
} else {
    return b
}

//can also be written as (without the else):

if condition {
    return a
}
return b
```

Anyway, thanks for the fix! :)
Really appreciated.

Brandon-T (approved):

iccub (changes requested):
Please address the change that Brandon requested and we can merge this PR :)
Looks good

iccub (approved):
Thank you sir!

PR #1827, opened 2019-10-30
Fix #409: Always save Domain on background context.

### LINKED ISSUE #409: Refactor `getOrCreateForUrl` for Background Writes

iccub:
## Summary of Changes

This pull request fixes issue #409 

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Data/models/Domain.swift
+++ b/Data/models/Domain.swift
@@ -4,6 +4,9 @@ import UIKit
 import CoreData
 import Foundation
 import BraveShared
+import Shared
+
+private let log = Logger.browserLogger
 
 public final class Domain: NSManagedObject, CRUD {
     
@@ -31,7 +34,7 @@ public final class Domain: NSManagedObject, CRUD {
     
     public class func getOrCreate(forUrl url: URL, persistent: Bool) -> Domain {
         let context = persistent ? DataController.viewContext : DataController.viewContextInMemory
-        return getOrCreateInternal(url, context: context, save: true)
+        return getOrCreateInternal(url, context: context, save: true, persistent: persistent)
     }
     
     // MARK: Shields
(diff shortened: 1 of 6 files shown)


kylehickinson (approved):
Let me know if you still want @jhreis to take a look before merging

Data/models/Domain.swift
```diff
@@ -105,26 +108,39 @@ extension Domain {
     
     class func getOrCreateInternal(_ url: URL,
                                    context: NSManagedObjectContext = DataController.viewContext,
-                                   save: Bool = true) -> Domain {
+                                   save: Bool = true,
+                                   persistent: Bool = true) -> Domain {
```
jhreis: Can you document function to include the implications of the `save` and `persistent` flags. Probably necessary, but the combination of these two flags feels kind of weird. 

Like whenever `save=false` then `persistent=true`. Where whenever a persistence flag is explicit, we will generally want to also save.

I am thinking we should either:
- create an enum that outlines our 3 situations (save=no persistence=yes / save=yes persistence=yes / save=yes persistence=no)
   I think values would be something like: `standard`, `memory`, `delay`
   `standard` = both save and persistent set to true
   `memory` = save but no persistence
   `delay` (referring to batch insertions, or wherever we don't save immediately) = persistent but no explicit saving

- adjust names to be a bit more clear: save→batchingInserts (or something to indicate that lots of inserts are taking place and a save will happen at the end) persistent→inMemory (reverse the logic, this aligns more closely with our background context's IMO).

-----

I think regardless, it is probably time to make these both explicitly required. The implications of calling this function are getting increasingly unknown, and not setting defaults requires the caller to actually think about how a domain object should be treated.

Data/models/Domain.swift
```diff
@@ -105,26 +108,39 @@ extension Domain {
...
-        //  A solution to consider is creating a new background context here, creating, saving, and then re-fetching
-        //   that object in the requested context (regardless if it is `viewContext` or not)
+
         var newDomain: Domain!
-        context.performAndWait {
-            newDomain = Domain(entity: Domain.entity(context), insertInto: context)
+        
+        // Background context to use if main thread context was passed.
+        // Domains in private mode are saved to in-memory store.
+        let bgContext = persistent ? DataController.newBackgroundContext() : DataController.newBackgroundContextInMemory()
```
jhreis: Double ternary here feels slightly weird, esp since this generates a background context that may never even be used. 

Not a big deal, but curious what you think about one of these other approaches:

```
var writeContext = context
if context.concurrencyType == .mainQueueConcurrencyType {
    writeContext = persistent 
        ? DataController.newBackgroundContext() 
        : DataController.newBackgroundContextInMemory()
}
```

or setup an anonymous closure and execute it immediately:

```
let writeContext: NSManagedObjectContext = {
    if context.concurrencyType == .mainQueueConcurrencyType {
        // Main thread, creating new write context to use
        return persistent 
            ? DataController.newBackgroundContext()
            : DataController.newBackgroundContextInMemory()
    }
    return context
}()
```

jhreis (approved):

jhreis:
Love the improvements. This PR _feels_ so small and looks so tightly. Really solid job @iccub!

PR #1847, opened 2019-10-31
Fix #1846: Now support new beta Testflight releases.

### LINKED ISSUE #1846: Setup TestFlight Beta Channel
### Description: 
Currently Brave iOS has a beta channel that does not support external testing. In order to setup an actual beta channel, the existing beta needs to be re-defined to `dev` and the new beta setup correctly.

This requires
- [x] new provisioning profiles
- [x] new Testflight group
- [x] Update `xcconfig` files to contain correct information
    This will lead to the current `beta` information being moved to `dev` `xcconfig`, and the current `beta` `xcconfig` file being updated with new information.
- [x] Update `isRelease` `AppBuildChannel` to be called something more like `isPublicFacing` and add `.beta` to this value
- [x] Verify app behavior aligns with "release" behavior (e.g. debug settings panel)
- [x] Uploading?
- [x] Setup new `internal` fastlane script to replace the new beta (current `beta`)

jhreis:
## Summary of Changes

This pull request fixes issue #1846 

Couple things to note:
- The exact naming of these channels is quite confusing right now. Enterprise vs Dev vs Internal, etc... This will be addressed at some point, but requires some larger changes that I wanted to avoid making right now.
- The Bundle Ids are pretty similar, this is mostly due to 'old beta' (now dev/enterprise/internal) having the fixed / uneditable build id of BrowserBeta. At some point we may destroy the Testflight app and start fresh.

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Client/Configuration/FirefoxBeta.xcconfig
```diff
@@ -2,14 +2,16 @@
...
+
 #include "Base.xcconfig"
 
 MOZ_BUNDLE_DISPLAY_NAME = BraveBeta
 
 BRAVE_API_KEY = key
 
 // Bundle Identifier
-MOZ_BUNDLE_ID = $(BASE_BUNDLE_ID).BrowserBeta
+MOZ_BUNDLE_ID = $(BASE_BUNDLE_ID).browser.beta
```
iccub, line 14: What about a better name like `$(BASE_BUNDLE_ID).ExternalBeta`?
Edit: I would leave it for now, and change after 1.13

iccub (changes requested):
1. DAU tests are failing, should we use `stable` channel for external beta pings?
2. Should we enable Rewards logs for external beta? If not we have to change rewardsLogger in `Logger.swift` line 25

iccub:
I fixed DAU tests but we still need to decide about the channels

iccub (approved):
Looks good, verified on my end, binaries for all 3 channels are working

PR #1854, opened 2019-10-31
Fix: #1840 - Pub favicon is only shown for the first entry on tips details.

Brandon-T:
## Summary of Changes

This pull request fixes issue https://github.com/brave/brave-ios/issues/1840
- Issue happens because the code originally only checked for the first cell that matches the tip identifier. - However, user/qa would tip the same publisher twice so we can now have duplicate matching identifiers instead of limiting it to just one.
- Solution is to update ALL cells where the identifier matches.
- Tested and works.

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


kylehickinson (approved):

kylehickinson (changes requested):
Just realized, can you duplicate these fixes in the AutoContributeDetails controller as well?

Brandon-T:
> Just realized, can you duplicate these fixes in the AutoContributeDetails controller as well?

@kylehickinson AutoContributeDetails doesn't seem to have this problem.

It updates the image in `cellForRow` which means it'll update for every cell perfectly fine.

kylehickinson (approved):

PR #1931, opened 2019-11-11
Fix: #1916, Fix: #1904 - Makes notifications wider on iPad.

Brandon-T:
## Summary of Changes

This pull request fixes issue #1916 & #1904
- Makes notifications wider on iPad and presents them on top of everything else just like system notifications by rendering them on the window instead of the controller's view.
- This ticket includes two fixes because the only way for me to make it wider was to put it on the window which inherently fixes the notification stacking below the panel bug.

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/BraveRewardsUI/Ads/AdsViewController.swift
+++ b/BraveRewardsUI/Ads/AdsViewController.swift
@@ -45,7 +45,8 @@ public class AdsViewController: UIViewController {
       $0.top.greaterThanOrEqualTo(view).offset(4) // Makes sure in landscape its at least 4px from the top
       
       if UIDevice.current.userInterfaceIdiom == .pad {
-        $0.width.equalTo(max(view.bounds.width, view.bounds.height) * 0.40).priority(.high)
+        let isWidthLarger = UIScreen.main.bounds.width > UIScreen.main.bounds.height
+        $0.width.equalTo(isWidthLarger ? view.snp.width : view.snp.height).multipliedBy(0.40).priority(.high)
       } else {
         $0.width.equalTo(view).priority(.high)
       }
--- a/BraveRewardsUI/Ads/AdsViewController.swift
+++ b/BraveRewardsUI/Ads/AdsViewController.swift
@@ -41,7 +41,7 @@ public class AdsViewController: UIViewController {
       $0.leading.greaterThanOrEqualTo(view).inset(8)
       $0.trailing.lessThanOrEqualTo(view).inset(8)
       $0.centerX.equalTo(view)
-      $0.width.equalTo(400).priority(.high)
+      $0.width.equalTo(view).priority(.high)
       $0.top.equalTo(view.safeAreaLayoutGuide.snp.top)
       $0.top.greaterThanOrEqualTo(view).offset(4) // Makes sure in landscape its at least 4px from the top
     }
@@ -299,15 +299,24 @@ extension AdsViewController: UIGestureRecognizerDelegate {
 }
 
 extension AdsViewController {
+  
   /// Display a "My First Ad" on a presenting controller and be notified if they tap it
   public static func displayFirstAd(on presentingController: UIViewController, completion: @escaping (AdsNotificationHandler.Action, URL) -> Void) {
     let adsViewController = AdsViewController()
     
-    presentingController.addChild(adsViewController)
-    presentingController.view.addSubview(adsViewController.view)
-    adsViewController.didMove(toParent: presentingController)
+    guard let window = presentingController.view.window else {
+      return
+    }
+    
+    window.addSubview(adsViewController.view)
     adsViewController.view.snp.makeConstraints {
-      $0.edges.equalToSuperview()
+      $0.edges.top.bottom.equalToSuperview()
+      
+      if UIDevice.current.userInterfaceIdiom == .pad {
+        $0.edges.width.equalTo(max(window.bounds.width, window.bounds.height)).multipliedBy(0.40)
+      } else {
+        $0.edges.width.equalTo(400.0).priority(.high)
+      }
     }
     
     let notification = AdsNotification.customAd(
@@ -319,9 +328,7 @@ extension AdsViewController {
     adsViewController.display(ad: notification, handler: { (notification, action) in
       completion(action, notification.url)
     }, animatedOut: {
-      adsViewController.willMove(toParent: nil)
       adsViewController.view.removeFromSuperview()
-      adsViewController.removeFromParent()
     })
   }
 }
--- a/BraveRewardsUI/Ads/AdsViewController.swift
+++ b/BraveRewardsUI/Ads/AdsViewController.swift
@@ -41,9 +41,14 @@ public class AdsViewController: UIViewController {
       $0.leading.greaterThanOrEqualTo(view).inset(8)
       $0.trailing.lessThanOrEqualTo(view).inset(8)
       $0.centerX.equalTo(view)
-      $0.width.equalTo(view).priority(.high)
       $0.top.equalTo(view.safeAreaLayoutGuide.snp.top)
       $0.top.greaterThanOrEqualTo(view).offset(4) // Makes sure in landscape its at least 4px from the top
+      
+      if UIDevice.current.userInterfaceIdiom == .pad {
+        $0.width.equalTo(max(view.bounds.width, view.bounds.height) * 0.40).priority(.high)
+      } else {
+        $0.width.equalTo(view).priority(.high)
+      }
     }
     view.layoutIfNeeded()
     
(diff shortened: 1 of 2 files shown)


kylehickinson (changes requested):
1. You only fixed this for my first ad
2. I don't know that we should be adding this to the window. The reason for #1904 is simply because we have a dispatch after 3s but don't check to make sure we can show the first ad before it happens. We should fix #1904 by simply exiting early if we can't show it when the 3s `asyncAfter` goes off
3. Did we research whether or not the 40% size is true for both portrait & landscape iPad orientations?
4. We can't check if the interface is iPad, we have to check the horizontal size class. The user can be on an iPad and have it multitasked to a smaller window and running in iPhone layout.

Brandon-T:
image

image

BraveRewardsUI/Ads/AdsViewController.swift
```diff
@@ -299,15 +299,24 @@ extension AdsViewController: UIGestureRecognizerDelegate {
...
+      return
+    }
+    
+    window.addSubview(adsViewController.view)
     adsViewController.view.snp.makeConstraints {
-      $0.edges.equalToSuperview()
+      $0.edges.top.bottom.equalToSuperview()
+      
+      if UIDevice.current.userInterfaceIdiom == .pad {
+        $0.edges.width.equalTo(max(window.bounds.width, window.bounds.height)).multipliedBy(0.40)
```
kylehickinson: Also you're multiplying the height by 40% here as well

Brandon-T:
> 1. You only fixed this for my first ad
> 2. I don't know that we should be adding this to the window. The reason for #1904 is simply because we have a dispatch after 3s but don't check to make sure we can show the first ad before it happens. We should fix #1904 by simply exiting early if we can't show it when the 3s `asyncAfter` goes off
> 3. Did we research whether or not the 40% size is true for both portrait & landscape iPad orientations?
> 4. We can't check if the interface is iPad, we have to check the horizontal size class. The user can be on an iPad and have it multitasked to a smaller window and running in iPhone layout.

1. Hmm I don't know where else this should be fixed. All the steps to reproduce showed it on my first ad :S

2. I don't understand the issue. Why can't we show it on the window? It's above everything else and would behave like when you get a system notification whether or not the panel is open. Maybe it's better instead of messing with timings and async stuff? I don't know though.

3. It's both according to the ticket. 40% of 1000x600 is 40% of 1000 = 400 according to the ticket. Likewise if it's 500x700 then it'd be 40% of 700. I only took the numbers from the ticket. So formula became `max(width, height)` or `(width > height ? width : height) * 0.40`. So far it works in both portrait and landscape and split screen.

4. I tested multi-tasking. It seems to be working fine in every scenario because I'm not constraining it to the device size. I'm constraining it to the size of the current window's bounds. Not sure why I need size class?

kylehickinson:
I can see that bleeding into the status bar (due to not constraining to safe area).
Shouldn't it be smaller then if its 40% of the width of the **window**? It seems like the 40% is not even working at all?

Brandon-T:
> I can see that bleeding into the status bar (due to not constraining to safe area).
> Shouldn't it be smaller then if its 40% of the width of the **window**? It seems like the 40% is not even working at all?

I will check the 40% calculation again to make sure it's working. It is constrained to safeArea (original code): `$0.top.equalTo(view.safeAreaLayoutGuide.snp.top)`. Guessing it doesn't like `UIWindow`.. I'll check it and see.

kylehickinson:
> 1. Hmm I don't know where else this should be fixed. All the steps to reproduce showed it on my first ad :S

https://github.com/brave/brave-ios/blob/802db557a60a894568088890535a300ec6a93fd4/BraveRewardsUI/Ads/AdsNotificationHandler.swift#L47

> 2. I don't understand the issue. Why can't we show it on the window? It's above everything else and would behave like when you get a system notification whether or not the panel is open. Maybe it's better instead of messing with timings and async stuff? I don't know though.

Mostly to avoid odd states. We only show ads when the user is viewing the browser. First ad is separate entirely, its duplicated logic basically but it doesn't trigger the same way a normal ad would. We would never present a real ad when the users rewards panel is up, seen here: https://github.com/brave/brave-ios/blob/802db557a60a894568088890535a300ec6a93fd4/BraveRewardsUI/Ads/AdsNotificationHandler.swift#L121

> 3. It's both according to the ticket. 40% of 1000x600 is 40% of 1000 = 400 according to the ticket. Likewise if it's 500x700 then it'd be 40% of 700. I only took the numbers from the ticket. So formula became max(width, height) or (width > height ? width : height) * 0.40.

I meant, a real system notification. I'm not sure if the ticket is 100% correct/investigated on whether 40% is true for system notifications in both orientations.

> 4. I tested multi-tasking. It seems to be working fine in every scenario because I'm not constraining it to the device size. I'm constraining it to the size of the current window's bounds.

I have a feeling when the 40% code works, this will break because you're saying you want it to be 40% of the width of the window on iPad regardless of size class, so it'd end up being super skinny (192pt's wide on a 320pt 1/4 multitask window)

> I will check the 40% calculation again to make sure it's working. However, the status bar problem is from the original code: $0.top.equalTo(view.safeAreaLayoutGuide.snp.top). Guessing it doesn't like UIWindow.. I'll check it and see.

See, already issues using UIWindow 😆

Brandon-T:
@kylehickinson Fixed.. Now it truly works in ALL sizes, orientations and devices and has the correct SafeAreaInsets  and is on UIWindow with no dirty hacks.

Respects the 40% if it can otherwise falls back to full width with proper insets.
Also works with the first ad when it shows up while the panel is open..

Now the only question is if we want it on `UIWindow` or not. Other than that, the constraints are fine and have no warnings either.
Your call if you want it or not.

Screen Shot 2019-11-11 at 3 12 52 PM
Screen Shot 2019-11-11 at 3 12 48 PM
Screen Shot 2019-11-11 at 3 12 41 PM
Screen Shot 2019-11-11 at 3 12 34 PM
Screen Shot 2019-11-11 at 3 12 29 PM

BraveRewardsUI/Ads/AdsViewController.swift
```diff
@@ -41,9 +41,14 @@ public class AdsViewController: UIViewController {
       $0.leading.greaterThanOrEqualTo(view).inset(8)
       $0.trailing.lessThanOrEqualTo(view).inset(8)
       $0.centerX.equalTo(view)
-      $0.width.equalTo(400).priority(.high)
```
Brandon-T, line 44: Not sure why this was hardcoded to 400 in the first place.
Seems to work fine with `view`.

kylehickinson: It was so that on larger screens it maxed out at 400pt's

kylehickinson:
Still missing fixes to real ads (point 1 above)

BraveRewardsUI/Ads/AdsViewController.swift
```diff
@@ -41,9 +41,14 @@ public class AdsViewController: UIViewController {
...
       $0.top.equalTo(view.safeAreaLayoutGuide.snp.top)
       $0.top.greaterThanOrEqualTo(view).offset(4) // Makes sure in landscape its at least 4px from the top
+      
+      if UIDevice.current.userInterfaceIdiom == .pad {
+        $0.width.equalTo(max(view.bounds.width, view.bounds.height) * 0.40).priority(.high)
```
kylehickinson: Also, if the user multitasks while the ad is currently visible on screen (possible I guess since it stays up for 30s) -- The width of this isn't going to update because its constrained to a constant

kylehickinson (changes requested):

BraveRewardsUI/Ads/AdsViewController.swift
Brandon-T: @kylehickinson I tested the multi-tasking, it still works. I can resize the split view controller and this add resizes with it because of the less than or equal to constraint. Alternatively, I can determine the `max(width, height)` and then constraint to the actual `NSLayoutConstraint` instead with a multiplier.. but so far I haven't seen anything broken with the way it is now

For actual ads, I can't see why I need to fix it there since the actual ads won't show when the panel is open anyway and won't show when the controller isn't top view controller.. but I've added the change there too anyway.

kylehickinson: This is what I mean:

IMG_93650C18A835-1

STR:
1. Start multitasking at onboarding start with another app (I choose App Store), half half
2. Make App Store take up 2/3 (making the size of the ad shrink)
3. Make Brave take entire screen

Ad doesn't resize back to 40% because its a constant, it'll always stay as whatever 40% was when it first got created

Brandon-T: Fixed.!

BraveRewardsUI/Ads/AdsViewController.swift
```diff
@@ -41,9 +41,15 @@ public class AdsViewController: UIViewController {
       $0.leading.greaterThanOrEqualTo(view).inset(8)
       $0.trailing.lessThanOrEqualTo(view).inset(8)
       $0.centerX.equalTo(view)
-      $0.width.equalTo(400).priority(.high)
       $0.top.equalTo(view.safeAreaLayoutGuide.snp.top)
       $0.top.greaterThanOrEqualTo(view).offset(4) // Makes sure in landscape its at least 4px from the top
+      
+      if UIDevice.current.userInterfaceIdiom == .pad {
+        let isWidthLarger = UIScreen.main.bounds.width > UIScreen.main.bounds.height
```
kylehickinson, line 48: Is this what we have to do to determine whether or not we're in landscape?…

kylehickinson (approved):

PR #1937, opened 2019-11-12
No Bug: Fix deprecation warnings related to bytes operations.

iccub:
## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Shared/Bytes.swift
+++ b/Shared/Bytes.swift
@@ -13,11 +13,13 @@ open class Bytes {
     open class func generateRandomBytes(_ len: UInt) -> Data {
         let len = Int(len)
         var data = Data(count: len)
-        data.withUnsafeMutableBytes { (p: UnsafeMutablePointer<UInt8>) in
-            if SecRandomCopyBytes(kSecRandomDefault, len, p) != errSecSuccess {
+        data.withUnsafeMutableBytes {
+            if let baseAddress = $0.baseAddress,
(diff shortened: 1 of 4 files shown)


Shared/Bytes.swift
```diff
@@ -11,14 +11,12 @@ public typealias GUID = String
  */
 open class Bytes {
     open class func generateRandomBytes(_ len: UInt) -> Data {
-        let len = Int(len)
-        var data = Data(count: len)
-        data.withUnsafeMutableBytes { (p: UnsafeMutablePointer<UInt8>) in
-            if SecRandomCopyBytes(kSecRandomDefault, len, p) != errSecSuccess {
```
Brandon-T, line 17: Your new code definitely works the same and the outcome is the same (a cryptographically strong random array of bytes).

However, to fix this warning we don't actually have to change the functions being called. We can use the `baseAddress` instead which is what is recommended:

IE:

```
data.withUnsafeMutableBytes {
  if let baseAddress = $0.baseAddress, SecRandomCopyBytes(kSecRandomDefault, len, baseAddress) != errSecSuccess {
    fatalError("Random byte generation failed.")
  }
}
```

Which is the exact same minus the deprecation warning :)

Approved because they are equivalent anyway.

Brandon-T (approved):

iccub:
Since we went for the easy solution that doesn't change the crypto under the hood I removed the security review label

Brandon-T (approved):

Issue #1609, opened 2019-10-08
Does brave now support NFC authentication on iOS 13? (Yubikey 5 NFC for example)


Nexulo:
I'm not sure but i read several articles about the fact that brave is the first browser which supports the authentication over NFC on iOS 13? 

I tested that but it's not working, does someone know more about that? Will there come an update for brave?

Best regards
maisen20

Nexulo:
Is there someone working on or are there any news about that? Sorry but don't know how that works here on github.

jumde:
@maisen20 - NFC integration is in progress. Webauthn/U2F using lightning keys should work fine.

Nexulo:
> @maisen20 - NFC integration is in progress. Webauthn/U2F using lightning keys should work fine.

USB-C keys (YubiKey 5C Nano) on iPad Pro do not work, is that right? Is the integration of this in progress too?

jumde:
That is correct, we are tracking that as a separate issue: https://github.com/brave/brave-ios/issues/2019

anthonypkeane:
Moving milestone to 1.14.2

kjozwiak:
It looks like iPhone 6s & 6s+ only support Payments and don't support Read/Writes as per https://www.bluebite.com/nfc/iphone-nfc-compatibility. I checked a few other sources and basically found the same information. I'll go through the lighting port Yubico keys to make sure they're still working. @srirambv will go through NFC has he as an iPhone that supports it.

kjozwiak:
Prerequisite: using a Yubico key with lighting connector with `FW 5.2.3`

Verification **PASSED** on `iPhone 6s+` running `iOS 12.4.1` using `1.14.2 (19.12.11.21)`

* went through `https://demo.yubico.com/webauthn` and ensured the following
   * received the correct animation screen when the Yubico key is disconnected and inserted
   * ensured that quickly disconnecting/connecting the Yubico key doesn't cause any issues/crashes
   * completed the process several times without any issues
   * ensured cancelling/retrying several times doesn't crash/reproduce https://github.com/brave/brave-ios/issues/1388
* ensured that `https://webauthn.io` worked without any issues
* ensured that `https://mdp.github.io/u2fdemo/` worked without any issues
* ensured that `https://twitter.com/` worked without any issues
* ensured that `https://github.com/` worked without any issues

Verification **PASSED** on `iPad Mini 4` running `iOS 12.4.1` using `1.14.2 (19.12.11.21)`

[...]

PR #1939, opened 2019-11-12
Adding support for NFC Yubikeys

jumde:
## Summary of Changes

This pull request fixes issue https://github.com/brave/brave-ios/issues/1609

## Submitter Checklist:

## Test Plan:
Please run through the complete FIDO/FIDO2 support test plan(https://github.com/brave/brave-ios/pull/1158) for this issue. 

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Info.plist
+++ b/Client/Info.plist
@@ -69,6 +69,8 @@
 	<string>$(DEVELOPMENT_TEAM)</string>
 	<key>MozWhatsNewTopic</key>
 	<string></string>
+        <key>NFCReaderUsageDescription</key>
+        <string>The application needs access to NFC reading to communicate with your Yubikey.</string>
 	<key>NSAppTransportSecurity</key>
 	<dict>
 		<key>NSAllowsArbitraryLoads</key>
@@ -154,5 +156,10 @@
 	</array>
 	<key>UIFileSharingEnabled</key>
 	<true/>
+        <key>com.apple.developer.nfc.readersession.iso7816.select-identifiers</key>
+        <array>
+    	    <string>A000000527471117</string>
+            <string>A0000006472F0001</string>
+        </array>
 </dict>
 </plist>
--- a/Client/WebAuthN/SceneObserver.swift
+++ b/Client/WebAuthN/SceneObserver.swift
@@ -27,7 +27,7 @@ class SceneObserver: NSObject {
     private var sceneActivationClosure: () -> Void
     
     // MARK: - Object lifecycle
-    
+    @available(iOS 13.0, *)
     init(sceneActivationClosure: @escaping () -> Void) {
         self.window = SceneObserver.applicationWindow
         self.sceneActivationClosure = sceneActivationClosure
--- a/Client/WebAuthN/SceneObserver.swift
+++ b/Client/WebAuthN/SceneObserver.swift
@@ -1,0 +1,75 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import UIKit
+
+/// The Core NFC APIs don't properly notify when the system Core NFC action sheet goes away. This issue
+/// is present since iOS 11, when the NDEFReaderSession was introduced, but Apple didn't provide since then
+/// a notification or other mechanism to detect this. To overcome this the SceneObserver uses a newly
+/// introduced notification in iOS 13, UISceneDidActivateNotification, to detect when the application scene
+/// becomes again active after a system scene is dismissed.
+///
+/// The Apple documentation states:
+///  "UIKit posts this notification for temporary interruptions, such as when displaying system alerts.".
+///  https://developer.apple.com/documentation/uikit/uiscenewilldeactivatenotification?language=objc
+
+class SceneObserver: NSObject {
+
+    private(set) static var applicationWindow: UIWindow!
+    
+    // This is set only once in the AppDelegate.
+    class func setupApplication(window: UIWindow) {
+        applicationWindow = window
+    }
+
+    private var window: UIWindow
+    private var sceneActivationClosure: () -> Void
+    
+    // MARK: - Object lifecycle
+    
+    init(sceneActivationClosure: @escaping () -> Void) {
+        self.window = SceneObserver.applicationWindow
+        self.sceneActivationClosure = sceneActivationClosure
+        super.init()
+        addSceneOnservation()
+    }
+    
+    deinit {
+        removeSceneOnservation()
+    }
+    
+    // MARK: - Scene Observation
+    
+    private var sceneDidActivateNotificationToken: NSObjectProtocol?
+    
+    private func addSceneOnservation() {
+        guard sceneDidActivateNotificationToken == nil else {
+            return
+        }
+        let center = NotificationCenter.default
+        let mainQueue = OperationQueue.main
+                        
+        if #available(iOS 13.0, *) {
+            sceneDidActivateNotificationToken = center.addObserver(forName: UIScene.didActivateNotification, object: nil, queue: mainQueue) { [weak self] notification in
+                guard let self = self else {
+                    return
+                }
+                guard let scene = notification.object as? UIScene else {
+                    return
+                }
+                if self.window.windowScene == scene {
+                    self.sceneActivationClosure()
+                }
+            }
+        }
+    }
+    
+    private func removeSceneOnservation() {
+        guard let token = sceneDidActivateNotificationToken else {
+            return
+        }
+        let center = NotificationCenter.default
+        center.removeObserver(token)
+    }
+}
--- a/Client/WebAuthN/U2FExtensions.swift
+++ b/Client/WebAuthN/U2FExtensions.swift
@@ -201,29 +201,25 @@ class U2FExtensions: NSObject {
     
     /// Returns the service associated with the desired session.
     private var keyFido2Service: YKFKeyFIDO2ServiceProtocol? {
-        get {
-            if keyType == .accessory {
-                return YubiKitManager.shared.accessorySession.fido2Service
-            } else {
-                if #available(iOS 13.0, *) {
-                    return YubiKitManager.shared.nfcSession.fido2Service
-                }
+        if keyType == .accessory {
+            return YubiKitManager.shared.accessorySession.fido2Service
+        } else {
+            if #available(iOS 13.0, *) {
+                return YubiKitManager.shared.nfcSession.fido2Service
             }
-            return nil
         }
+        return nil
     }
     
     private var keyU2fService: YKFKeyU2FServiceProtocol? {
-        get {
-            if keyType == .accessory {
-                return YubiKitManager.shared.accessorySession.u2fService
-            } else {
-                if #available(iOS 13.0, *) {
-                    return YubiKitManager.shared.nfcSession.u2fService
-                }
+        if keyType == .accessory {
+            return YubiKitManager.shared.accessorySession.u2fService
+        } else {
+            if #available(iOS 13.0, *) {
+                return YubiKitManager.shared.nfcSession.u2fService
             }
-            return nil
         }
+        return nil
     }
 
     init(tab: Tab) {
--- a/Client/WebAuthN/U2FExtensions.swift
+++ b/Client/WebAuthN/U2FExtensions.swift
@@ -78,21 +78,21 @@ class U2FExtensions: NSObject {
     
     // There can be multiple Registration/Authentication requests at any time
     // We use handles to keep track of different requests
-    fileprivate var fido2RegHandles: [Int] = []
-    fileprivate var fido2AuthHandles: [Int] = []
-    fileprivate var fidoRegHandles: [Int] = []
-    fileprivate var fidoSignHandles: [Int] = []
+    private var fido2RegHandles: [Int] = []
+    private var fido2AuthHandles: [Int] = []
+    private var fidoRegHandles: [Int] = []
+    private var fidoSignHandles: [Int] = []
     
-    fileprivate var requestId: [Int: Int] = [:]
-    fileprivate var fidoRequests: [Int: Int] = [:]
+    private var requestId: [Int: Int] = [:]
+    private var fidoRequests: [Int: Int] = [:]
     
-    fileprivate var fido2RegisterRequest: [Int: WebAuthnRegisterRequest] = [:]
-    fileprivate var fido2AuthRequest: [Int: WebAuthnAuthenticateRequest] = [:]
+    private var fido2RegisterRequest: [Int: WebAuthnRegisterRequest] = [:]
+    private var fido2AuthRequest: [Int: WebAuthnAuthenticateRequest] = [:]
     
-    fileprivate var fidoRegisterRequest: [Int: FIDORegisterRequest] = [:]
-    fileprivate var fidoSignRequests: [Int: [FIDOSignRequest]] = [:]
+    private var fidoRegisterRequest: [Int: FIDORegisterRequest] = [:]
+    private var fidoSignRequests: [Int: [FIDOSignRequest]] = [:]
     
-    fileprivate static var observationContext = 0
+    private static var observationContext = 0
     
     private var nfcSesionStateObservation: NSKeyValueObservation?
     private var sceneObserver: SceneObserver?
@@ -100,21 +100,21 @@ class U2FExtensions: NSObject {
     // Popup modals presented to the user in different session or key states
     
     /// Show when user has to touch his auth key
-    fileprivate let touchKeyPopup = AlertPopupView(imageView: lottieAnimation(for: "webauth_touch_key"),
+    private let touchKeyPopup = AlertPopupView(imageView: lottieAnimation(for: "webauth_touch_key"),
                                                    title: Strings.touchKeyMessage, message: "", titleWeight: .semibold, titleSize: 21)
     
     /// Show when user's key hasn't been inserted yet
-    fileprivate let insertKeyPopup = AlertPopupView(imageView: lottieAnimation(for: "webauth_insert_key"),
+    private let insertKeyPopup = AlertPopupView(imageView: lottieAnimation(for: "webauth_insert_key"),
                                                     title: Strings.insertKeyMessage, message: "", titleWeight: .semibold, titleSize: 21)
  
     /// Show to enter key's pin
-    fileprivate let pinVerificationPopup = AlertPopupView(imageView: UIImageView(image: #imageLiteral(resourceName: "enter_pin")),
+    private let pinVerificationPopup = AlertPopupView(imageView: UIImageView(image: #imageLiteral(resourceName: "enter_pin")),
                                                           title: Strings.pinTitle, message: "",
                                                           inputType: .default, secureInput: true,
                                                           inputPlaceholder: Strings.pinPlaceholder, titleWeight: .semibold, titleSize: 21)
     
     /// Show when key's pin authentication is pending
-    fileprivate let verificationPendingPopup =
+    private let verificationPendingPopup =
         AlertPopupView(imageView: lottieAnimation(for: "webauth_verify_key"),
                        title: Strings.verificationPending, message: "", titleWeight: .semibold, titleSize: 21)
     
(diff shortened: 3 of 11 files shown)


Client/WebAuthN/U2FExtensions.swift
```diff
@@ -88,6 +94,8 @@ class U2FExtensions: NSObject {
     
     fileprivate static var observationContext = 0
     
+    fileprivate var nfcSesionStateObservation: NSKeyValueObservation?
```
Brandon-T: Should use the most strict accessor for the required scope.
For me, it looks like these should be `private` instead of `fileprivate`

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -121,83 +129,157 @@ class U2FExtensions: NSObject {
...
+    /**
+     Returns the service associated with the desired session.
+     */
+    private var keyFido2Service: YKFKeyFIDO2ServiceProtocol? {
+        get {
+            if keyType == .accessory {
+                return YubiKitManager.shared.accessorySession.fido2Service
+            } else {
+                if #available(iOS 13.0, *) {
+                    return YubiKitManager.shared.nfcSession.fido2Service
```
Brandon-T: Hmm.. is there no backwards compatibility for when it's not just an accessory?
iOS 11+ supports CoreNFC as well.

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -121,83 +129,157 @@ class U2FExtensions: NSObject {
...
-        
-        [touchKeyPopup, insertKeyPopup, pinVerificationPopup].forEach {
+
+        [touchKeyPopup, pinVerificationPopup].forEach {
             $0.addButton(title: Strings.keyCancel, tapped: handleCancelButton)
         }
         
         // Make sure the session is started
-        YubiKitManager.shared.keySession.startSession()
+        YubiKitManager.shared.accessorySession.startSession()
```
Brandon-T, line 244: Do we need to start `nfcSession` when on iOS 13+? I ask because the code before this shows `nfcSession` and it's guarded by an if-available statement and doesn't use `accessorySession`.

Brandon-T (changes requested):

Client/WebAuthN/U2FExtensions.swift
jumde: Read/Write capability for NFC was supported in iOS 13, so the YubiKit APIs are gated by this check. For iOS 11/12, we just present the option to use lightning keys.

Client/WebAuthN/U2FExtensions.swift
jumde: We start the nfcsession here: https://github.com/brave/brave-ios/pull/1939/files#diff-de38284f9d7a27ae89ede4cf95b6aad6R730

NFC session unlike accessory session have a blocking UI. So we cannot enable it at start.

bootstrap.sh
```diff
@@ -98,7 +98,6 @@ git clone -b master --single-branch https://github.com/YubicoLabs/yubikit-ios/ $
 
 mkdir -p $SRCDIR/$YUBIKIT_OUT/include
 pushd $YUBIKIT_DIR/YubiKit
-git checkout 2.0.1
```
iccub, line 101: Do you think it's better to always get fresh master build instead of pinning one version?

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -121,83 +129,157 @@ class U2FExtensions: NSObject {
...
+                keySession.removeObserver(self, forKeyPath: #keyPath(YKFAccessorySession.fido2Service.keyState))
+                keySession.removeObserver(self, forKeyPath: #keyPath(YKFAccessorySession.u2fService.keyState))
             }
         }
     }
+    
+    private var keyType: FIDO2ViewControllerKeyType = .unknown
+    
+    /**
+     Returns the service associated with the desired session.
```
iccub: We use 3slash `///` format for documentation

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -658,17 +724,24 @@ class U2FExtensions: NSObject {
...
+        } else {
+            insertKeyPopup.update(title: Strings.insertKeyMessage + currentURL)
+            insertKeyPopup.showWithType(showType: .flyUp)
+        }
     }
     
     // This modal is presented when the key bootstrap is complete
     private func presentInteractWithKeyModal() {
-        guard let fido2Service = YubiKitManager.shared.keySession.fido2Service else {
+        defer {observeServiceStateUpdates = true}
```
iccub: Add space between brackets,  not sure if swiftlint catches it

Client/Info.plist
```diff
@@ -154,5 +156,10 @@
 	</array>
 	<key>UIFileSharingEnabled</key>
 	<true/>
+        <key>com.apple.developer.nfc.readersession.iso7816.select-identifiers</key>
+        <array>
+    	    <string>A000000527471117</string>
```
iccub, line 161: What's this magic numbers?

iccub (changes requested):
Looks good overall. Can you provide some steps on how to test it?
I want to run this code

Client/Info.plist
Brandon-T: This is the tag's AID (Application Identifier). A complete list can be found here (even contains Yubico's other AIDs): https://www.eftlab.com/knowledge-base/211-emv-aid-rid-pix/

but that specific AID (`A000000527471117`) doesn't seem to be on the list. I assume it's new and possibly belonging to a beta key or a new Yubico key or something?

jumde: You can find more details about the identifiers here: https://github.com/Yubico/yubikit-ios/blob/master/docs/NFC-Notes.md

bootstrap.sh
jumde: I'll be pinning the change to yubikit release candidate, removed this temporarily since there are some fixes that need to be merged for the error handling to work correctly.

jumde: @iccub - Pinned to v3.0.0-Preview here: https://github.com/brave/brave-ios/pull/1939/files#diff-3af436e6760a0c5f1e3d16a3384853a8R97

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -121,83 +130,145 @@ class U2FExtensions: NSObject {
...
+                keySession.removeObserver(self, forKeyPath: #keyPath(YKFAccessorySession.u2fService.keyState))
+            }
+        }
+    }
+    
+    private var keyType: FIDO2ViewControllerKeyType = .unknown
+    
+    /// Returns the service associated with the desired session.
+    private var keyFido2Service: YKFKeyFIDO2ServiceProtocol? {
+        get {
```
iccub: `get` closure can be removed.
It should be used only when adding a setter `set {}` as well

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -121,83 +130,145 @@ class U2FExtensions: NSObject {
...
+                if #available(iOS 13.0, *) {
+                    return YubiKitManager.shared.nfcSession.fido2Service
+                }
             }
+            return nil
+        }
+    }
+    
+    private var keyU2fService: YKFKeyU2FServiceProtocol? {
+        get {
```
iccub: remove `get ` closure

Client/WebAuthN/SceneObserver.swift
```diff
@@ -0,0 +1,75 @@
...
+    class func setupApplication(window: UIWindow) {
+        applicationWindow = window
+    }
+
+    private var window: UIWindow
+    private var sceneActivationClosure: () -> Void
+    
+    // MARK: - Object lifecycle
+    
+    init(sceneActivationClosure: @escaping () -> Void) {
```
iccub, line 31: I would set `@available(iOS 13.0, *)` here, to make sure it never gets called in earlier versions. I know we have some version guard already present in U2FExtensions

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -373,26 +431,49 @@ class U2FExtensions: NSObject {
...
                 
-                guard let response = response else {
-                    self.sendFIDO2RegistrationError(handle: handle)
-                    return
+                fido2Service.execute(makeCredentialRequest) { [weak self] response, error in
+                    guard let self = self else {
+                        log.error(U2FErrorMessages.ErrorRegistration.rawValue)
+                        return
+                    }
+                    guard error == nil else {
```
iccub: Switch to `if`. We generally use `guard` only if an object needs to be unwrapped

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -446,20 +527,43 @@ class U2FExtensions: NSObject {
...
         
-        handlePinVerificationRequired { [weak self] (error) in
+        let requestExecutionClosure = { [weak self] in
             guard let self = self else {
                 return
             }
-            self.cleanupPinVerificationPopup()
-            ensureMainThread {
-                self.verificationPendingPopup.dismissWithType(dismissType: .flyDown)
+            self.handlePinVerificationRequired { [weak self] (error) in
```
iccub: Remove brackets around `error`

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -446,20 +527,43 @@ class U2FExtensions: NSObject {
...
+            requestExecutionClosure()
+        } else {
+            guard #available(iOS 13.0, *) else {
                 self.sendFIDO2RegistrationError(handle: handle)
                 return
             }
-            self.handleFIDO2Registration(handle: handle, request: request)
+            
+            YubiKitManager.shared.nfcSession.stopIso7816Session()
+            self.sceneObserver = SceneObserver(sceneActivationClosure: {  [weak self] in
```
iccub: Remove extra space here `{  [weak self]`

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -557,28 +656,51 @@ class U2FExtensions: NSObject {
...
+                    }
+                    if self.keyType == .accessory {
+                        self.finalizeFIDO2Authentication(handle: handle, response: response, clientDataJSON: clientDataJSON, error: nil)
+                    } else {
+                        guard #available(iOS 13.0, *) else {
+                            self.sendFIDO2RegistrationError(handle: handle)
+                            return
+                        }
+                        
+                        self.sceneObserver = SceneObserver(sceneActivationClosure: {  [weak self] in
```
iccub: remove extra space `{  [weak self]`

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -557,28 +656,51 @@ class U2FExtensions: NSObject {
...
-                // The reponse from the key must not be empty at this point.
-                guard let response = response else {
-                    self.sendFIDO2AuthenticationError(handle: handle)
-                    return
+                fido2Service.execute(getAssertionRequest) { [weak self] response, error in
+                    guard let self = self else {
+                        log.error(U2FErrorMessages.ErrorAuthentication.rawValue)
+                        return
+                    }
+                    guard error == nil else {
```
iccub: convert to `if error != nil`

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -686,8 +845,12 @@ class U2FExtensions: NSObject {
         ensureMainThread {
             let currentURL = self.tab?.url?.host ?? ""
             self.pinVerificationPopup.addButton(title: Strings.confirmPin, type: .primary) { [weak self] in
-                self?.verificationPendingPopup.update(title: Strings.verificationPending + currentURL)
-                self?.verificationPendingPopup.showWithType(showType: .flyUp)
+                // Only show Verification UI for accessory since Core NFC UI
+                // is displayed on top of it
+                if self?.keyType == .accessory {
```
iccub: could unwrap `self` as you did in many other places

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -697,19 +860,28 @@ class U2FExtensions: NSObject {
...
+                return
+            }
+            guard let fido2Service = self.keyFido2Service else {
                 completion(true)
                 return
             }
-            completion(false)
-            return
+            
+            fido2Service.execute(verifyPinRequest) { (error) in
```
iccub: Remove bracket around `error`

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -697,19 +860,28 @@ class U2FExtensions: NSObject {
...
+                guard error == nil else {
```
iccub: convert to `if`

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -697,19 +860,28 @@ class U2FExtensions: NSObject {
...
+                    completion(true)
+                    return
+                }
+                completion(false)
+                return
```
iccub: This return isn't needed

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -758,70 +919,93 @@ class U2FExtensions: NSObject {
...
-                        if error != nil {
-                            let errorDescription = error?.localizedDescription ?? U2FErrorMessages.ErrorRegistration.rawValue
-                            log.error(errorDescription)
+            u2fservice.execute(registerRequest) { [weak self] response, error in
+                guard let self = self else {
+                    log.error(U2FErrorMessages.Error.rawValue)
+                    return
+                }
+                
+                guard error == nil else {
```
iccub: convert to `if`

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -758,70 +919,93 @@ class U2FExtensions: NSObject {
...
+                    let errorMessage = error?.localizedDescription ?? Strings.U2FRegistrationError
+                    self.sendFIDORegistrationError(handle: handle, requestId: requestId, errorCode: U2FErrorCodes.other_error, errorMessage: errorMessage)
+                    return
+                }
+                guard let clientData = response?.clientData.websafeBase64String() else {
```
iccub: `clientData` and `registrationData` could go under the same `guard` condition, both return the same error

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -901,60 +1076,92 @@ class U2FExtensions: NSObject {
...
+                u2fservice.execute(signRequest) { [weak self] response, error in
+                    guard let self = self else {
+                        log.error(U2FErrorMessages.ErrorAuthentication.rawValue)
+                        return
+                    }
+                    
+                    count -= 1
+                    self.fidoRequests[handle] = count
+                    
+                    guard error == nil else {
```
iccub: convert to `if`

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -901,60 +1076,92 @@ class U2FExtensions: NSObject {
...
+                        if count == 0 {
+                            let errorMessage = error?.localizedDescription ?? Strings.U2FAuthenticationError
+                            self.sendFIDOAuthenticationError(handle: handle, requestId: requestId, errorCode: U2FErrorCodes.other_error, errorMessage: errorMessage)
+                        }
+                        return
+                    }
+                    guard let keyHandle = response?.keyHandle else {
```
iccub: `keyHandle`, `signature` and `clientData` can go under the same `guard` condition, just split it with new lines to avoid one long line

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -979,20 +1186,78 @@ class U2FExtensions: NSObject {
...
             return
         }
         fidoSignHandles.remove(at: index)
         fidoSignRequests.removeValue(forKey: handle)
         requestId.removeValue(forKey: handle)
-        YubiKitManager.shared.keySession.cancelCommands()
+        cleanup()
+    }
+    
+    private func executeKeyRequestWith(execution: @escaping () -> Void) {
```
iccub: I think the word `With` isn't needed here

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -979,20 +1186,78 @@ class U2FExtensions: NSObject {
...
+            if nfcSession.iso7816SessionState == .open {
+                execution()
+                return
+            }
+            
+            // The ISO7816 session is started only when required since it's blocking the application UI with the NFC system action sheet.
+            YubiKitManager.shared.nfcSession.startIso7816Session()
+            
+            // Execute the request after the key(tag) is connected.
+            nfcSesionStateObservation = nfcSession.observe(\.iso7816SessionState, changeHandler: { [weak self] session, change in
```
iccub, line 1230: unwrap `self` with guard

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -979,20 +1186,78 @@ class U2FExtensions: NSObject {
...
+                if session.iso7816SessionState == .open {
+                    execution()
+                    self?.nfcActive = false
+                    self?.nfcSesionStateObservation = nil // remove the observation
```
iccub: don't need `return` after this executes?
Or do we want to potentially launch the code below

Client/WebAuthN/U2FExtensions.swift
```diff
@@ -979,20 +1186,78 @@ class U2FExtensions: NSObject {
...
+        defer {
+            observeSessionStateUpdates = true
+        }
+        
+        let accessoryState = YubiKitManager.shared.accessorySession.sessionState
+
+        if keyType == .nfc || accessoryState == .open { // The key session is ready to be used.
+            if accessoryState == .open {
+                if nfcActive, #available(iOS 13.0, *) {
+                    nfcActive = false
```
iccub, line 1265: Thinking whether we should `didSet` to `nfcActive` variable, so it calls `stopIso7816Session` always when it's set to false?
What do you think about it? Not 100% if we always want to stop the session when this property goes to `false`

iccub (changes requested):
The webauth/u2f code looks cleaner and cleaner with each iteration, good job

iccub:
I also run the app and it seems to be working well, so only code style changes required and I can approve this PR

Client/WebAuthN/U2FExtensions.swift
jumde: @iccub - I used the closure to return the `fido2Service` depending on the `keyType`. I can convert this to a function but a get closure felt cleaner. Let me know what you think.

iccub: I know, you can literally just remove the `get` keyword, and leave the rest

so it's
```
private var keyFido2Service: YKFKeyFIDO2ServiceProtocol? {
    // code logic
}
```

instead of 
```
private var keyFido2Service: YKFKeyFIDO2ServiceProtocol? {
    get {
        // code logic
    }
}
```

per Swift documentation
```
You can simplify the declaration of a read-only computed property by removing the get keyword and its braces:
```

Client/WebAuthN/U2FExtensions.swift
jumde: done!

iccub (approved):

Brandon-T (approved):

jhreis (approved):

bootstrap.sh
```diff
@@ -94,11 +94,10 @@ SRCDIR=$PWD
 rm -rf $YUBIKIT_DIR
 rm -rf $YUBIKIT_OUT
 
-git clone -b master --single-branch https://github.com/YubicoLabs/yubikit-ios/ $YUBIKIT_DIR
+git clone -b 3.0.0-Preview --single-branch https://github.com/Yubico/yubikit-ios/ $YUBIKIT_DIR
```
bridiver, line 97: this is incorrect, it should be https://github.com/Yubico/yubikit-ios.git

jumde: This works without `.git` as well. If its failing I can do a quick fix for this.

bridiver: it does not for me

bridiver:
```
Cloning into 'Carthage/Checkouts/yubikit-ios'...
Warning: Permanently added 'github.com,140.82.112.3' (RSA) to the list of known hosts.
ERROR: Repository not found.
fatal: Could not read from remote repository.
```

PR #2150, opened 2020-01-07
Fix #2145: Crash with Sync.getExistingObjects

### LINKED ISSUE #2145: Sync 'getExistingObjects' is still crashing.
### Steps to Reproduce 
Hard to reproduce, couldn't repro it even once

**Actual result:** 

Zrzut ekranu 2020-01-6 o 17 39 45

**Expected result:**
Shouldn't crash that much

**Brave Version:** 
It is happening since 1.11.4 or earlier

**Device details:** 

**Website problems only:**

### Additional Information
Previous attempt to fix it: https://github.com/brave/brave-ios/issues/2049

iccub:
## Summary of Changes
Call `context.object` can throw a fault sometimes. Changing it to `existingObject` so we can fail silently.
Not 100% sure this is going to help, doesn't hurt to add it though.

Please note that we use `context.object` in other places and never had problems, in the future we might consider replacing it with safer approach, using either `existingObject` or `registeredObject` methods

This pull request fixes issue #2145 

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Data/models/Device.swift
+++ b/Data/models/Device.swift
@@ -63,7 +63,12 @@ extension Device {
             return device
         }
         
-        return context.object(with: deviceId) as? Device
+        do {
+            return try context.existingObject(with: deviceId) as? Device
+        } catch {
+            log.error("Failed to fetch device: \(error)")
+            return nil
+        }
     }
     
     class func add(name: String?, isCurrent: Bool = false) {


kylehickinson (approved):
I assume we already handle nil returns from this method since its always been optional, so looks good. Didn't even know `existingObject(with:)` could throw

Issue #2197, opened 2020-01-14
Swift should use camelCase convention


iccub:
This is a cleanup task for our project, no code logic should change.

PR #2199, opened 2020-01-14
Fix wrong String name in TabMOTests

freak4pc:
I guess we missed this while working on #2197. CI is failing right now. #trivial


iccub (approved):
oops, thanks, we still haven't restored CI for external contributors, hope to get that this or next week

PR #2216, opened 2020-01-20
NTP Downloader

Brandon-T:
Security Review: https://github.com/brave/security/issues/59

- Added NTP Download to download the files from the server and optionally unzip them.
- Handles errors downloading by not overwriting existing files.
- Uses libarchive (ships with iOS so no third-party libs or linking issues) to unzip: `tar, bz2, zip, gz, tar.gz, tar.bz2, etc`.

## Summary of Changes

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,215 @@
...
+
+class NTPDownloader {
+    private static let BASE_URL = "https://brave-ntp-crx-input-dev.s3-us-west-2.amazonaws.com/"
+    
+    private let isZipped: Bool
+    private let defaultLocale = "US"
+    private let supportedLocales: [String] = []
+    private let currentLocale = Locale.current.regionCode
+    
+    init(isZipped: Bool = false) {
```
iccub: I would remove default param here, so when called we have have to be intentional whether zip or not is used

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,215 @@
...
+    func update() {
+        // Download the NTP Info to a temporary directory
+        self.download { url, error in
+            if let error = error {
+                print(error)
+                return
+            }
+            
+            guard let url = url else {
+                print(error)
```
iccub: Replace prints with logs after you're done with this PR

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,215 @@
...
+                return
+            }
+ 
+            do {
+                let downloadsURL = supportDirectory.appendingPathComponent("NTPDownloads")
+                if FileManager.default.fileExists(atPath: downloadsURL.absoluteString) {
+                    try FileManager.default.removeItem(at: downloadsURL)
+                }
+                
+                try FileManager.default.moveItem(at: url, to: downloadsURL)
```
iccub: I'm thinking whether we should do cleanup and remove old resources at some point.
Can be done later outside of this PR

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,215 @@
...
+            }
+            
+            //Move contents of `url` directory
+            //to somewhere more permanent where we'll load the images from..
+            guard let supportDirectory = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first else {
+                return
+            }
+ 
+            do {
+                let downloadsURL = supportDirectory.appendingPathComponent("NTPDownloads")
```
iccub: Extract folder name to a constant

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,215 @@
...
+                guard let data = data else {
+                    return completion(nil, "Invalid Photos.json for NTP Download")
+                }
+                
+                self.unzip(data: data) { url, error in
+                    completion(url, error)
+                }
+            }
+        } else {
+            self.download(path: "photo.json") {[weak self] data, error in
```
iccub: extract file name to a constant

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,215 @@
...
+struct NTPItemInfo: Codable {
+    let logoImageUrl: String
+    let logoAltText: String
+    let logoDestinationUrl: URL
+    let logoCompanyName: String
+    let wallpaperImageUrls: [String]
+}
+
+class NTPDownloader {
+    private static let BASE_URL = "https://brave-ntp-crx-input-dev.s3-us-west-2.amazonaws.com/"
```
iccub: Use camelCase `baseURL`

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,215 @@
...
+        if self.isZipped {
+            return url.appendingPathComponent(getSupportedLocale())
+                      .appendingPathExtension("zip")
+        }
+        
+        return url.appendingPathComponent(getSupportedLocale())
+    }
+    
+    private func getSupportedLocale() -> String {
+        guard let region = Locale.current.regionCode else {
```
iccub, line 305: Can use `currentLocale` constant here

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,215 @@
...
+            if let error = error {
+                return completion(nil, error)
+            }
+            
+            guard let response = response as? HTTPURLResponse else {
+                return completion(nil, error)
+            }
+            
+            if response.statusCode < 200 || response.statusCode > 299 {
+                return completion(nil, nil)
```
iccub: Would be good to return some error here containing status code, can return a simple String, as String conforms to error in our code(see `ErrorExtensions.swift`)

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,215 @@
...
+            }
+            
+            completion(data, nil)
+        }
+    }
+    
+    // Unzips Data to the temporary directory and returns a URL to the directory
+    private func unzip(data: Data, _ completion: @escaping (URL?, Error?) -> Void) {
+        let tempDirectory = FileManager.default.temporaryDirectory
+        let directory = tempDirectory.appendingPathComponent("NTPDownloads")
```
iccub: Use constant for folder name

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,215 @@
...
+        } catch {
+            completion(nil, error)
+        }
+    }
+    
+    // Unzips NTPItemInfo by downloading all of its assets to a temporary directory
+    // and returning the URL to the directory
+    private func unzip(item: NTPItemInfo, _ completion: @escaping (URL?, Error?) -> Void) {
+        let tempDirectory = FileManager.default.temporaryDirectory
+        let directory = tempDirectory.appendingPathComponent("NTPDownloads")
```
iccub: Use constant for folder name

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,215 @@
...
+    private let supportedLocales: [String] = []
+    private let currentLocale = Locale.current.regionCode
+    
+    init(isZipped: Bool = false) {
+        self.isZipped = isZipped
+    }
+    
+    func update() {
+        // Download the NTP Info to a temporary directory
+        self.download { url, error in
```
iccub: Don't we need `weak self` here?

Client/Frontend/Browser/HomePanel/Unzip.swift
```diff
@@ -0,0 +1,117 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+
+class Unzip {
+    public static func unpack_archive(data: Data, to path: String) throws {
+        var data = data
+        let count = data.count
+        let path = URL(string: path)!
```
iccub: Unwrap

Client/Frontend/Browser/HomePanel/Unzip.swift
```diff
@@ -0,0 +1,117 @@
...
+            fileHandle.closeFile()
+        }
+        
+        if archive_read_close(archive) != ARCHIVE_OK {
+            throw String(cString: archive_error_string(archive), encoding: .utf8) ?? "Unknown Error Closing Archive"
+        }
+    }
+
+    public static func unpack_archive(from fromPath: String, to toPath: String) throws {
+        let path = URL(string: toPath)!
```
iccub: Unwrap

Client/Frontend/Browser/HomePanel/Unzip.swift
```diff
@@ -0,0 +1,117 @@
...
+        }
+        
+        do {
+            try FileManager.default.createDirectory(atPath: toPath, withIntermediateDirectories: true, attributes: nil)
+        } catch {
+            archiveHandle.closeFile()
+            throw error
+        }
+        
+        if archive_read_open_fd(archive, archiveHandle.fileDescriptor, 512) != ARCHIVE_OK {
```
iccub: Move that magic number to a variable to explain what it is for

Client/Frontend/Browser/HomePanel/libarchive.h
```diff
@@ -0,0 +1,152 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+//From https://github.com/libarchive/libarchive/blob/master/libarchive/archive.h
```
iccub: Move this file to `ThirdParty` folder is possible.
https://github.com/brave/brave-ios/tree/development/ThirdParty

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,215 @@
...
+                    return completion(nil, error)
+                }
+                
+                guard let data = data else {
+                    return completion(nil, "Invalid Photos.json for NTP Download")
+                }
+                
+                do {
+                    let item = try JSONDecoder().decode(NTPItemInfo.self, from: data)
+                    self.unzip(item: item) { url, error in
```
iccub: Why unzip here? Isn't this non-zip code branch?

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
iccub: I think we should add some logic to not re-download if the file hasn't changed on server.
For adblock resources we used etags to determine it

iccub (changes requested):
left some comments, will revisit this PR later as it's still WIP

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
Brandon-T: I couldn't think of a name for this function. It doesn't unzip, but it does the same job as the unzip function which is to grab all assets. This one downloads the assets individually though.

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
Brandon-T: Fixed.

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
Brandon-T: I used `moveItem` for that reason :D To move the entire folder from temp directory to application support where we can load the information from :)

Brandon-T:
image

Brandon-T:
image

iccub (changes requested):
CI fails with
```
BrowserViewController.swift:558:20: value of type 'NTPDownloader' has no member 'update'
```

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,378 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import Shared
+
+private let logger = Logger.browserLogger
+
+struct NTPFocalPoint: Codable {
```
iccub, line 11: Some of these structures will be duplicated with Joel's PR

https://github.com/brave/brave-ios/pull/2203/files#diff-516529ec2ce543f7446b5454f973fe52R11

We can consolidate them after both PR are merged, shouldn't be a problem for now

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,378 @@
...
+            }
+        }
+    }
+    
+    private func loadNTPInfo() -> NTPItemInfo? {
+        guard let supportDirectory = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first else {
+            return nil
+        }
+        
+        let downloadsURL = supportDirectory.appendingPathComponent(NTPDownloader.ntpDownloadsFolder)
```
iccub: nitpick: I feel it's easier to read if you name it `downloadsFolder` not `URL`, same for metaDataUrl etagURL

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,378 @@
...
+    }
+    
+    func getNTPInfo(_ completion: @escaping (NTPItemInfo?) -> Void) {
+        // Download the NTP Info to a temporary directory
+        self.download { [weak self] url, cacheInfo, error in
+            guard let self = self else { return }
+            
+            if let error = error {
+                logger.error(error)
+                return completion(self.loadNTPInfo())
```
iccub: Why not just return nil for these error paths?
Returning the same closure for each cases looks weird

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,378 @@
...
+            let metadata = try Data(contentsOf: metaDataURL)
+            let itemInfo = try JSONDecoder().decode(NTPItemInfo.self, from: metadata)
+            
+            let logo = NTPLogo(imageUrl: metaDataURL.appendingPathComponent(itemInfo.logo.imageUrl).path,
+                               alt: itemInfo.logo.alt,
+                               companyName: itemInfo.logo.companyName,
+                               destinationUrl: itemInfo.logo.destinationUrl)
+            
+            let wallpapers = itemInfo.wallpapers.map({
+                return NTPWallpaper(imageUrl: metaDataURL.appendingPathComponent($0.imageUrl).path,
```
iccub: `return` can be removed here

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,378 @@
...
+        do {
+            let metadata = try Data(contentsOf: metaDataURL)
+            let itemInfo = try JSONDecoder().decode(NTPItemInfo.self, from: metadata)
+            
+            let logo = NTPLogo(imageUrl: metaDataURL.appendingPathComponent(itemInfo.logo.imageUrl).path,
+                               alt: itemInfo.logo.alt,
+                               companyName: itemInfo.logo.companyName,
+                               destinationUrl: itemInfo.logo.destinationUrl)
+            
+            let wallpapers = itemInfo.wallpapers.map({
```
iccub: Rounded brackets after `map` can be removed

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,378 @@
...
+            try FileManager.default.createDirectory(at: downloadsURL, withIntermediateDirectories: true, attributes: nil)
+        } catch {
+            logger.error("NTPDownloader: \(error)")
+            return
+        }
+        
+        let etagURL = downloadsURL.appendingPathComponent(NTPDownloader.etagFile)
+        
+        do {
+            let etag = etag.replacingOccurrences(of: "\"", with: "")
```
iccub, line 240: Why drop the backward slash here?
Will this work when sending etag to the server for comparison?

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,378 @@
...
+    }
+    
+    private func removeETag() throws {
+        guard let supportDirectory = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first else {
+            logger.error("Cannot find Support Directory for writing NTPDownloader ETag")
+            return
+        }
+        
+        let downloadsURL = supportDirectory.appendingPathComponent(NTPDownloader.ntpDownloadsFolder)
+        let etagURL = downloadsURL.appendingPathComponent(NTPDownloader.etagFile)
```
iccub: This is called many times, I would extract it to a helper property `getDownloadsFolder`, can do similar for etag and metadata files too

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,378 @@
...
+        }
+        
+        if let etag = response.allHeaderFields["ETag"] as? String {
+            return CacheResponse(statusCode: response.statusCode, etag: etag)
+        }
+        
+        return CacheResponse(statusCode: response.statusCode, etag: "")
+    }
+    
+    // Downloads the item at the specified url relative to the BASE_URL
```
iccub: rename `BASE_URL` in comment to `baseUrl`

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,378 @@
...
+    private func unzip(item: NTPItemInfo, _ completion: @escaping (URL?, Error?) -> Void) {
+        let tempDirectory = FileManager.default.temporaryDirectory
+        let directory = tempDirectory.appendingPathComponent(NTPDownloader.ntpDownloadsFolder)
+        
+        do {
+            try FileManager.default.createDirectory(at: directory, withIntermediateDirectories: true, attributes: nil)
+            
+            var error: Error?
+            let group = DispatchGroup()
+            let urls = [item.logo.imageUrl] + item.wallpapers.map({ $0.imageUrl })
```
iccub: round brackets can be removed

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,378 @@
...
+        } else {
+            self.download(path: NTPDownloader.metadataFile, etag: self.getETag()) { [weak self] data, cacheInfo, error in
+                guard let self = self else { return }
+                
+                if let error = error {
+                    return completion(nil, nil, error)
+                }
+                
+                if let cacheInfo = cacheInfo {
+                    if cacheInfo.statusCode == 304 {
```
iccub: add this `if` to condition above, no need for two indentation levels here

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,378 @@
...
+        
+        let downloadsURL = supportDirectory.appendingPathComponent(NTPDownloader.ntpDownloadsFolder)
+        let etagURL = downloadsURL.appendingPathComponent(NTPDownloader.etagFile)
+        
+        if FileManager.default.fileExists(atPath: etagURL.path) {
+            try FileManager.default.removeItem(at: etagURL)
+        }
+    }
+    
+    private func download(_ completion: @escaping (URL?, CacheResponse?, Error?) -> Void) {
```
iccub: rename to `downloadMetaData`

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,378 @@
...
+    private let isZipped: Bool
+    private let defaultLocale = "US"
+    private let supportedLocales: [String] = []
+    private let currentLocale = Locale.current.regionCode
+    
+    init(isZipped: Bool) {
+        self.isZipped = isZipped
+    }
+    
+    func getNTPInfo(_ completion: @escaping (NTPItemInfo?) -> Void) {
```
iccub: from my tests, this completion always returns nil, resources are downloaded fine
I think reading file from disk always returns nil

iccub (changes requested):
I know this is still wip, leaving another small feedback anyway since we are on tight schedule

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
Brandon-T: Fixed.

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
Brandon-T: Fixed.

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
Brandon-T: It must load from the cache the NTPInfo when an error has occurred.

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
Brandon-T: Fixed.

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
jhreis: Yeah, let's leave this as-is, and I'll edit when merging. We'll just have to see where makes the most sense. I figured this would probably happen anyway, since we both need the data.

jhreis:
I'm hesitant to include all of the unzip logic in this code, as it is certainly not needed now. 

I think this would be better being abstracted, and just open sourced on the Brave Org or something. It is completely safe to assuming the files won't be zipped at this point.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -552,6 +554,10 @@ class BrowserViewController: UIViewController {
         Bookmark.restore_1_12_Bookmarks() {
             log.info("Bookmarks from old database were successfully restored")
         }
+        
+        downloader.getNTPInfo { info in
+            print(info)
```
jhreis: Remove print statement.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -443,6 +443,8 @@ class BrowserViewController: UIViewController {
         // Re-show toolbar which might have been hidden during scrolling (prior to app moving into the background)
         scrollController.showToolbars(animated: false)
     }
+    
+    let downloader = NTPDownloader()
```
jhreis: This should follow the adblock downloading approach, and execute inside `willFinishLaunchingWithOptions` inside app delegate.

jhreis (changes requested):

Client/Frontend/Browser/BrowserViewController.swift
jumde: @jhreis @Brandon-T - Shouldn't the download be gated by the brave rewards being enabled?

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -555,9 +553,9 @@ class BrowserViewController: UIViewController {
             log.info("Bookmarks from old database were successfully restored")
         }
         
-        downloader.getNTPInfo { info in
-            print(info)
-        }
+        
+        //Sample Usage of NTPDownloader
+        NTPDownloader.shared.delegate = self
```
iccub: Do we really need to use singleton? BrowserViewController lives for the whole app's lifecycle I think.

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -54,6 +91,9 @@ class NTPDownloader {
                 return completion(nil)
             }
             
+            //Start the timer no matter what..
+            self.startNTPTimer()
```
iccub, line 89: `self` needed here?

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -39,7 +44,39 @@ class NTPDownloader {
     private let supportedLocales: [String] = []
     private let currentLocale = Locale.current.regionCode
     
-    func getNTPInfo(_ completion: @escaping (NTPItemInfo?) -> Void) {
+    private var timer: Timer?
+    private var backgroundObserver: NSObjectProtocol?
+    private var foregroundObserver: NSObjectProtocol?
+    
+    public static let shared = NTPDownloader()
```
iccub: public not needed in anything from `Client` module

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -39,7 +44,39 @@ class NTPDownloader {
...
+    
+    public weak var delegate: NTPDownloaderDelegate? {
```
iccub: `public` not needed

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -93,8 +133,64 @@ class NTPDownloader {
...
     
-    private func loadNTPInfo() -> NTPItemInfo? {
+    private func notifyObservers() {
+        self.getNTPInfo { [weak self] item in
+            self?.delegate?.onNTPUpdated(ntpInfo: item)
+        }
+    }
+    
+    private func startNTPTimer() {
+        let baseTime = 5.0 * 60.0 * 60.0  //5 hours base time
```
iccub: Can use `5.hours` helper method

iccub (changes requested):

Client/Frontend/Browser/BrowserViewController.swift
Brandon-T: We don't need a singleton. I can remove it and we can allocate it.

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
Brandon-T: It's ran inside of a block.

iccub (approved):
Good job @Brandon-T, looking nice

Client/Frontend/Browser/HomePanel/NTPDownloader.swift
```diff
@@ -0,0 +1,460 @@
...
+                }
+                
+                try FileManager.default.moveItem(at: url, to: downloadsFolderURL)
+                
+                //Store the ETag
+                if let cacheInfo = cacheInfo {
+                    self.setETag(cacheInfo.etag)
+                }
+                
+                completion(self.loadNTPInfo())
```
jhreis: This should be outside the do-catch loop, if it is identical.

jhreis (changes requested):

jhreis (approved):

PR #2270, opened 2020-01-29
Fix #2262: Hide Sync's webview better.

### LINKED ISSUE #2262: translucent element visible in certain cases but always visible on iOS 12
### Description: 

You'll notice that there's a translucent box element that's basically visible in some cases. However, on `iPhone iOS 12`, it's basically visible constantly, even when browsing websites. Seems like this issue was around in `1.14.2` but has gotten a lost worse on `iPhone iOS 12`.

### Steps to Reproduce

  1. install & launch `1.14.3 (20.1.28.18)`
  2. go through onboarding and enable rewards/ads
  3. start browsing websites and you'll notice the that there's a translucent element on every page

It's hard to see, but it's also visible on the NTP as well.

**Actual result:** 

IMG_9986

IMG_9987

IMG_9996

**Expected result:**

Users shouldn't see a translucent element whenever browsing with Brave on iOS 12

**Brave Version:** 

* Using `1.14.3 (20.1.28.18)`

**Device details:** 

* `iPhone 6s+` running `iOS 12` - Reproduced

**Website problems only:**

### Additional Information

CCing @iccub @jhreis @anthonypkeane @srirambv

iccub:
It was visible in some cases.

## Summary of Changes

This pull request fixes issue #2262 

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -599,7 +599,7 @@ class BrowserViewController: UIViewController {
         #endif
         
         DispatchQueue.main.async {
-            sync.webView.alpha = 0.01
+            sync.webView.alpha = CGFloat.leastNormalMagnitude
             UIApplication.shared.keyWindow?.insertSubview(Sync.shared.webView, at: 0)
         }
     }


jhreis (approved):

jhreis:
This works for now, but we should attempt using JSCore, and avoiding a webview entirely.

Edit: Actually probably not worth it, since we're planning to move to C++ anyway.

PR #3035, opened 2020-11-12
Fix #2961: Problems with Sharing a PDF File

### LINKED ISSUE #2961: Problems with Sharing a PDF File
### Description: 

While Sharing a PDF file in Brave browser, some of the applications have problems.

- Mail app: Instead of attaching the PDF file, the URL of the PDF file is copied as message.
- iBooks: A pop-up showing "Creating a PDF" is shown and nothing happens

### Steps to Reproduce 
  1. Open a PDF file in the browser
  2. Tap 'share'
  3. Share vie 'Books.app.

**Actual result:** 
Nothing happens

**Expected result:**
- Mail app: Should Attach the PDF to Template Email
- iBooks: Should save the PDF to Books

100%

**Brave Version:** 
1.20

**Device details:** 
iOS 14, did not check previous versions

**Website problems only:**

### Additional Information
Reported by @PrivacyMatters

soner-yuksel:
WIP:

Fixed:
- Creating PDF from the WebPage and passing it to iBooks If activity is chosen
- Sharing Downloaded Files

Need to fix:

- Sharing PDF file with iBooks and Mail app. 

## Summary of Changes

This pull request fixes #2961

## Submitter Checklist:

## Test Plan:

- Load Brave Browser with a PDF link and Press Share from Menu
- Choose iBook Apps for sharing or Mail

## Screenshots:
N/A

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Share/ShareExtensionHelper.swift
+++ b/Client/Frontend/Share/ShareExtensionHelper.swift
@@ -14,17 +14,6 @@ class ShareExtensionHelper: NSObject {
         case password
         case iBooks
         case `default`
-        
-        var description: String? {
-            switch self {
-            case .password:
-                return "All Password Activity Types pwsafe / 1Password"
-            case .iBooks:
-                return "PDF Sharing Activity Type"
-            case .default:
-                return "Other Types"
-            }
-        }
     }
     
     fileprivate weak var selectedTab: Tab?
--- a/Client/Frontend/Share/ShareExtensionHelper.swift
+++ b/Client/Frontend/Share/ShareExtensionHelper.swift
@@ -9,6 +9,24 @@ import OnePasswordExtension
 private let log = Logger.browserLogger
 
 class ShareExtensionHelper: NSObject {
+    
+    enum ShareActivityType {
+        case password
+        case iBooks
+        case `default`
+        
+        var description: String? {
+            switch self {
+            case .password:
+                return "All Password Activity Types pwsafe / 1Password"
+            case .iBooks:
+                return "PDF Sharing Activity Type"
+            case .default:
+                return "Other Types"
+            }
+        }
+    }
+    
     fileprivate weak var selectedTab: Tab?
 
     fileprivate let selectedURL: URL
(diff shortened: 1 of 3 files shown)


Client/Frontend/Share/ShareExtensionHelper.swift
```diff
@@ -9,6 +9,24 @@ import OnePasswordExtension
 private let log = Logger.browserLogger
 
 class ShareExtensionHelper: NSObject {
+    
+    enum ShareActivityType {
+        case password
+        case iBooks
+        case `default`
+        
+        var description: String? {
```
iccub: What is this `description` used for?

Client/Frontend/Share/ShareExtensionHelper.swift
```diff
@@ -54,27 +74,69 @@ class ShareExtensionHelper: NSObject {
...
+                    completionHandler(completed, activityType, url)
+                }
+            }
         }
+        
         return activityViewController
     }
+    
+    func writeWebPagePDFDataToURL(_ completion: @escaping (URL?, Error?) -> Void) {
+        #if compiler(>=5.3)
```
iccub, line 101: this means save pdf to iBooks works only on iOS14+?

iccub (changes requested):

Client/Frontend/Share/ShareExtensionHelper.swift
soner-yuksel: It shouldnt be but the only solution I managed to find out is to create the pdf and get the link and pass it to books app to open. I might have to come back to this If we re aiming for < iOS 14 solution.

iccub: from what I've read in the corresponding ticket, it works in Firefox, can you check it and see if it works on FF pre iOS14?

soner-yuksel: Ok I misunderstood here, I thought we want to share any website as pdf using iBooks. But actually we want to share pdf file opened in browser in iBooks.

- Sharing any website as pdf using iBooks also fails in Firefox however sharing pdf file opened in browser doesnt fail 

I will look into 2nd pdf file sharing in detail and add it to this PR.

Client/Frontend/Share/ShareExtensionHelper.swift
soner-yuksel: I was using it for some debuggin but it is not needed, removing.

soner-yuksel:
Converting this PR into Draft because the spec of the problem got bigger after the investigation of the problem. And realized the problem is not with iBooks only.

soner-yuksel:
Closing this PR to open a new one with additional fixes

PR #3135, opened 2020-12-10
Add initial guardian alerts to StatsSectionProvider

lechium:
## Summary of Changes

Updated GRDGatewayAPI to have the proper function to get the alert totals having it return AND save to defaults (i'm not great with swift) these defaults are added to BraveVPN as properties that return the correlating user default OR "-" as the case may be. StatsSectionProvider.swift was updated to have a new row, (this row needs to be gated properly within the Guardian subscription parameters) the colors arent themeable and the strings arent localized, aside from that there should be very little work needed to round out this implementation.

## Submitter Checklist:

## Screenshots:

IMG_9751

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


iccub (changes requested):
Thanks for this PR, the Guardian api integration would be helpful.
We want to have a different UI for it though

iccub:
Closing it for now, we are waiting on final designs still, but this will he helpful as a reference when the work is going to get started

PR #3160, opened 2020-12-22
[WIP] Fix#936: Add Support for OpenSearch - Auto/ Manual

### LINKED ISSUE #936: Add support for OpenSearch
Users should be able to add websites as search engines if they support the [OpenSearch standard](https://developer.mozilla.org/en-US/docs/Web/OpenSearch). This would allow users to add custom search engines, which is helpful now that many search engines have been removed from settings.

This feature is already supported by Brave Core (desktop).

For example, GitHub supports OpenSearch. Every page on GitHub has a `<link rel="search"` tag in the `<head>` of its HTML:
```HTML
<link rel="search" type="application/opensearchdescription+xml" href="/opensearch.xml" title="GitHub">
```

The linked XML file (https://github.com/opensearch.xml) contains the URL format for search terms:
```XML
<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/" xmlns:moz="http://www.mozilla.org/2006/browser/search/">
<ShortName>GitHub</ShortName>
<Description>Search GitHub</Description>
<InputEncoding>UTF-8</InputEncoding>
<Image width="16" height="16" type="image/x-icon">https://github.com/favicon.ico</Image>
<Url type="text/html" method="get" template="https://github.com/search?q={searchTerms}&ref=opensearch"/>
<moz:SearchForm>https://github.com/search</moz:SearchForm>
</OpenSearchDescription>
```

soner-yuksel:
This PR is implementing adding Search Engine that has [OpenSearch-Compliance](https://developer.mozilla.org/en-US/docs/Web/OpenSearch).

This is done by 

- Evaluating Open Search engine script and finding out If the website supports Open Search.
- Parse the response and retrieve the title and reference URL
- Create the URL to download the XML + also retrieve the Fav Icon
- And create Search Engine object with the data and add it to search engine list

It also includes the UX elements like add button addition on the toolbar and dialogs for approval and error handling situations

WIP: Open Search Manual Addition / Tests

## Summary of Changes

This pull request fixes #936

## Submitter Checklist:

## Test Plan:

- Go to a website that coplies to OpenSearch Compliance - Search Engines [Ecosia](https://www.ecosia.org/?c=en)
- Press Search Textfield and observe Add Button on toolbar
- Press the add button and accept adding the search engine
- Observe Engine is added in settings or while populating suggestions

## Screenshots:

ezgif-6-92ad65bbb277

IMG936

## Reviewer Checklist:
  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/OpenSearch/BrowserViewController+OpenSearch.swift
+++ b/Client/Frontend/Browser/BrowserViewController/OpenSearch/BrowserViewController+OpenSearch.swift
@@ -148,8 +148,8 @@ extension BrowserViewController {
             }
             
             NetworkManager().downloadResource(with: url).uponQueue(.main) { [weak self] response in
-                guard let openSearchEngine =
-                        OpenSearchParser(pluginMode: true).parse(response.data, referenceURL: referenceURL, image: searchEngineIcon) else {
+                guard let openSearchEngine = OpenSearchParser(pluginMode: true).parse(
+                        response.data, referenceURL: referenceURL, image: searchEngineIcon, isCustomEngine: true) else {
                     return
                 }
                 
--- a/Client/Frontend/Browser/BrowserViewController/OpenSearch/BrowserViewController+OpenSearch.swift
+++ b/Client/Frontend/Browser/BrowserViewController/OpenSearch/BrowserViewController+OpenSearch.swift
@@ -62,13 +62,10 @@ extension BrowserViewController {
         
         let matches = self.profile.searchEngines.orderedEngines.filter {$0.referenceURL == referenceObject.reference}
         
-        // TODO: Change the button !!!!!!!!!!!!!!! Enable Disable
         if !matches.isEmpty {
-            customSearchEngineButton.tintColor = .gray
-            customSearchEngineButton.isUserInteractionEnabled = false
+            self.customSearchEngineButton.state = .disabled
         } else {
-            customSearchEngineButton.tintColor = .blue
-            customSearchEngineButton.isUserInteractionEnabled = true
+            self.customSearchEngineButton.state = .enabled
         }
         
         /*
@@ -133,13 +130,15 @@ extension BrowserViewController {
     }
 
     func downloadOpenSearchXML(_ url: URL, referenceURL: String, title: String, iconURL: URL) {
-        
-        // TODO: Change the button !!!!!!!!!!!!!!! Loading
-        
+        customSearchEngineButton.state = .loading
+
         WebImageCacheManager.shared.load(from: iconURL, completion: { (image, _, _, _, _) in
             guard let image = image else {
                 let alert = ThirdPartySearchAlerts.failedToAddThirdPartySearch()
-                self.present(alert, animated: true, completion: nil)
+                self.present(alert, animated: true) {
+                    self.customSearchEngineButton.state = .enabled
+                }
+                
                 return
             }
             
@@ -162,12 +161,11 @@ extension BrowserViewController {
                 let toast = SimpleToast()
                 toast.showAlertWithText(Strings.thirdPartySearchEngineAdded, bottomContainer: self.webViewContainer)
                 
-                // TODO: Change the button !!!!!!!!!!!!!!! Enabled
-
+                self.customSearchEngineButton.state = .disabled
             } catch {
                 let alert = ThirdPartySearchAlerts.failedToAddThirdPartySearch()
                 self.present(alert, animated: true) {
-                    // TODO: Change the button !!!!!!!!!!!!!!! Disabled
+                    self.customSearchEngineButton.state = .enabled
                 }
             }
         }
--- a/Client/Frontend/Browser/BrowserViewController/OpenSearch/BrowserViewController+OpenSearch.swift
+++ b/Client/Frontend/Browser/BrowserViewController/OpenSearch/BrowserViewController+OpenSearch.swift
@@ -143,7 +143,7 @@ extension BrowserViewController {
                 return
             }
             
-            NetworkManager().downloadResource(with: url).upon() { [weak self] response in
+            NetworkManager().downloadResource(with: url).uponQueue(.main) { [weak self] response in
                 guard let openSearchEngine =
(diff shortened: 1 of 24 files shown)


Client/Frontend/Browser/BrowserViewController/OpenSearch/BrowserViewController+OpenSearch.swift
```diff
@@ -0,0 +1,212 @@
...
+    /// Adding Toolbar button over the keyboard for adding Open Search Engine
+    /// - Parameter webView: webview triggered open seach engine
+    func evaluateWebsiteSupportOpenSearchEngine(_ webView: WKWebView) {
+        let script = """
+                        var link = document.querySelector("link[type='application/opensearchdescription+xml']")
+                        var dict = {
+                            href : link.getAttribute("href"),
+                            title : link.getAttribute("title")
+                        };
+                        JSON.stringify(dict)
```
Brandon-T, line 39: Is this valid JS?

`JSON.stringify(dict)` isn't returning a value from a function or anything.
IE: 

`var link = document.querySelector(.....)`
`var dict = {....}`
`JSON.stringify(dict)` but not assigned to anything or returned anywhere?

Client/Frontend/Browser/BrowserViewController/OpenSearch/BrowserViewController+OpenSearch.swift
```diff
@@ -0,0 +1,212 @@
...
+         */
+        guard let webContentView = UIView.findSubViewWithFirstResponder(webView) else {
+            /*
+             In some cases the URL bar can trigger the keyboard notification. In that case the webview isnt the first responder
+             and a search button should not be added.
+             */
+            return
+        }
+        
+        guard let input = webContentView.perform(#selector(getter: UIResponder.inputAccessoryView)),
```
Brandon-T, line 91: You should definitely check if that selector exists first before executing it because if this changes on a different version of iOS, we'll get a crash. Happened to me with videos in WKWebView.

Client/Frontend/Browser/BrowserViewController/OpenSearch/BrowserViewController+OpenSearch.swift
```diff
@@ -0,0 +1,212 @@
...
+            return
+        }
+                
+        guard let scheme = tabManager.selectedTab?.webView?.url?.scheme,
+              let host = tabManager.selectedTab?.webView?.url?.host else {
+            log.error("Selected Tab doesn't have URL")
+            return
+        }
+        
+        while referenceURLString.hasPrefix("/") {
```
Brandon-T, line 125: Why do this? If you're going to reconstruct it from `scheme://host/path` you can probably get away with using `referenceURL.path` and reconstruct it or use `URLComponent`?

Client/Frontend/Browser/BrowserViewController/OpenSearch/BrowserViewController+OpenSearch.swift
```diff
@@ -0,0 +1,212 @@
...
+            buttonGroup.barButtonItems = buttonGroup.barButtonItems.filter { $0 != customSearchBarButton }
+            customSearchBarButton = nil
+        }
+
+        if self.customSearchEngineButton.superview != nil {
+            self.customSearchEngineButton.removeFromSuperview()
+        }
+
+        UIView.animate(withDuration: state.animationDuration) {
+            UIView.setAnimationCurve(state.animationCurve)
```
Brandon-T, line 213: Instead fo `setAnimationCurve` will `animate(withDuration:delay:options:animations:completion:)` work if you supply the `curve` as `options` parameter?

Brandon-T (changes requested):

Client/Frontend/Browser/BrowserViewController/OpenSearch/BrowserViewController+OpenSearch.swift
soner-yuksel: Yes ofcourse it is valid and it works. 
In webview we are evaluating the result of [opensearchdecription](https://developer.mozilla.org/en-US/docs/Web/OpenSearch) and getting the reference link and title from the json object and use it while saving the search engine.

Client/Frontend/Browser/BrowserViewController/OpenSearch/BrowserViewController+OpenSearch.swift
soner-yuksel: Yeah this part was from exactly how Firefox used to add a toolbar button on top of keyboard.
I will add some scrambling to the value for the key but as far as I know we cant check responds to selector

Client/Frontend/Browser/BrowserViewController/OpenSearch/BrowserViewController+OpenSearch.swift
soner-yuksel: Honestly the problem is there is no convention about what we might receive as reference.
Sometimes you need to construct and sometimes you can fetch string starts // or / and different types.

Here are some of the examples:
```
href: /w/opensearch_desc.php
href: https://cdn.ecosia.org/opensearch.xml
href: lilosearch.xml
href: /opensearch.xml
href: //opensearch_desc.xml
```

Client/Frontend/Browser/BrowserViewController/OpenSearch/BrowserViewController+OpenSearch.swift
soner-yuksel: It can work yes but in that case you will need to use a convenience method to get the animation option according to which type of curve is used. 

```
func animationOption(for curve: UIViewAnimationCurve) -> UIViewAnimationOptions {
	switch (curve) {
	case .easeInOut:
		return .curveEaseInOut
	case .easeIn:
		return .curveEaseIn
	case .easeOut:
		return .curveEaseOut
	case .linear:
		return .curveLinear
	}
}
```

So I believe their way for handling this is easier

soner-yuksel:
Closing PR because Auto and manual search is combined under https://github.com/brave/brave-ios/pull/3160

soner-yuksel:
Closing this Draft PR To open a new one with changes in Settings for Engines

PR #3189, opened 2021-01-12
Fix #3177: Update pre-populated search engine list

### LINKED ISSUE #3177: Updated pre-populated search engine list
Follow desktop https://github.com/brave/brave-browser/issues/13283

iccub:
This also replaces Yahoo with Ecosia on search onboarding screen.

## Summary of Changes

This pull request fixes #3177 

## Submitter Checklist:

## Test Plan:

1. New user
- Set region with Ecosia enabled, United States for example
- Fresh install the app
- Verify that Ecosia is one of search options showed during onboarding
- Verify that Yahoo is not listed as one of search options
- Select it, make sure search queries use this search engine
- Verify that correct search parameters are passed, you can refer to https://github.com/brave/brave-browser/wiki/Search-engine-integrations

2. Existing user
- Install previous app build, select some search engine
- Update to build with Ecosia, make sure eligible region(like US) is set
- Go to Brave Settings-> search engines
- Verify that Ecosia is on the list, select it
- Verify that correct search parameters are passed

3. Existing Yahoo user
- Install previous app build, select Yahoo as default search engine
- Update to build with Ecosia, make sure eligible region(like US) is set
- Verify that default search is still Yahoo

## Screenshots:

Zrzut ekranu 2021-01-12 o 15 21 20

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


bsclifton (approved):
Changes look great to me! We'll keep Yahoo as an available engine (not in onboarding of course) until we can support having it migrated off (maybe after merging OpenSearch code)

iccub:
@bsclifton  yes, i tried to move yahoo out, but found no simple way to do it, we will revisit it later

kylehickinson (approved):

Client/Frontend/Browser/Search/InitialSearchEngines.swift
```diff
@@ -119,16 +134,20 @@ class InitialSearchEngines {
...
+        }
         
         if Self.yahooEligibleRegions.contains(region) {
-            priorityEngine = .yahoo
+            replaceOrInsert(engineId: .yahoo, customId: nil)
         }
     }
     
+    private func priorityOverrides() {
+        // No priority engines are live at the moment.
```
soner-yuksel, line 148: Asking it for context, when there is no other, our default priority engine is google in this case isnt it ? I think it got complicated with default engine and priority engine at the same time.

Client/Frontend/Browser/Search/InitialSearchEngines.swift
```diff
@@ -137,7 +156,7 @@ class InitialSearchEngines {
             .sorted { e, _ in e.id == priorityEngine }
     }
     
-    private func replaceOrInsert(engineId: SearchEngineID, customId: String) {
+    private func replaceOrInsert(engineId: SearchEngineID, customId: String?) {
         guard let engineIndex = engines.firstIndex(of: .init(id: engineId)) else {
```
soner-yuksel: I believe no need to create an instance here, we can probably use 
`        guard let engineIndex = engines.firstIndex(where: { $0.id == engineId }) else {
`

Client/Frontend/Browser/Search/SearchEngines.swift
```diff
@@ -250,7 +250,8 @@ class SearchEngines {
 
     /// Get all bundled (not custom) search engines, with the default search engine first,
     /// but the others in no particular order.
-    class func getUnorderedBundledEnginesFor(selected: [String] = []) -> [OpenSearchEngine] {
+    class func getUnorderedBundledEngines(selected: [String] = [],
```
soner-yuksel: I think function can be renamed like
`class func getUnorderedBundledEngines(for selectedEngines: [String] = [],  isOnboarding: Bool) `

Client/Frontend/Browser/Search/InitialSearchEngines.swift
```diff
@@ -10,7 +10,16 @@ import Foundation
 class InitialSearchEngines {
     /// Type of search engine available to the user.
     enum SearchEngineID: String {
-        case google, bing, duckduckgo, yandex, qwant, startpage, yahoo
+        case google, bing, duckduckgo, yandex, qwant, startpage, yahoo, ecosia
+        
+        var excludedFromOnboarding: Bool {
```
soner-yuksel, line 15: This property is temporary right until we remove the yahoo engine totally ?

soner-yuksel (approved):

Client/Frontend/Browser/Search/InitialSearchEngines.swift
iccub: Priority engine is a special case, it is not set as default, but on onboarding it is shown at the top of the list.
For user added engines you work on there is no priority-engine logic, it's for predefined engines only

So full list goes like

1. Priority engine
2. Default engine(preselected)
3. other engine
4. other engine
..

Client/Frontend/Browser/Search/InitialSearchEngines.swift
iccub: I think we can keep it, if we add a lot of predefined search engines in the future we can choose which of them to show during onboarding, to avoid having a long list of search engines

soner-yuksel (approved):

PR #3548, opened 2021-04-19
Fix #3532 - Update playlist toast gradient

### LINKED ISSUE #3532: Playlist: Add to Brave Playlist toast design change
Feedback has suggested that the "Add to Playlist" toast may not feel like part of the Brave UI. One way to get closer to Brave UI is to update the color of the toast.

Keeping existing toast design. Change the current blue background to a Brave brand gradient.

`linear-gradient(304.74deg, #6F4CD2 15.81%, #BF14A2 63.17%, #F73A1C 100%);`

Brandon-T:
## Summary of Changes
- Adds a linear gradient angled at X degrees to the toast, clamped to the range [0, 1] (unit circle) and extrapolated by iOS internals to [0, 1] (unit square).

This pull request fixes #3532

## Submitter Checklist:

## Screenshots:

Zrzut ekranu 2021-04-19 o 23 26 27

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Brandon-T:
In case you guys want to see how the circle is graphed: https://www.desmos.com/calculator/h69qnkg8q7

Clamped with the range [0, 1] the **maximum radius** of such a circle that fits in such a square (bounds/rect) is **0.5**.
Because diameter (length and width) of the unit square/circle is **1** so radius is **0.5**

When solving the formula while keeping the constraints that two opposite points lie on such a circle, that's how we get the angle between them. Doing the reverse calculation gives us both startPoint and endPoint.. where they are opposite of each other. Angle is in radians.

**The defaults for iOS:**

StartPoint: (0.5,0.0)
EndPoint: (0.5, 1.0)

This is why we shift it by +0.5 and compress it by 0.5x to keep it to a minimum of [0, 1] so it never goes into the negative range.

soner-yuksel (approved):

Client/Frontend/Browser/Playlist/PlaylistToast.swift
```diff
@@ -146,6 +146,16 @@ class PlaylistToast: Toast {
             $0.contentEdgeInsets = UIEdgeInsets(top: 10.0, left: 10.0, bottom: 10.0, right: 20.0)
             $0.titleEdgeInsets = UIEdgeInsets(top: 0.0, left: 10.0, bottom: 0.0, right: -10.0)
             
+            let colors = [#colorLiteral(red: 0.4352941176, green: 0.2980392157, blue: 0.8235294118, alpha: 1), #colorLiteral(red: 0.7490196078, green: 0.07843137255, blue: 0.6352941176, alpha: 1), #colorLiteral(red: 0.968627451, green: 0.2274509804, blue: 0.1098039216, alpha: 1)]
+            let gradientView = GradientView(colors: colors,
+                                            positions: [0.1581, 0.6317, 1.0],
+                                            angle: 304.74)
```
iccub: Can you export this value to new file called `Gradient.swift` in `BraveUI` framework?
This may require moving `GradientView` there too

The gradient should be named `gradient02-alt-light`  or something similar

iccub (changes requested):
Add gradient to two new toasts we added for playlists

Client/Frontend/Browser/Playlist/PlaylistToast.swift
Brandon-T: Fixed.

iccub (approved):
Looks great

PR #4043, opened 2021-08-13
Fix #3814: Ads reporting for SPAs and redirection requests.

### LINKED ISSUE #3814: Conversion are not counted on Single Page Applications
### Description: 
Brave Ads does not count conversions when the Conversion URL is a single-page application. We need to access the redirect chain to ensure that we measure conversion in iOS in the same way we do for desktop/android.

### Steps to Reproduce 
  1. Open Brave for iOS
  2. Adjust creatives in db to use only `Womplay` ads
  3. Trigger ad
  4. either click ad or navigate to URL `https://womplay.io/?utm_source=brave&utm_campaign=USA-UK&utm_content=You%20Play%20%F0%9F%8E%AE%20We%20Pay%20%F0%9F%92%B0%20Womplay`
  5. complete registration process
  6. confirm that you reached `https://womplay.io/register/ready`, thereby satisfying the conversion URL pattern `https://womplay.io/register/ready*`
  7. confirm if conversion has been counted.

**Actual result:** 
Conversion are not counted

**Expected result:**
Conversion are counted even when the conversion URL is on an SPA

Relatively easily reproduced. For Womplay, you must use a net new email/account to register otherwise you will be put in to log-in path and not be redirected to the conversion url page

**Brave Version:** 
Release Channel 1.26

**Device details:** 
I believe this to be all iOS ver

### Additional Information
* you must use a net new email/account to register otherwise you will be put in to log-in path and not be redirected to the conversion url page

cc: @tmancey

iccub:
## Summary of Changes

This pull request fixes #3814 

## Submitter Checklist:

## Test Plan:

- Confirm when loading a webpage the most recent url in the redirect change is processed for purchase intent
- Confirm when loading a webpage the page content is classified (including all pages for single page applications)
- Confirm ads transfer (land) after clicking an ad and waiting for ~10 seconds (match the URL against the original URL, i.e. first URL in the redirect chain)
- Confirm ads conversions work as expected (matching the pattern against all URLs in the redirect chain)

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1341,7 +1341,7 @@ class BrowserViewController: UIViewController {
                 // Notify Rewards of new page load.
                 if let rewardsURL = rewardsXHRLoadURL,
                     url.host == rewardsURL.host {
-                    tabManager.selectedTab?.reportPageNaviagtion(to: rewards)
+                    tabManager.selectedTab?.reportPageNavigation(to: rewards)
                     // Not passing redirection chain here, in page navigation should not use them.
                     tabManager.selectedTab?.reportPageLoad(to: rewards, redirectionURLs: [])
                 }
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1270,6 +1270,10 @@ class BrowserViewController: UIViewController {
         }
         return false
     }
+    
+    // This variable is used to keep track of current page. It is used to detect internal site navigation
+    // to report internal page load to Rewards lib
+    var rewardsXHRLoadURL: URL?
 
     override func observeValue(forKeyPath keyPath: String?, of object: Any?, change: [NSKeyValueChangeKey: Any]?, context: UnsafeMutableRawPointer?) {
         
@@ -1330,6 +1334,17 @@ class BrowserViewController: UIViewController {
                 // Catch history pushState navigation, but ONLY for same origin navigation,
                 // for reasons above about URL spoofing risk.
                 navigateInTab(tab: tab)
+            }
+            
+            // Rewards reporting
+            if let url = change?[.newKey] as? URL, !url.isLocal {
+                // Notify Rewards of new page load.
+                if let rewardsURL = rewardsXHRLoadURL,
+                    url.host == rewardsURL.host {
+                    tabManager.selectedTab?.reportPageNaviagtion(to: rewards)
+                    // Not passing redirection chain here, in page navigation should not use them.
+                    tabManager.selectedTab?.reportPageLoad(to: rewards, redirectionURLs: [])
+                }
             }
         case .title:
             // Ensure that the tab title *actually* changed to prevent repeated calls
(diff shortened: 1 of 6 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1331,6 +1335,17 @@ class BrowserViewController: UIViewController {
                 // for reasons above about URL spoofing risk.
                 navigateInTab(tab: tab)
             }
+            
+            // Rewards reporting
```
iccub, line 1339: I moved it directly under URL KVO and removed `mediaSite` path, now we report page load on all internal page navigations, previously it was only for youtube, twitch etc..

This was also moved because `updateURLBar` is called from few places, and we want to avoid duplicate calls for `reportPageLoad`

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1331,6 +1335,17 @@ class BrowserViewController: UIViewController {
...
+            if let url = change?[.newKey] as? URL, !url.isLocal {
```
iccub, line 1340: `tab.url` was often reporting previous url instead of new value, changed to use the kvo's `newKey`

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -475,4 +476,9 @@ extension BrowserViewController: WKNavigationDelegate {
             tabsBar.reloadDataAndRestoreSelectedTab()
         }
     }
+    
+    func webView(_ webView: WKWebView, didReceiveServerRedirectForProvisionalNavigation navigation: WKNavigation!) {
+        guard let tab = tabManager[webView], let url = webView.url, rewards.isEnabled else { return }
+        tab.redirectURLs.append(url)
```
iccub, line 482: Redirects are called before navigation `didFinish`, we catch all redirection here, then pass them in didFinish and clear the redirection url array

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1331,6 +1335,17 @@ class BrowserViewController: UIViewController {
...
+                // Notify Rewards of new page load.
+                if let rewardsURL = rewardsXHRLoadURL,
+                    url.host == rewardsURL.host {
+                    tabManager.selectedTab?.reportPageNaviagtion(to: rewards)
+                    // Not passing redirection chain here, in page navigation should not use them.
+                    tabManager.selectedTab?.reportPageLoad(to: rewards, redirectionURLs: [])
```
iccub, line 1346: URL kvo gets called before `didFinish` or `didReceiveServerRedirectForProvisionalNavigation`, 
it does not have access to redirection urls

tmancey: `reportPageNaviagtion` spelling error. Should be `reportPageNavigation`

tmancey (approved):
LGTM++

kylehickinson (approved):
Thanks for picking this up :)

soner-yuksel (approved):

PR #4086, opened 2021-08-25
Fix #4087 - Created a proxy object to hide JS properties.

### LINKED ISSUE #4087: JS leaks private properties via window.__firefox__
### Description: 
- Going to any page and typing `window.__firefox__` will leak private properties so that if someone does `Object.keys(window.__firefox__)` it will NOT show the properties, but enumerating it with `Object.getOwnPropertyNames(window.__firefox__)` WILL display a list of all properties and they can be enumerated & changed by malicious JS.

**Actual result:** 
Properties are enumerable and they leak to the page.

**Expected result:**
Should not leak properties. They should NOT be enumerable.

Easily reproduced

**Brave Version:** 
Any

Brandon-T:
Properties can only be hidden if they are non-configurable & writable.
If a property is non-configurable, readonly, it is impossible to hide as a TypeError will be thrown.

## Summary of Changes
- Hides JS properties like those added `window.__firefox__`.
- The properties can only be retrieved in code if you know the EXACT name of it. Since we use the `securityToken` everywhere (hopefully), all properties should be hidden and inaccessible (aka WKContentWorld for below iOS 14), but also in cases where the page must have access to the JS where ContentWorld sandbox can't be used, this is the alternative :)

This pull request fixes #4087

## Submitter Checklist:

## Test Plan:
N/A

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Brandon-T:
Without this: 
image

With this:
image

By returning `[]` from `ownKeys` function, we get:
image

Making it impossible to enumerate the properties. Alternatively, instead of returning `[]` we can just make all the properties `configurable: true` OR `enumerable: true` and they will be hidden automatically just like `userScripts` and `BSHE7BF....` (BraveSearch) as shown in the first screenshot. Doing so will result in `Object.getOwnPropertyNames(window.__firefox__)` returning `[]`

iccub (approved):
Needs a test plan, we can sanity check that adding items to playlist works, does it affect other js too? like reader mode

Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentEnd/__firefox__.js
```diff
@@ -10,19 +10,82 @@ if (!window.__firefox__) {
...
-      userScripts: {},
-      includeOnce: function(userScript, initializer) {
-        if (!__firefox__.userScripts[userScript]) {
-          __firefox__.userScripts[userScript] = true;
-          if (typeof initializer === 'function') {
-            initializer();
-          }
-          return false;
-        }
+    value: (function() {
```
jumde, line 13: nit: I'm not sure I understand why we need `__firefox__.js` added at both `AtDocumentStart` and `AtDocumentEnd` but we can file a follow-up issue to figure this out.

jumde (approved):
lgtm

PR #4242, opened 2021-09-28
Fix #4241: Resolve Privileged Request against BaseURL.

### LINKED ISSUE #4241: Going back from a Brave News article is not possible anymore.
### Steps to Reproduce 
  1. Go to NTP, tap on a news article
  2. Try to go back to ntp

**Actual result:** 
You can't

**Expected result:**

**Brave Version:** 

**Device details:** 
iOS 15

**Website problems only:**

Brandon-T:
## Summary of Changes

- Local host URLs like `about:home` require resolving against baseURL. Not exactly sure why `http://localhost:6596/about/home` requires this, but it does because the `URLComponents` class will strip the host. It does not do this for any other URLs which is why sharing URLs with Brave works fine.

This pull request fixes #4241

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Shared/PrivilegedRequest.swift
+++ b/Shared/PrivilegedRequest.swift
@@ -38,7 +38,7 @@ public class PrivilegedRequest: NSMutableURLRequest {
     }
 
     private static func store(url: URL) -> URL? {
-        guard var components = URLComponents(url: url, resolvingAgainstBaseURL: false) else { return nil }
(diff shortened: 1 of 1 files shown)


iccub (approved):

iccub (changes requested):
does not seem to work for me

iccub (approved):
works :)

PR #4250, opened 2021-09-29
Fix #4249 - Certificate Pinning Robustness

### LINKED ISSUE #4249: DST Root CA X3 (Let's Encrypt Certificate) is Expired as of September 30th, 2021
### Description: 
- DST Root CA X3 (Let's Encrypt Certificate) is Expired as of September 30th, 2021
- https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/
- We need to remove the expired certificate and also make it more robust to detect expired certificates while developing the app.

### Steps to Reproduce 
  1. N/A

Brandon-T:
## Summary of Changes
- Security Review: https://github.com/brave/security/issues/599
- Update the Unit test certificates to ones that aren't expired.
- Removed expired DST Root CA X3 certificate from the iOS app.
- Added `AssertionFailure` to detect when pinning fails (instead of only relying on unit tests) while **developing** the app.
- Renamed the `self-signed` anchoring option and added another option to validate certificate trust chain + validating against the system trusts by default.
- Note: iOS 12 added `SecTrustSetSignedCertificateTimestamps` but we don't have a CT log to validate against, so not added. https://developer.apple.com/documentation/security/3081274-sectrustsetsignedcertificatetime

This pull request fixes #4249

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


soner-yuksel (approved):

BraveShared/CertificatePinning.swift
```diff
@@ -86,14 +89,27 @@ public class PinningCertificateEvaluator: NSObject, URLSessionDelegate {
...
-            guard SecTrustSetAnchorCertificatesOnly(trust, true) == errSecSuccess else {
-                throw error(reason: "Self Signed Certificate Anchor Only Failed")
+            if options.contains(.anchorSpecificTrustsOnly) {
+                // Trust only the passed in certificates (true)
+                //
+                // This is the default behaviour, however we do it explicitly to throw an exception
+                // immediately upon failure
+                // The default behaviour will silently ignore the exception until validation
+                guard SecTrustSetAnchorCertificatesOnly(trust, true) == errSecSuccess else {
+                    throw error(reason: "Self-Signed Certificate Anchor Only Failed")
```
fmarier, line 106: I'm confused about the "self-signed" bit here. Which certs are self-signed? The pinned root CAs?

BraveShared/CertificatePinning.swift
```diff
@@ -141,16 +157,33 @@ public class PinningCertificateEvaluator: NSObject, URLSessionDelegate {
...
-        public static let all: PinningOptions = [.default, .validateHost, .allowSelfSigned]
+        
+        /// Anchor ONLY the specified trusts passed to `PinningCertificateEvaluator.evaluate(trust:host:)`
+        public static let anchorSpecificTrustsOnly = PinningOptions(rawValue: 1 << 2)
+        
+        /// Anchor the specified trusts passed to `PinningCertificateEvaluator.evaluate(trust:host:)`
+        /// Also anchors the default trusts of the System
+        public static let anchorSpecificAndSystemTrusts = PinningOptions(rawValue: 1 << 3)
+        
+        /// Default pinning policies + Host Validation + Anchor the specified trusts + Anchor system trusts
```
fmarier, line 186: Are the default pinning policies the ones that Apple ships?

BraveSharedTests/CertificatePinningTest.swift
```diff
@@ -109,7 +109,7 @@ class CertificatePinningTest: XCTestCase {
 
         let host = "unit-test.brave.com"
```
fmarier, line 110: I'm not sure what this one is testing because that hostname doesn't resolve.

If you want to test a correctly-pinned site, you can use `pcdn.brave.com`.

If you want to test an incorrectly-pinned site, you can use `ssl-pinning.someblog.org` (also controlled by us).

fmarier (commented):
One more thing you might want to roll into this PR is the new pinning test hostname I just added:
https://github.com/brave/brave-core/blob/ec6f175ad9667fe0658c71c41ca4fac08aa31727/chromium_src/net/tools/transport_security_state_generator/input_file_parsers.cc#L473-L474

BraveShared/CertificatePinning.swift
Brandon-T: Only the ones in the unit test is self-signed. 

Self-Signed Certificate:
image

BraveShared/CertificatePinning.swift
Brandon-T: Yes. The default certificates are the ones from the system. IE: "Built-In anchor"
image

BraveSharedTests/CertificatePinningTest.swift
Brandon-T: It tests that our certificate-pinning can work on self-signed certificates if we allow it via the options enum.
So it is testing that anchoring works.

PR #4397, opened 2021-10-22
Fix #4396: Remove MediaPlaybackState API (iOS 15.0+)

### LINKED ISSUE #4396: [hackerone] #1376708 - Address bar spoofing
https://hackerone.com/reports/1376708

Brandon-T:
## Summary of Changes
- First, it is worth noting that the URL bar isn't actually spoofed at all (at least not that I can see after hours of testing). The bug seems to be that the page does NOT finish loading and thus displays the content from the parent page. The URL bar will still show `about:blank` and not `badssl.com` when attempting to reproduce the issue, so I don't see any URL bar spoofing happening. However, the page MUST continue to load and behave like Safari and other browsers.

- Second, this only happens on iOS 15.0+

- Remove the tab/webView from the view hierarchy and get rid of the iOS 15.0+ (aka old 14.5+) API for media playback state (Might break Brave-Talk backgrounding).

- I don't see another fix for this URL spoofing bug atm. Somehow the API is returning that media is playing even when none is. WebKit still allows a WebView to control the JS and requests of its child WebView.

- We should probably find another way to fix this later on (maybe a follow up ticket).

This pull request fixes #4396

## Submitter Checklist:

## Test Plan:

See https://bravesoftware.slack.com/archives/G2KN13Z8C/p1634946451010500?thread_ts=1634932309.006300&channel=G2KN13Z8C&message_ts=1634946451.010500
for steps

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -2227,21 +2227,7 @@ extension BrowserViewController: TabManagerDelegate {
             wv.accessibilityLabel = nil
             wv.accessibilityElementsHidden = true
             wv.accessibilityIdentifier = nil
-            
-            #if swift(>=5.4)
-            if #available(iOS 15.0, *) {
-                wv.alpha = 0.0
-                wv.requestMediaPlaybackState { state in
-                    if state != .playing && wv != tabManager.selectedTab?.webView {
-                        wv.alpha = 1.0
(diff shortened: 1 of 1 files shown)


iccub (approved):
This means switching tab for youtube and BraveTalk will be broken again right

PR #4559, opened 2021-11-24
Fix #4534: Added Sync QR-Code v2 support

### LINKED ISSUE #4534: Changes to sync QR code
### Description: 
See https://github.com/brave/internal/issues/832

Brandon-T:
## Summary of Changes
- Added Sync QR-Code v2 support
- Security Review: https://github.com/brave/security/issues/675

This pull request fixes #4534

## Submitter Checklist:

## Test Plan:

- Test joining an old sync chain
- Test creating a new sync chain
- Test joining the new sync chain
- Test joining sync chains from different timezones
- Test joining a sync chain with an expired code
- Test joining an old sync chain after January 1st, 2022
- Test joining a new sync chain after January 1st, 2022
- Test joining a sync chain with an expiry date longer than 60 minutes
- Test joining a sync chain with an old version in the QR code (version <= 1)
- Test joining a sync chain with a newer version in the QR code (version >= 3)
- Test joining a sync chain with invalid hex code
- Test joining a sync chain with no expiry date

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Client/Frontend/Sync/SyncPairCameraViewController.swift
```diff
@@ -181,11 +142,112 @@ class SyncPairCameraViewController: SyncViewController {
...
+        AudioServicesPlayAlertSound(SystemSoundID(kSystemSoundID_Vibrate))
+        
+        // Forced timeout
+        let task = DispatchWorkItem { [weak self] in
+            guard let self = self else { return }
+            self.cameraLocked = false
+            self.cameraView.cameraOverlayError()
+        }
+        
+        DispatchQueue.main.asyncAfter(deadline: .now() + 25.0, execute: task)
```
soner-yuksel: Can we add move this 25 expiration to a static value in this class? So it might easier to track later.

Client/Frontend/Sync/SyncPairCameraViewController.swift
```diff
@@ -181,11 +142,112 @@ class SyncPairCameraViewController: SyncViewController {
...
+        // Guard against multi-scanning
+        if cameraLocked { return }
+        cameraLocked = true
+        
+        // Pause scanning
+        cameraView.cameraOverlaySuccess()
+        cameraView.stopRunning()
+        
+        // Vibrate.
+        AudioServicesPlayAlertSound(SystemSoundID(kSystemSoundID_Vibrate))
```
soner-yuksel, line 163: I believe this will play a sound if vibration is off `AudioServicesPlaySystemSound`. Just check if it is what we wanted.

Client/Frontend/Sync/BraveCore/BraveSyncQRCodeModel.swift
```diff
@@ -0,0 +1,142 @@
...
+            } catch {
+                log.error("Error converting QR Code to BraveSyncQRCodeModel: \(error)")
+            }
+        }
+        return nil
+    }
+    
+    func validate(syncAPI: BraveSyncAPI) -> BraveSyncQRCodeError {
+        // Older verison of the QR code (not backwards compatible)
+        if version < BraveSyncQRCodeModel.currentlySupportedVersion {
```
soner-yuksel, line 92: Is this for future proofing, why do we check the version as smaller or bigger. Cant we use it as .notSupported? 
Do we have different error messages for these cases here ?

Client/Frontend/Sync/BraveCore/BraveSyncAPIExtensions.swift
```diff
@@ -158,3 +158,40 @@ extension BraveSyncAPI {
...
+            return nil
+        }
+        
+        guard let filter = CIFilter(name: "CIQRCodeGenerator") else {
+            return nil
+        }
+
+        filter.do {
+            $0.setValue(syncCodeData, forKey: "inputMessage")
+            $0.setValue("H", forKey: "inputCorrectionLevel")
```
soner-yuksel, line 182: Curious why did we wanna use inputCorrectionLevel, it is not a big image we are using there right ?

soner-yuksel (approved):
Just some nits, great job.

Client/Frontend/Sync/BraveCore/BraveSyncAPIExtensions.swift
```diff
@@ -158,3 +158,40 @@ extension BraveSyncAPI {
...
+        }
+        
+        if let image = filter.outputImage {
+            let scaleX = size.width / image.extent.size.width
+            let scaleY = size.height / image.extent.size.height
```
iccub, line 189: Do we care about divide by 0 here?

Client/Frontend/Sync/BraveCore/BraveSyncQRCodeModel.swift
```diff
@@ -0,0 +1,142 @@
...
+        // Currently supported version
+        if notValidAfter < BraveSyncQRCodeModel.braveSyncEpoch {
+            // Impossible to have a QR code of this version
+            // before the brave sync epoch
+            return .invalidFormat
+        }
+        
+        // Date validation
+        let minDate = Date().timeIntervalSince1970 - BraveSyncQRCodeModel.validityDuration
+        let maxDate = Date().timeIntervalSince1970 + BraveSyncQRCodeModel.validityDuration
```
iccub: This model would be nice to be unit tested imo,
if so, we should export this `Date()` calls to a function argument or make some mocks

Client/Frontend/Sync/BraveCore/BraveSyncQRCodeModel.swift
```diff
@@ -0,0 +1,142 @@
...
+            do {
+                return try JSONDecoder().decode(Self.self, from: data)
+            } catch {
+                log.error("Error converting QR Code to BraveSyncQRCodeModel: \(error)")
+            }
+        }
+        return nil
+    }
+    
+    func validate(syncAPI: BraveSyncAPI) -> BraveSyncQRCodeError {
```
iccub, line 90: Plenty of conditions here, would love to see unit tests if not too hard to add

iccub (changes requested):
Looks solid. If we have time I would consider adding some unit tests for `BraveSyncQRCodeModel` if that's not too hard to add.

Client/Frontend/Sync/BraveCore/BraveSyncQRCodeModel.swift
```diff
@@ -0,0 +1,142 @@
...
+    static let braveSyncEpoch = 1635811200
+    
+    /// TimeInterval of Brave Sync Sunset
+    /// 2022-01-01T00:00:00+0000 [yyyy-MM-dd'T'hh:mm:ssZ]
+    static let sunsetDate = 1640995200
+    
+    init(syncHexCode: String) {
+        self.version = BraveSyncQRCodeModel.currentlySupportedVersion
+        self.syncHexCode = syncHexCode
+        self.notValidAfter = Int64(Date().addingTimeInterval(BraveSyncQRCodeModel.validityDuration).timeIntervalSince1970)
```
fmarier, line 48: The generated code should only be valid for 30 minutes.

Client/Frontend/Sync/BraveCore/BraveSyncQRCodeModel.swift
```diff
@@ -0,0 +1,142 @@
...
+            // before the brave sync epoch
+            return .invalidFormat
+        }
+        
+        // Date validation
+        let minDate = Date().timeIntervalSince1970 - BraveSyncQRCodeModel.validityDuration
+        let maxDate = Date().timeIntervalSince1970 + BraveSyncQRCodeModel.validityDuration
+        
+        // Currently Expired (QR code date is too far into the past)
+        if TimeInterval(notValidAfter) <= minDate {
```
fmarier, line 113: This should be `<= now` since the 30-min buffer is added at code generation time.

Client/Frontend/Sync/BraveCore/BraveSyncQRCodeModel.swift
```diff
@@ -0,0 +1,142 @@
...
+        
+        // Currently supported version
+        if notValidAfter < BraveSyncQRCodeModel.braveSyncEpoch {
+            // Impossible to have a QR code of this version
+            // before the brave sync epoch
+            return .invalidFormat
+        }
+        
+        // Date validation
+        let minDate = Date().timeIntervalSince1970 - BraveSyncQRCodeModel.validityDuration
```
fmarier: This should be just `Date().timeIntervalSince1970` since the 30-min buffer is added at code generation time.

Client/Frontend/Sync/BraveCore/BraveSyncQRCodeModel.swift
```diff
@@ -0,0 +1,142 @@
...
+        }
+        
+        // Invalid sync code validation
+        let codeWords = syncAPI.syncCode(fromHexSeed: syncHexCode)
+        if codeWords.isEmpty {
+            return .invalidFormat
+        }
+        
+        let codeWordsCount = codeWords.separatedBy(" ").count
+        if codeWordsCount < 24 || codeWordsCount > 25 {
```
fmarier: This should always be 24. The change for 25 words are no longer in-scope for Phase 1.

fmarier (changes requested):

Client/Frontend/Sync/BraveCore/BraveSyncAPIExtensions.swift
Brandon-T: We want the highest level of error correction in case the QR code is corrupted or can't be read properly.
https://developer.apple.com/documentation/coreimage/ciqrcodedescriptor/errorcorrectionlevel/levelh

Client/Frontend/Sync/BraveCore/BraveSyncQRCodeModel.swift
Brandon-T: It is. We don't want to support any QR-Codes less than the current version "2". So if the version is "1", we want an error (For example, QR code version is insecure or super old).

If the version is > the current supported version, we want a different error (For example, user might need to update their app to scan a newer QR code version).

Client/Frontend/Sync/SyncPairCameraViewController.swift
Brandon-T: This is the original code that's always been there, I just moved it :)

Client/Frontend/Sync/BraveCore/BraveSyncAPIExtensions.swift
Brandon-T: We don't because it should never generate an image of size zero. However, I added a check anyway.

Client/Frontend/Sync/BraveCore/BraveSyncQRCodeModel.swift
Brandon-T: Fixed

Client/Frontend/Sync/BraveCore/BraveSyncQRCodeModel.swift
Brandon-T: Fixed

Client/Frontend/Sync/BraveCore/BraveSyncQRCodeModel.swift
Brandon-T: Fixed.

iccub (approved):
Changes requested by me looking good, thanks for adding unit tests to it

soner-yuksel (approved):

Client/Frontend/Sync/BraveCore/BraveSyncQRCodeModel.swift
```diff
@@ -0,0 +1,145 @@
...
+        }
+        
+        // Invalid sync code validation
+        let codeWords = syncAPI.syncCode(fromHexSeed: syncHexCode)
+        if codeWords.isEmpty {
+            return .invalidFormat
+        }
+        
+        let codeWordsCount = codeWords.separatedBy(" ").count
+        if codeWordsCount < 24 {
```
fmarier: We can be more strict here: `!= 24`

fmarier (approved):
One small change to the logic, but otherwise, this looks all good.

Thanks for adding a unit test!

PR #4859, opened 2022-01-17
Fix #4858: WKContentWorld Sandboxing - Sandbox as much as possible

### LINKED ISSUE #4858: WKContentWorld Sandboxing
## Problem Description 
- As of iOS 14+ we can use `WKContentWorld` to sandbox scripts so they are separated from the page

## Feature Overview 
- Solves security issues where pages can call our custom message handlers and functions and XSS.
- Fixes potential exploits in JS injection by enforcing a sandboxing policy when injecting JS.

## Implementation Details 
- All scripts injected must use an explicit content world to make it clear to the security team whether a script lives outside the sandbox or not.
- If a script lives outside the sandbox, we need to explain why to the security team, and try to make this as strict as possible so that only things that are absolutely necessary can live outside the sandbox (overriding JS functions and variables for example).

## User Experience 

1. User will be much much safer

### Additional information 

What needs to be re-tested to make sure it works as it did before:

- Downloading an item (PDF for example)
- Printing the current page (Share Menu -> Print OR call `window.print()`)
- Suppressing Javascript alerts when a page is annoying (Some annoying page that sends infinite alerts)
- Reader-Mode (Wikipedia or any reader enabled pages)
- Bookmarklets (Save a bookmarklet like: `javascript:alert('test');` and execute it on any page)
- Brave-Search
- Login/Auto-Fill (Test login auto-fill to make sure it still works on pages where credentials was saved. Test that it prompts to save)
- Find-In-Page (Settings -> Find In Page. Test that it still works and that going previous and next result works)
- Tracking protection stats (visit a page with tracking such as `Youtube`, and test that the stats increase)

Brandon-T:
## Summary of Changes
**Sandboxes:**
- Reader-Mode
- LoginsHelper
- MetadataHelper
- FindInPage
- NoImageMode
- CustomSearchHandler
- WindowRenderHelper
- ResourceDownloader

**Isolates:**
- Bookmarklets

**Doesn't Sandbox (these override JS functions or define prototypes or Object.define):**
- Playlist (can be done if we use MutationObserver)
- Media Backgrounding
- BraveSearch
- Cookie Control
- Fingerprinting Protection
- Fullscreen Helper
- Youtube Adblock
- ArchiveIsCompat

This pull request fixes #4858

## Submitter Checklist:

## Test Plan:
- Included in the ticket

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


iccub (approved):

thypon (approved):
It seems to be in line with the previous changeset.

The possibility to use `escapeArgs: false` still seems a bad idea to me.

Brandon-T:
> It seems to be in line with the previous changeset.
> 
> The possibility to use `escapeArgs: false` still seems a bad idea to me.

I addressed some of it just now. I removed a lot of the code that was doing `escapeArgs: false` unnecessarily.
I will make a follow-up PR that removes it entirely. For now, I don't want to remove this legacy code because it can potentially break things and would introduce more testing for QA.
This will go out with 1.35.x. 
In 1.36.x, we will remove it entirely and test what breaks and what doesn't, and fix each individual case :)

PR #4952, opened 2022-02-07
VPN/SKUs

### LINKED ISSUE #4805: Detect and store VPN state after sign-in
### Description: 
Desktop is supporting VPN purchases using the account.brave.com website and (in browser) using the SKU SDK (see implementation in https://github.com/brave/brave-core/pull/10847; not merged yet).

The goal would be to:
- have the `account.brave.com` page detect the APIs (this covered by https://github.com/brave/brave-ios/issues/4804)
- when user purchases VPN on the website, the website will call these exposed APIs
- a Chromium profile setting (`skus.credential.has_vpn`; in code as `kSkusVPNHasCredential`) gets set. See `src/brave/components/brave_vpn/brave_vpn_service_desktop.cc` in https://github.com/brave/brave-core/pull/10847 for an example

When this is set, we now can use the credential and authenticate with Guardian VPN.

### Additional details:
- there is a new 1.2 Guardian VPN API call created for `GetSubscriberCredential` on Desktop which can be seen here:
    https://github.com/brave/brave-core/blob/fce7128db24e80f348b29c54f0ba1a9e4387babd/components/brave_vpn/brave_vpn_service.cc#L204-L220 - we'll need to add it into our code (ex: in https://github.com/brave/brave-ios/tree/development/Client/Frontend/BraveVPN/GRDAPI)
- When the Chromium profile value for `skus.credential.has_vpn` is set to `true`, we need to call this new Guardian API to get subscriber credentials using the desktop credentials delivered by the SKU service
- If a profile has VPN purchased, we should consider the user as having purchased VPN in iOS. They should see the same options and UI that a person would if they did the In-App-Purchase. Even though a profile setting is being set, we should (in iOS) set another value so we know this user has used account.brave.com.
- Having this stored value of the person DID at one time purchase VPN from account.brave.com would let us know that once the Chromium profile value has `false` for `kSkusVPNHasCredential` that the person either didn't renew their billing OR that they cancelled their subscription to the product. For this case, we may consider showing a special UI (like "Need to renew?" or similar). cc: @anthonypkeane @mattmcalister for what that experience might be like
- If VPN is NOT purchased, we do NOT need to show a link to account.brave.com inside the app yet (ex: Desktop is linking to account.brave.com/plans and showing a link "Already purchased VPN?".) For iOS, we can consider In-App-Purchase as the primary way to get VPN. But we will explain to folks (via support, press release, etc) that they have the ability to visit account.brave.com to redeem something they purchased before.

#### iOS
- [Wire up SKU SDK to a JavaScript DOM element](https://github.com/brave/brave-ios/issues/4804)

#### Android
- [(Android) Detect and store VPN state after sign-in](https://github.com/brave/brave-browser/issues/20374)

#### Desktop
- [Integrate the SKU SDK into the browser](https://github.com/brave/brave-browser/issues/17614)
- [(Desktop) Detect and store VPN state after sign-in](https://github.com/brave/brave-browser/issues/17254)

iccub:
## Summary of Changes

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/BraveShared/BraveStrings.swift
+++ b/BraveShared/BraveStrings.swift
@@ -2092,10 +2092,16 @@ extension Strings {
                               value: "Expires",
                               comment: "Table cell title for cell that shows when the VPN subscription expires.")
         
-        public static let settingsManageSubscription =
+        public static let settingsManageBraveAccountSubscription =
+            NSLocalizedString("vpn.settingsManageBraveAccountSubscription",
+                              bundle: .braveShared,
+                              value: "Manage Brave Account Subscription",
+                              comment: "Button to manage your VPN subscription from Brave website")
+        
+        public static let settingsManageAppleSubscription =
             NSLocalizedString("vpn.settingsManageSubscription",
                               bundle: .braveShared,
-                              value: "Manage Subscription",
+                              value: "Manage Apple Subscription",
                               comment: "Button to manage your VPN subscription")
         
         public static let settingsServerHost =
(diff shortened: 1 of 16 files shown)


iccub:
Stale PR, work continues here https://github.com/brave/brave-ios/pull/5712

PR #5176, opened 2022-03-30
No Bug: Use os.log.Logger instead of XCGLogger for logs

iccub:
## Summary of Changes

- XCGLogger dep is removed.
- Old rolling log files are deleted during migration
- os.log is used instead
- added new logger type to distinct between brave-core and brave-rewards
- removed the need to initialize a private variable for easy log access

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


iccub:
will revisit it later

PR #5181, opened 2022-03-30
Added JS event listener for tab web-state.

### LINKED ISSUE #5180: [Bug] - Websites with Dynamic Loading not storing screenshots & state properly
### Description: 
- Some websites like Youtube & Twitter uses dynamic loading. They modify the history stack in the browser, so we don't get a callback like `didFinish` in `WKWebView`. This can cause problems where we attempt to save screenshots for TabTray or saving History. The history part was solved in an earlier PR, so this PR will solidify that and solve the screenshots part.

### Steps to Reproduce 
  1. See: https://github.com/brave/brave-ios/issues/5075

**Actual result:** 
- Youtube or Twitter not session restoring properly.
- Screenshots in TabTray sometimes blank.

**Expected result:**
- Should restore correct page
- Should show correct tabs screenshots in TabTray.

Easily reproduced

**Brave Version:** 
1.35

Brandon-T:
## Security Review
- https://github.com/brave/security/issues/810

## Summary of Changes
- Fix tab screenshots in TabTray
- Solidify Session Restore of dynamic pages

This pull request fixes #5180

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/Helpers/ReadyStateScriptHelper.swift
+++ b/Client/Frontend/Browser/Helpers/ReadyStateScriptHelper.swift
@@ -1,0 +1,82 @@
+// Copyright 2022 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import Shared
+import WebKit
+
+private let log = Logger.browserLogger
+
+struct ReadyState: Codable {
+  let securityToken: String
+  let state: State
+  
+  enum State: String, Codable {
+    // Page State
+    case loading
+    case interactive
+    case complete
+    case loaded
+    
+    // History State
+    case pushstate
+    case replacestate
+    case popstate
+  }
+  
+  public static func from(message: WKScriptMessage) -> ReadyState? {
+    if !JSONSerialization.isValidJSONObject(message.body) {
+      return nil
+    }
+
+    do {
+      let data = try JSONSerialization.data(withJSONObject: message.body, options: [])
+      return try JSONDecoder().decode(ReadyState.self, from: data)
+    } catch {
+      log.error("Error Decoding ReadyState: \(error)")
+    }
+
+    return nil
+  }
+  
+  private enum CodingKeys: String, CodingKey {
+    case securityToken = "securitytoken"
+    case state
+  }
+}
+
+class ReadyStateScriptHelper: TabContentScript {
+  private weak var tab: Tab?
+  private var debounceTimer: Timer?
+
+  required init(tab: Tab) {
+    self.tab = tab
+  }
+  
+  class func name() -> String {
+    return "ReadyStateScriptHelper"
+  }
+
+  func scriptMessageHandlerName() -> String? {
+    return "ReadyState_\(UserScriptManager.messageHandlerTokenString)"
+  }
+
+  func userContentController(_ userContentController: WKUserContentController, didReceiveScriptMessage message: WKScriptMessage, replyHandler: (Any?, String?) -> Void) {
+    
+    defer { replyHandler(nil, nil) }
+
+    guard let readyState = ReadyState.from(message: message) else {
+      log.error("Invalid Ready State")
+      return
+    }
+    
+    guard readyState.securityToken == UserScriptManager.securityTokenString else {
+      log.error("Invalid or Missing security token")
+      return
+    }
+    
+    tab?.onPageReadyStateChanged?(readyState.state)
+  }
+}
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -491,7 +491,27 @@ class TabManager: NSObject {
     if flushToDisk && !zombie && !isPrivate {
       saveTab(tab, saveOrder: true)
     }
-
+    
+    // When the state of the page changes, we debounce a call to save the screenshots and tab information
+    // This fixes pages that have dynamic URL via changing history
+    // as well as regular pages that load DOM normally.
+    var debounce_timer: Timer?
+    tab.onPageReadyStateChanged = { [weak self] state in
+      guard let self = self else { return }
+      
+      debounce_timer?.invalidate()
+      debounce_timer = Timer.scheduledTimer(withTimeInterval: 1.0, repeats: false) { _ in
+        debounce_timer?.invalidate()
+        
+        if state == .complete || state == .loaded || state == .pushstate || state == .popstate {
+          // Saving Tab Private Mode - not supported yet.
+          if !tab.isPrivate {
+            self.preserveScreenshots()
+            self.saveTab(tab)
+          }
+        }
+      }
+    }
   }
 
   func indexOfWebView(_ webView: WKWebView) -> UInt? {
--- a/Client/Frontend/Browser/UserScriptManager.swift
+++ b/Client/Frontend/Browser/UserScriptManager.swift
@@ -363,6 +363,31 @@ class UserScriptManager {
       forMainFrameOnly: true,
       in: .defaultClient)
   }()
+  
+  private let ReadyStateScript: WKUserScript? = {
+    guard let path = Bundle.main.path(forResource: "ReadyState", ofType: "js"), let source = try? String(contentsOfFile: path) else {
+      log.error("Failed to load ReadyState.js")
+      return nil
+    }
+
+    var alteredSource = source
+    let token = UserScriptManager.securityTokenString
+
+    let replacements = [
+      "$<security_token>": token,
+      "$<handler>": "ReadyState_\(messageHandlerTokenString)",
+    ]
+
+    replacements.forEach({
+      alteredSource = alteredSource.replacingOccurrences(of: $0.key, with: $0.value, options: .literal)
+    })
+
+    return WKUserScript.create(
+      source: alteredSource,
+      injectionTime: .atDocumentStart,
+      forMainFrameOnly: true,
+      in: .page)
+  }()
 
   private func reloadUserScripts() {
     tab?.webView?.configuration.userContentController.do {
@@ -401,6 +426,10 @@ class UserScriptManager {
       }
 
       if isNightModeEnabled, let script = NightModeScript {
+        $0.addUserScript(script)
+      }
+      
+      if let script = ReadyStateScript {
         $0.addUserScript(script)
       }
 
--- a/Client/Frontend/UserContent/UserScripts/ReadyState.js
+++ b/Client/Frontend/UserContent/UserScripts/ReadyState.js
@@ -66,12 +66,8 @@
   };
   
   // Hide the pushState trampoline
-  History.prototype.pushState.toString = function() {
-    return "function () { [native code] }";
-  };
+  History.prototype.pushState.toString = pushState.toString;
   
   // Hide the replaceState trampoline
-  History.prototype.replaceState.toString = function() {
-    return "function () { [native code] }";
-  };
+  History.prototype.replaceState.toString = replaceState.toString;
 })();
--- a/Client/Frontend/UserContent/UserScripts/ReadyState.js
+++ b/Client/Frontend/UserContent/UserScripts/ReadyState.js
@@ -1,0 +1,77 @@
+// Copyright 2021 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+(function() {
+  // Listen for document ready state
+  document.addEventListener('readystatechange', (event) => {
+    window.webkit.messageHandlers.$<handler>.postMessage({
+      "securitytoken": "$<security_token>",
+      "state": document.readyState
+    });
+  });
+  
+  // Listen for document load state
+  window.addEventListener('load', (event) => {
+    window.webkit.messageHandlers.$<handler>.postMessage({
+      "securitytoken": "$<security_token>",
+      "state": "loaded"
+    });
+  });
+  
+  // Listen for history popped
+  window.addEventListener('popstate', (event) => {
+    if (event.state) {
+      // Run on the browser's next run-loop
+      setTimeout(() => {
+        window.webkit.messageHandlers.$<handler>.postMessage({
+          "securitytoken": "$<security_token>",
+          "state": "popstate"
+        });
+      }, 0);
+    }
+  });
+  
+  // Listen for history pushed
+  const pushState = History.prototype.pushState;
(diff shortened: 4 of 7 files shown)


Client/Frontend/Browser/UserScriptManager.swift
```diff
@@ -363,6 +363,31 @@ class UserScriptManager {
...
+    var alteredSource = source
+    let token = UserScriptManager.securityTokenString
+
+    let replacements = [
+      "$<security_token>": token,
+      "$<handler>": "ReadyState_\(messageHandlerTokenString)",
+    ]
+
+    replacements.forEach({
+      alteredSource = alteredSource.replacingOccurrences(of: $0.key, with: $0.value, options: .literal)
```
pes10k, line 382: I don't know if in Swift this ends up alloc'ing multiple copies of the same inserted script, but you can get the same behavior without needing the find / replace or string operations by doing the approach @cuba used here:
https://github.com/brave/brave-ios/pull/5156/files#diff-a059035e1311558281829f11147315c797da830d78176db17a5515210e98e1a3R72

Anyway, not a blocking issue, but perf and code cleanliness suggestion (since you always have valid JS)

Client/Frontend/UserContent/UserScripts/ReadyState.js
```diff
@@ -0,0 +1,77 @@
...
+  // Listen for document ready state
+  document.addEventListener('readystatechange', (event) => {
+    window.webkit.messageHandlers.$<handler>.postMessage({
+      "securitytoken": "$<security_token>",
+      "state": document.readyState
+    });
+  });
+  
+  // Listen for document load state
+  window.addEventListener('load', (event) => {
```
pes10k, line 16: small suggestions, but you might remove this event handler after it fires, otherwise a page could trigger the `load` (and similar) events multiple times (using `dispatchEvent`), which will likely cause the backend / swift code to do unexpected things (if, for example, if the backend code expects each page will only `load` once, etc)

Client/Frontend/UserContent/UserScripts/ReadyState.js
```diff
@@ -0,0 +1,77 @@
...
+          "securitytoken": "$<security_token>",
+          "state": "replacestate"
+        });
+      }, 0);
+    }
+  };
+  
+  // Hide the pushState trampoline
+  History.prototype.pushState.toString = function() {
+    return "function () { [native code] }";
```
pes10k: would this be better as (run before any page modifications)?
```
const pushState = History.prototype.pushState;
…
History.prototype.pushState.toString = _ => pushState.toString()
```

You'd get the same effect, but without needing to hardcode any special strings, and wouldn't need to update anything if the "real" `History.prototype.pushState.toString()` changes

pes10k (approved):
looks good to me privacy-wise, though i made a few small suggestions for cleanup and consistency

Client/Frontend/Browser/UserScriptManager.swift
Brandon-T: The file linked above does do a `find and replace` operation.
If you mean caching, I will be doing it this way in a future PR: https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Frontend/TabContentScripts/UserScriptManager.swift#L22

Client/Frontend/UserContent/UserScripts/ReadyState.js
Brandon-T: I need it called more than once for pages that change the document (twitter sometimes calls `window.location.reload(true);`).
If it is called too frequently, the `debounce` will trigger and delay the final call. But if I do need it called only once, I will guard it with a boolean on the Swift side. For now though, I want it this way.

Client/Frontend/UserContent/UserScripts/ReadyState.js
Brandon-T: Fixed.

iccub (approved):
Good job, i ran this code for a bit and seems to work for all cases and websites

Client/Frontend/Browser/UserScriptManager.swift
pes10k: I think we might be talking about different things; im referring to the technique @cuba does to avoid having to splice up JS code for the farbling.  Client/Frontend/UserContent/UserScripts/FarblingProtection.js is a complete JS implementation, and then he just concatenates an invocation with it at the location i linked to.

I'm suggesting this technique for two reasons:
- It ensures proper JS is in the repo, so it can be linted / checked / etc
- it (might, depending on what Swift does under the hood) prevents the need to keep multiple copies of the JS code in memory, since you don't need a copy per copy/replace operation

Client/Frontend/Browser/Helpers/ReadyStateScriptHelper.swift
```diff
@@ -0,0 +1,82 @@
...
+  public static func from(message: WKScriptMessage) -> ReadyState? {
+    if !JSONSerialization.isValidJSONObject(message.body) {
+      return nil
+    }
+
+    do {
+      let data = try JSONSerialization.data(withJSONObject: message.body, options: [])
+      return try JSONDecoder().decode(ReadyState.self, from: data)
+    } catch {
+      log.error("Error Decoding ReadyState: \(error)")
```
diracdeltas, line 38: could this error contain any part of the JSON which might contain sensitive info? (like page URL)

Client/Frontend/Browser/UserScriptManager.swift
Brandon-T: I understood what you meant, but it doesn't apply here. The reason `FarblingProtection` is a complete JS implementation is because it doesn't have to replace the `SecurityToken` in the JS. That's because it doesn't have any callbacks to Swift Code. IE: It doesn't pass any information from JS to Swift. My code passes information from JS to Swift, and so in order to prevent pages from accessing the message handlers directly, we have a security token replacement done at runtime, as this cannot be sandboxed fully.

Same code as he is doing here: https://github.com/brave/brave-ios/pull/5156/files#diff-a059035e1311558281829f11147315c797da830d78176db17a5515210e98e1a3R87

Which is done in the rest of our code already (including this PR).

pes10k: I understand, but why implement your injected script as a function, and then pass the security token as a parameter to the function (this is what the linked to code does, it handles the farbling values in the same way your code could handle the security token

Client/Frontend/Browser/Helpers/ReadyStateScriptHelper.swift
Brandon-T: @diracdeltas Hi. No it won't contain any sensitive info. It will only contain a reason as to why the JSON failed to deserialize. IE: `"Expected array, but got dictionary"`. `"Expected String, but received Integer"`, `"Missing field: name_of_missing_field",` etc.

Client/Frontend/Browser/UserScriptManager.swift
pes10k: Like this https://gist.github.com/pes10k/fb6041957406b11f81188c18e6361c2a

And then inject `alteredSource` + "window.bravePageEvents(\(<swift_security_token>), \(<swift_handler_name));delete window.bravePageEvents";

Brandon-T: Why would I do that though? That requires me to `evaluateJavascript` every time the script is injected. That means my script won't be able to run `onDocumentStart` which is done automatically by WebKit.

If I do it your way, I have to do `webKit.evaluateSafeJavascript(function: "window.braveEvents", args: ["securityToken", ...])` somewhere in the code. Possibly when the page is already finished loading. Plus this is also `async`, so I may not get the event at all.

Client/Frontend/Browser/TabManager.swift
```diff
@@ -491,7 +491,27 @@ class TabManager: NSObject {
...
+    var debounce_timer: Timer?
+    tab.onPageReadyStateChanged = { [weak self] state in
+      guard let self = self else { return }
+      
+      debounce_timer?.invalidate()
+      debounce_timer = Timer.scheduledTimer(withTimeInterval: 1.0, repeats: false) { _ in
+        debounce_timer?.invalidate()
+        
+        if state == .complete || state == .loaded || state == .pushstate || state == .popstate {
+          // Saving Tab Private Mode - not supported yet.
```
diracdeltas, line 507: just curious do we plan to support it? seems it'd need a privacy review if so

Client/Frontend/Browser/UserScriptManager.swift
Brandon-T: Once we switch over to the code I linked from Firefox for ALL the scripts, it'd be done better than that:
`let wrappedSource = "(function() { const SECURITY_TOKEN = '\(UserScriptManager.securityToken)'; \(source) })()"` https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Frontend/TabContentScripts/UserScriptManager.swift#L41

So we won't have to execute a function and then delete it right after.
Our script manager is the older version of theirs.

pes10k: To my mind, that still seems likely to wind up with a bunch of large strings one-off strings that aren't cacheable.  I dont understand why thats preferable, but its also not a privacy concern so im fine dropping the issue and closing the privacy review

Brandon-T: `source + "window.functionName(\(UserScriptManager.securityToken), \(UserScriptManager.messageHandlerName)); delete window.functionName";`

Is the same as:
`"(function() { const SECURITY_TOKEN = "\(UserScriptManager.securityToken)"; const MESSAGE_HANDLER = "\(UserScriptManager.messageHandler)"; \(source) })()"`

All the source has to do is use the `SECURITY_TOKEN` and `MESSAGE_HANDLER` just the same as if it was a parameter, but instead is a constant. No function name required.

The result of either one would be cached.

pes10k: At least in some languages that would not be true, in terms of number of allocs and memory used.  I don't know how swift would handle that kind of string concatenation or interpolation, so maybe im generalizing incorrectly.  I'm more than happy to defer to your Swift knowledge here and drop this concern

Brandon-T: In case you're curious, I did some tests and interpolation `"\(a) \(b) \(c)"` is much better than appending `a + b + c` in Swift with regards to allocations, memory space, instruction count, bridging, and performance.

image
image

Client/Frontend/Browser/TabManager.swift
Brandon-T: In the future yes we might support it (or so I've heard), but it is a long way from now I think. I put the comment here because we don't currently support saving private tabs to disk. I think we will when we move to Chromium's Tabs implementation along with TabGroups and so on. That will require us to refactor everywhere with `Tab` (including the above). We will definitely open a security and privacy review if/when we do support it. We will have to open more than one (BraveCore + iOS).

diracdeltas: sgtm

diracdeltas:
@Brandon-T i was not done reviewing this yet - in the future please wait til sec review is closed

Client/Frontend/UserContent/UserScripts/ReadyState.js
```diff
@@ -0,0 +1,73 @@
+// Copyright 2021 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+(function() {
+  // Listen for document ready state
+  document.addEventListener('readystatechange', (event) => {
+    window.webkit.messageHandlers.$<handler>.postMessage({
+      "securitytoken": "$<security_token>",
```
diracdeltas, line 10: are we sure that there's no way webpages can see the token here using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/toString or similar?

Client/Frontend/Browser/UserScriptManager.swift
```diff
@@ -363,6 +363,31 @@ class UserScriptManager {
...
+      log.error("Failed to load ReadyState.js")
+      return nil
+    }
+
+    var alteredSource = source
+    let token = UserScriptManager.securityTokenString
+
+    let replacements = [
+      "$<security_token>": token,
+      "$<handler>": "ReadyState_\(messageHandlerTokenString)",
```
diracdeltas, line 378: what happens if a webpage includes some JS which includes the string `$<security_token>`? will it get replaced and potentially allow that page to see the token?

Brandon-T:
> @Brandon-T i was not done reviewing this yet - in the future please wait til sec review is closed

Ahh sorry. I saw the `sgtm` (and I read it as `seems good to merge`). Sorry again. I should have checked.
I can revert it and then reopen this PR.

Client/Frontend/Browser/UserScriptManager.swift
Brandon-T: Web pages will not be able to do that because this is injected `atDocumentStart` so before any of the web page's source code is executed. This script only does the replacement in Swift and not in the web-page code as well. Since it is replaced in Swift code, it wouldn't matter what the page does before or after it is injected.

Client/Frontend/UserContent/UserScripts/ReadyState.js
Brandon-T: Yes I am sure. I made sure to test for this, and all our existing scripts do the same.
Any script that exposes a function for a web-page has the `toString` function overridden to return `[native method]` as a string, and not the implementation: https://github.com/brave/brave-ios/pull/5181#discussion_r839012778

`webkit.messageHandlers` is also not enumerable, and any functions on our `window.__firefox__` is stored as a `Proxy` object with all the functions overridden so that it cannot be accessed.

Client/Frontend/UserContent/UserScripts/ReadyState.js
```diff
@@ -0,0 +1,73 @@
...
+          "securitytoken": "$<security_token>",
+          "state": "popstate"
+        });
+      }, 0);
+    }
+  });
+  
+  // Listen for history pushed
+  const pushState = History.prototype.pushState;
+  History.prototype.pushState = function(state, unused, url) {
```
diracdeltas, line 38: what happens if a webpage redefines `pushState/replaceState`? i guess this feature would just not work?

Brandon-T: Correct it would just do nothing. I also tested this by doing: `History.prototype.pushState = function() { console.log("Hello"); }` in the Safari Debug Console.

It replaces my script's methods with the one that prints hello, and just breaks the page as well as doing nothing with regards to the script.

diracdeltas:
> I can revert it and then reopen this PR.

@Brandon-T thanks, no need in this case since my comments are mostly questions. but for future reference sgtm = `sounds good to me` and the sec review ticket should be the source of truth.

PR #5184, opened 2022-03-30
Fix #5179: Tab Tray does not work in landscape mode

### LINKED ISSUE #5179: Tab Tray does not work in landscape mode
### Steps to Reproduce 
  1. Use Brave in landscape mode on iPhone
  2. Tap on Tabs button to open up the tab tray

**Actual result:** 
The tab tray is displayed in portrait mode

**Expected result:**
Should be in landscape

**Brave Version:** 

**Device details:** 

**Website problems only:**

soner-yuksel:
Changing Tray Tray Navigation controller and small changes are made for search bar frame to adjust properly with landscape.

## Summary of Changes

This pull request fixes #5179

## Submitter Checklist:

## Test Plan:

- Use Brave in landscape mode on iPhone
- Tap on Tabs button to open up the tab tray

## Screenshots:

https://user-images.githubusercontent.com/6643505/160931543-fd549cc9-a69b-428a-8c20-9a0398c703a0.mp4

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/Tab Tray/TabTrayController.swift
+++ b/Client/Frontend/Browser/Tab Tray/TabTrayController.swift
@@ -95,6 +95,8 @@ class TabTrayController: LoadingViewController {
     }
 
     navigationItem.do {
+      // Place the search bar in the navigation item's title view.
+      $0.titleView = searchBarView
       $0.hidesSearchBarWhenScrolling = true
     }
 
@@ -126,10 +128,6 @@ class TabTrayController: LoadingViewController {
       UIBarButtonItem(barButtonSystemItem: .flexibleSpace, target: self, action: nil),
       UIBarButtonItem(customView: tabTrayView.doneButton),
     ]
-    
-    // Place the search bar in the navigation item's title view.
-    searchBarView?.frame = navigationController?.navigationBar.frame ?? .zero
-    navigationItem.titleView = searchBarView
   }
 
   private var initialScrollCompleted = false
--- a/Client/Frontend/Browser/Tab Tray/TabTrayController.swift
+++ b/Client/Frontend/Browser/Tab Tray/TabTrayController.swift
@@ -54,6 +54,7 @@ class TabTrayController: LoadingViewController {
   private var isTabTrayBeingSearched = false
   private let tabTraySearchController = UISearchController(searchResultsController: nil)
   private var tabTraySearchQuery = ""
+  private var searchBarView: TabTraySearchBar?
 
   private lazy var emptyStateOverlayView: UIView = EmptyStateOverlayView(description: Strings.noSearchResultsfound)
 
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/Tab Tray/TabTrayController.swift
```diff
@@ -146,6 +149,16 @@ class TabTrayController: LoadingViewController {
       initialScrollCompleted = true
     }
   }
+  
+  override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+    super.traitCollectionDidChange(previousTraitCollection)
+
+    searchBarView?.frame = navigationController?.navigationBar.frame ?? .zero
```
iccub, line 154: Maybe do not call this code if horizontal size class did not change? add something like 
```
if previousTraitCollection?.horizontalSizeClass == traitCollection.horizontalSizeClass { return }
```

need to be checked if this works

iccub (changes requested):

Client/Frontend/Browser/Tab Tray/TabTrayController.swift
```diff
@@ -127,6 +126,10 @@ class TabTrayController: LoadingViewController {
       UIBarButtonItem(barButtonSystemItem: .flexibleSpace, target: self, action: nil),
       UIBarButtonItem(customView: tabTrayView.doneButton),
     ]
+    
+    // Place the search bar in the navigation item's title view.
+    searchBarView?.frame = navigationController?.navigationBar.frame ?? .zero
```
StephenHeaps: I don't know if we need this frame assignment in `viewDidLoad` since it seems to layout as expected when view loads, just breaks during rotation when the view is visible. Then I think we can put the `titleView` assignment back in the `do {}` closure above? 🙂

StephenHeaps (approved):

Client/Frontend/Browser/Tab Tray/TabTrayController.swift
soner-yuksel: Good point let me try. Fixed at https://github.com/brave/brave-ios/pull/5184/commits/e4e833221ac82acfb129bedaf0019af87d2e4247

Client/Frontend/Browser/Tab Tray/TabTrayController.swift
soner-yuksel: This doesn't not going to work since the main part that makes this work is fetching the frame of navigation bar every time trait change happens and setting it to search bar View and updating titleView. This is **not** a costly operation but has to be done to make titleView work properly.

iccub (approved):
tested on few device configurations, works very well :)

PR #5190, opened 2022-03-31
Revert "Added JS event listener for tab web-state."

Brandon-T:
Reverts brave/brave-ios#5181


diracdeltas:
ok not to do this https://github.com/brave/brave-ios/pull/5181#issuecomment-1084988933

PR #5224, opened 2022-04-06
Fix #4862 - #5215: Some problems around Custom Search Engine

### LINKED ISSUE #4862: Search Suggestions fail for Custom Search Engine After Relaunch
### Description: 
Adding a custom search engine via `Menu` ---> `Add Search Engine` allows the search suggestions to work properly. However, if you relaunch Brave, they no longer work and the search engine has to be deleted and re-added. This can be quite frustrating and not a new issue.

### Steps to Reproduce 
  1. Add a custom search engine like SearX or Whoogle using `Menu` --> `Add Search Engine`
  2. Assign that search engine to be the default in `Standard Tab` in `Search Engine` settings
  3. Ensure `Show Search Suggestions` is toggled on.
  4. DO NOT CLOSE Brave and try to search for something.
  5. Notice how the search suggestions are populating.
  6. Now close Brave `completely`using the `App Switcher`
  7. Relaunch Brave and search for something.
  8. Search suggestions are no longer populating.

**Actual result:** 
Search suggestions don't show up after relaunching Brave and search engine has to be deleted and re-added.

**Expected result:**
Search suggestions should work normally regardless just like a search engines built into Brave.

**Reproduces how often:**
Easily

**Brave Version:**
v1.34 (22.1.10.19)

**Device details:** 
iPhone SE 2020. iOS 15.2.1

### LINKED ISSUE #5215: Deletion problem Custom-Search Engine
### Description: 

Cant delete the Custom-Added Search Engine from Search Engines Screen when the engine is default engine. And user is not warned about this behaviour at all. 

### Steps to Reproduce 
  1. Add 2 Custom Search Engines
  2. Go Settings Search Engines
  3. Try to delete first Custom Search Engine

**Actual result:** 

First Engine cant be deleted

**Expected result:**

All Custom Engines should be deleted and user should get a proper prompt.

engines

Easily Reproduced

**Brave Version:** 

1.37 Appstore

**Device details:** 

All Devices Models

**Website problems only:**

soner-yuksel:
Issue #4862 is around suggestions from custom search engines. Clearing and Launching the application makes custom search engines suggestions not work properly. As indicated in the ticket:

- Adding a custom search engine via Menu ---> Add Search Engine allows the search suggestions to work properly. However, if you relaunch Brave, they no longer work and the search engine has to be deleted and re-added. 

Issue #5215 is about deletion of Custom Search Engines.

- When a Custom Search Engine is defined as default search engine, we cant delete the engine until default engine is changed to some other engine. To prevent this behaviour and give more functionality we allowed users to delete custom search engines and added a warning related with this, also after deleting the custom search engine, default engine private mode will be changed as default engine on default browse.

## Summary of Changes

This pull request fixes #4862
This pull request fixes #5215

## Submitter Checklist:

## Test Plan:

Issue #4862

- Add a custom search engine like SearX or Whoogle using Menu --> Add Search Engine
- Assign that search engine to be the default in Standard Tab in Search Engine settings
- Ensure Show Search Suggestions is toggled on.
- DO NOT CLOSE Brave and try to search for something.
- Notice how the search suggestions are populating.
- Now close Brave completely using the App Switcher
- Relaunch Brave and search for something.
- Search suggestions should be working.

Issue #5215

- Add a custom search engine like SearX or Whoogle using Menu --> Add Search Engine
- Assign that search engine to be the default in Standard Tab in Search Engine settings
- Try to delete that engine edit or action
- An lert should be presented to user and accepting the warning should delete the engine 

## Screenshots:

https://user-images.githubusercontent.com/6643505/162036217-a123c4be-d975-411e-9c7f-99f452ef8ddb.mp4

https://user-images.githubusercontent.com/6643505/162036237-72586b39-307a-4153-913d-4f5a827f1b55.mp4

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/BraveShared/BraveStrings.swift
+++ b/BraveShared/BraveStrings.swift
@@ -605,7 +605,7 @@ extension Strings {
       comment: "The alert title shown to user when custom search engine will be deleted while it is default search engine. The parameter will be replace with name of the search engine.")
     
     public static let deleteEngineAlertDescription = NSLocalizedString(
-      "customSearchEngine.deleteEngineAlertTitle",
+      "customSearchEngine.deleteEngineAlertDescription",
       bundle: .braveShared,
       value: "Deleting a custom search engine while it is default will switch default engine automatically.",
       comment: "The warning description shown to user when custom search engine will be deleted while it is default search engine.")
--- a/BraveShared/BraveStrings.swift
+++ b/BraveShared/BraveStrings.swift
@@ -597,6 +597,18 @@ extension Strings {
       bundle: .braveShared,
       value: "The new search engine will appear in the quick search bar.",
       comment: "The message that asks the user to Add the search provider explaining where the search engine will appear")
+  
+    public static let deleteEngineAlertTitle = NSLocalizedString(
+      "customSearchEngine.deleteEngineAlertTitle",
+      bundle: .braveShared,
+      value: "Are you sure you want to delete %@?",
+      comment: "The alert title shown to user when custom search engine will be deleted while it is default search engine. The parameter will be replace with name of the search engine.")
+    
+    public static let deleteEngineAlertDescription = NSLocalizedString(
+      "customSearchEngine.deleteEngineAlertTitle",
+      bundle: .braveShared,
+      value: "Deleting a custom search engine while it is default will switch default engine automatically.",
+      comment: "The warning description shown to user when custom search engine will be deleted while it is default search engine.")
   }
 }
 
--- a/Client/Frontend/Browser/Search/OpenSearch.swift
+++ b/Client/Frontend/Browser/Search/OpenSearch.swift
@@ -93,11 +93,12 @@ class OpenSearchEngine: NSObject, NSSecureCoding {
     self.isCustomEngine = isCustomEngine
     self.image = image
     self.engineID = aDecoder.decodeObject(of: NSString.self, forKey: "engineID") as String?
-    self.suggestTemplate = nil
+    self.suggestTemplate = aDecoder.decodeObject(of: NSString.self, forKey: "suggestTemplate") as String?
   }
 
   func encode(with aCoder: NSCoder) {
     aCoder.encode(searchTemplate, forKey: "searchTemplate")
+    aCoder.encode(suggestTemplate, forKey: "suggestTemplate")
     aCoder.encode(shortName, forKey: "shortName")
     aCoder.encode(isCustomEngine, forKey: "isCustomEngine")
     aCoder.encode(image, forKey: "image")
--- a/Client/Frontend/Browser/Search/SearchEngines.swift
+++ b/Client/Frontend/Browser/Search/SearchEngines.swift
@@ -194,8 +194,8 @@ class SearchEngines {
   }
 
   func deleteCustomEngine(_ engine: OpenSearchEngine) throws {
-    // We can't delete a preinstalled engine or an engine that is currently the default.
-    if !engine.isCustomEngine || isEngineDefault(engine) {
+    // We can't delete a preinstalled engine
+    if !engine.isCustomEngine {
       return
     }
 
--- a/Client/Frontend/Settings/SearchSettingsTableViewController.swift
+++ b/Client/Frontend/Settings/SearchSettingsTableViewController.swift
@@ -312,23 +312,42 @@ class SearchSettingsTableViewController: UITableViewController {
   override func tableView(_ tableView: UITableView, commit editingStyle: UITableViewCell.EditingStyle, forRowAt indexPath: IndexPath) {
     if editingStyle == .delete {
       guard let engine = customSearchEngines[safe: indexPath.row] else { return }
+      
+      func deleteCustomEngine() {
+        do {
+          try searchEngines.deleteCustomEngine(engine)
+          tableView.deleteRows(at: [indexPath], with: .right)
+          tableView.reloadData()
+        } catch {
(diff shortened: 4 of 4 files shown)


BraveShared/BraveStrings.swift
```diff
@@ -597,6 +597,18 @@ extension Strings {
...
       comment: "The message that asks the user to Add the search provider explaining where the search engine will appear")
+  
+    public static let deleteEngineAlertTitle = NSLocalizedString(
+      "customSearchEngine.deleteEngineAlertTitle",
+      bundle: .braveShared,
+      value: "Are you sure you want to delete %@?",
+      comment: "The alert title shown to user when custom search engine will be deleted while it is default search engine. The parameter will be replace with name of the search engine.")
+    
+    public static let deleteEngineAlertDescription = NSLocalizedString(
+      "customSearchEngine.deleteEngineAlertTitle",
```
iccub: duplicate of previous key

iccub (changes requested):

Client/Frontend/Browser/Search/SearchEngines.swift
```diff
@@ -194,8 +194,8 @@ class SearchEngines {
   }
 
   func deleteCustomEngine(_ engine: OpenSearchEngine) throws {
-    // We can't delete a preinstalled engine or an engine that is currently the default.
-    if !engine.isCustomEngine || isEngineDefault(engine) {
+    // We can't delete a preinstalled engine
+    if !engine.isCustomEngine {
```
cuba, line 198: I would probably use a guard here that way you know without even looking inside the block it will never execute further.
```
guard engine.isCustomEngine else { return }
```

Client/Frontend/Settings/SearchSettingsTableViewController.swift
```diff
@@ -312,23 +312,47 @@ class SearchSettingsTableViewController: UITableViewController {
   override func tableView(_ tableView: UITableView, commit editingStyle: UITableViewCell.EditingStyle, forRowAt indexPath: IndexPath) {
     if editingStyle == .delete {
       guard let engine = customSearchEngines[safe: indexPath.row] else { return }
+      
+      func deleteCustomEngine() {
+        do {
+          try searchEngines.deleteCustomEngine(engine)
+          tableView.deleteRows(at: [indexPath], with: .right)
+          tableView.reloadData()
```
cuba, line 320: I don't think you have to do both `deleteRows` and `reloadData` it's usually one or the other.

Client/Frontend/Settings/SearchSettingsTableViewController.swift
```diff
@@ -312,23 +312,47 @@ class SearchSettingsTableViewController: UITableViewController {
...
-        log.error("Search Engine Error while deleting")
+      if engine == searchEngines.defaultEngine(forType: .standard) {
+        let alert = UIAlertController(
+          title: String(format: Strings.CustomSearchEngine.deleteEngineAlertTitle, engine.displayName),
+          message: Strings.CustomSearchEngine.deleteEngineAlertDescription,
+          preferredStyle: .alert)
+        
+        alert.addAction(UIAlertAction(title: Strings.cancelButtonTitle, style: .cancel) { _ in
+          return
+        })
```
cuba: you don't need that callback on the cancel action

cuba (commented):
Just some minor nitpicks

Client/Frontend/Browser/Search/SearchEngines.swift
iccub: for context: we used to use `guard` only when some value is unwrapped, never for bool logic alone,
nowadays people use it as they want, i'm fine with both conventions

soner-yuksel: I will keep as it is since we are also moving around the requirements here.

Client/Frontend/Settings/SearchSettingsTableViewController.swift
soner-yuksel: Not in this case, I want that row to be deleted with animation and the table should be updated to reflect the default search engine change.

Client/Frontend/Settings/SearchSettingsTableViewController.swift
soner-yuksel: Fix at https://github.com/brave/brave-ios/pull/5224/commits/a66d24d71d2d6b343d255ea1fa48af2aa6d479ea

BraveShared/BraveStrings.swift
soner-yuksel: Fix at https://github.com/brave/brave-ios/pull/5224/commits/a66d24d71d2d6b343d255ea1fa48af2aa6d479ea

Client/Frontend/Browser/Search/OpenSearch.swift
```diff
@@ -93,11 +93,12 @@ class OpenSearchEngine: NSObject, NSSecureCoding {
     self.isCustomEngine = isCustomEngine
     self.image = image
     self.engineID = aDecoder.decodeObject(of: NSString.self, forKey: "engineID") as String?
-    self.suggestTemplate = nil
+    self.suggestTemplate = aDecoder.decodeObject(of: NSString.self, forKey: "suggestTemplate") as String?
```
iccub, line 96: my guess was right :D

iccub (approved):

PR #5263, opened 2022-04-19
Fix #5251: Keep user on Sync chain always (until they manually leave)

### LINKED ISSUE #5251: Sync users are sometimes kicked off out of the sync chain
Setting as QA/No because we never managed to reproduce it

Brandon-T:
## Summary of Changes
- Do not ever leave the sync chain until a button is pressed
- Show error alert but do nothing after it is shown
- If no devices found on the chain, do nothing (technically this should never happen, but it seems like it is, based on this report: https://bravesoftware.slack.com/archives/C2HJYB45N/p1650373022612009?thread_ts=1649769016.099419&cid=C2HJYB45N)

This pull request fixes #5251

## Submitter Checklist:

## Test Plan:
- Test that syncing still works (joining, leaving, being removed by other device, etc)

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Sync/SyncAddDeviceViewController.swift
+++ b/Client/Frontend/Sync/SyncAddDeviceViewController.swift
@@ -98,11 +98,6 @@ class SyncAddDeviceViewController: SyncViewController {
     containerView.layer.cornerCurve = .continuous
     containerView.layer.masksToBounds = true
 
-    if !syncAPI.isInSyncGroup {
-      showInitializationError()
-      return
-    }
-
     qrCodeView = SyncQRCodeView(syncApi: syncAPI)
     containerView.addSubview(qrCodeView!)
     qrCodeView?.snp.makeConstraints { make in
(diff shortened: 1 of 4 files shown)


Client/Frontend/Sync/SyncAddDeviceViewController.swift
```diff
@@ -114,11 +109,19 @@ class SyncAddDeviceViewController: SyncViewController {
     self.codewordsView.text = syncAPI.getSyncCode()
     self.setupVisuals()
   }
+  
+  override func viewDidAppear(_ animated: Bool) {
+    super.viewDidAppear(animated)
+    
+    if !syncAPI.isInSyncGroup {
+      DispatchQueue.main.asyncAfter(deadline: .now() + 1.0) { [weak self] in
```
soner-yuksel: Do we need the delay here since we are already showing the error in viewDidAppear ?

soner-yuksel (approved):

Client/Frontend/Sync/SyncAddDeviceViewController.swift
Brandon-T: Fixed.

PR #5278, opened 2022-04-22
Fix #4219: Add de-amp settings

### LINKED ISSUE #4219: Add the ability to dump AMP
### Description: 
Add the ability to stop Google AMP pages loading, instead load the native URL.

Users will have the option to enable this setting or not and we'll need to decide whether it is defaulted on or off

Some work was done [here](https://github.com/brave/brave-core/pull/4950) but needs to be finished and incorporated into the products. 

### Links
https://github.com/brave/security/issues/850

cc @pes10k

cuba:
## Summary of Changes

This pull request fixes #4219

## Submitter Checklist:

## Test Plan:
1. Disable de-amp setting
2. Go to google [search](https://www.google.com)
3. Do a search for "top stories"
4. In the top stories section some links will take you to a link similar to: `https://www.google.com/amp/s/{some_url}` (take note of the part in {some_url})
5. Go to the following links and you should not be redirected anywhere:
* `https://www.cbc.ca/amp/1.6432270`
* `https://fingerprint-protection-test.herokuapp.com/amp`
6. Enable de-amp settings
7. Go to the following urls and you should be redirected
* `https://www.cbc.ca/amp/1.6432270` -> `https://www.cbc.ca/news/{some-path}` (notice `/amp` is missing)
* `https://fingerprint-protection-test.herokuapp.com/amp` -> `https://amp.dev/documentation/guides-and-tutorials/start/create/basic_markup/`
8. Go to google [search](https://www.google.com)
9. Do a search for "top stories"
10. In the top stories section when you click on links you should **not** be sent to `https://www.google.com/amp/s/{some_url}`

## Screenshots:

How it looks like in desktop (for reference)
Screen Shot 2022-04-21 at 12 23 28 PM

This is how it looks in iOS:
Simulator Screen Shot - iPad Air (5th generation) - 2022-04-27 at 15 09 21

### Links
https://github.com/brave/security/issues/850

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Client/WebFilters/ShieldStats/Adblock/AdBlockStats.swift
```diff
@@ -198,6 +198,15 @@ extension AdBlockStats {
       
       cssRules += "\(key){" + subRules + " !important}\n"
     }
+
+    var injectedScript = model.injectedScript
+
+    if !injectedScript.isEmpty, Preferences.Shields.autoRedirectAMPPages.value {
+      injectedScript = [
+        "const deAmpEnabled = true;",
```
Brandon-T, line 211: This variable isn't used anywhere in the `DeAMP.js` script.

Client/WebFilters/ShieldStats/Adblock/AdBlockStats.swift
```diff
@@ -198,6 +198,15 @@ extension AdBlockStats {
       
       cssRules += "\(key){" + subRules + " !important}\n"
     }
+
+    var injectedScript = model.injectedScript
+
+    if !injectedScript.isEmpty, Preferences.Shields.autoRedirectAMPPages.value {
```
Brandon-T, line 204: Why is `deAmpEnabled = true` only injected if `CosmeticFilters` are injected? Are they not independent of each other?

Client/Frontend/Browser/User Scripts/UserScriptType.swift
```diff
@@ -8,64 +8,80 @@ import WebKit
...
-        case .nacl:
-            return .nacl
-        }
+  /// Return a source typ for this script type
+  var sourceType: ScriptSourceType {
+    switch self {
+    case .farblingProtection:
+      return .farblingProtection
+    case .deAMP:
+      return .deAMP
```
Brandon-T: Why are we injecting `deAmp` on `decidePolicyFor` instead of in `UserScriptManager` globally?

Brandon-T (changes requested):

Brandon-T:
Might need a sec/privacy review since JS injection is involved.

Client/Frontend/Browser/User Scripts/UserScriptType.swift
```diff
@@ -8,64 +8,80 @@ import WebKit
...
+  var contentWorld: WKContentWorld {
+    switch self {
+    case .farblingProtection, .domainUserScript, .nacl:
+      return .page
+    case .deAMP:
+      // Note: this has the same content world as the `cosmeticFiltersScript`
+      // executed in the `WKNavigationDelegate`. This is currently no strictly necessary
+      // but if we ever share properties amongst them, they need to be in the
+      // same content world
+      return .cosmeticFiltersSandbox
```
Brandon-T: Should probably have its own sandbox or `defaultClient` as I don't see any reason why it would share the sandbox of cosmetic filters. If it needs to share anything in the future, we can change it then.

Brandon-T (changes requested):

Client/Frontend/UserContent/UserScripts/DeAMP.js
```diff
@@ -0,0 +1,63 @@
...
+      }
+      W.location.replace(destUrl.href)
+    } catch (_) {
+      // Invalid canonical URL detected
+      W.clearInterval(intervalId)
+      return
+    }
+  }
+
+  intervalId = W.setInterval(checkForAmp, 1000)
```
pes10k: my fault nit, but can this timeout be changed to 250, and `timesToCheck` be changed to 20?

Brandon-T (approved):

Client/Frontend/UserContent/UserScripts/DeAMP.js
```diff
@@ -0,0 +1,63 @@
...
+
+    const targetHref = canonicalLinkElm.getAttribute('href')
+    try {
+      const destUrl = new URL(targetHref)
+      if (W.location.href == destUrl.href) {
+        // Avoid an infinite loading loop in some cases
+        W.clearInterval(intervalId)
+        return
+      }
+      W.location.replace(destUrl.href)
```
thypon, line 54: Can we replace this call with something like:

```javascript
window.open('...', '_blank')
window.close()
```

Can we also verify that the protocol is either `http` or `https`?

`if (destUrl.href.protocol !== 'http' && destUrl.href.protocol !== 'https'` should return eagerly.

This will safeguard from malicious links running in the current google amp context.

thypon (changes requested):

Client/Frontend/UserContent/UserScripts/DeAMP.js
pes10k:
```
window.open('...', '_blank')
window.close()
```

I dont think we should do this.  This will open a new window and close the current one, which means you'll loose back-forward, etc.

@thypon is there a security reason to prefer the above?  `W.location = <url>` is (as far as i know) the/a common way to redirect the page.

> Can we also verify that the protocol is either http or https?

This seems like a great idea

thypon: The proposed solution is a typical source-sink approach. Sanitize/validate the `source`, sandbox the `sink`.

Source: @pes10k this check `if (destUrl.href.protocol !== 'http' && destUrl.href.protocol !== 'https'` should already guard us against 99% of the issues we may face.

Sink: Since we cannot forecast the future on how an attacker may craft an URL that bypass the first check, but is interpreted as a `javascript:...` - e.g. through special encoding/characters - we may also sandbox the output before an attacker reaches the target and pwns the parent's context.

@Brandon-T can you confirm that is not possible to do a `window.location.replace` with target `_blank` on iOS? Do you know any other solution to this problem?

pes10k:
I don’t think we need to worry about the child ruining the parent here: this code is only applied in the top level frame

pes10k:
But more broadly, writing to location isn’t a privileged operation; page script can already do it. The code here doesn’t give the attacker any new capabilities

thypon:
I may have misunderstood the threat model here, so correct me if I'm wrong. An AMP page/site should define this meta tag which is then copied into the google AMP context. To de-amp, we arbitrarily decide that this compatibility link is safe and we automatically navigate to it.

> I don’t think we need to worry about the child ruining the parent here: this code is only applied in the top level frame

Since we are executing in the Google AMP context a link provided by a child page we are effectively executing this content in the parent Google domain. In this case, an attacker may include `javascript:...(document.cookie)` which may exfiltrate the current Google context credentials.

From what I understood from a quick chat with @cuba , this applies to non-AMP pages as well, so the scope of this vulnerability may be bigger.

Client/Frontend/UserContent/UserScripts/DeAMP.js
Brandon-T: @thypon `window.location.replace` doesn't have a target parameter afaik. But also, any attempts to redirect to a `javascript` URL will execute the Javascript immediately and not redirect. If somehow the URL is decoded to be `javascript`, it is caught on the redirect on the iOS side (in Swift) via: https://github.com/brave/brave-ios/blob/development/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift#L139

which is `if url.isBookmarklet { ..cancel.. }` to cancel the navigation (checks if the scheme of the URL is `javascript:`.

pes10k:
> An AMP page/site should define this meta tag which is then copied into the google AMP context

I'm not sure what is meant here.  The page could get code executed in the current context (whether thats the Google AMP cache or some other origin serving AMP), but the page can already execute code in that same context.  The only situation i can think of where this would be unique attack surface would be if the attacker could control the value of the canonical url metatag, but *couldn't* otherwise alter page markup.  That seems like a pretty odd / one-off privacy boundary to try and guard against, unless i am misunderstanding the concern.

> From what I understood from a quick chat with @cuba , this applies to non-AMP pages as well, so the scope of this vulnerability may be bigger.

This script is injected into non-AMP pages, but the guards at the top of the function (like [this one](https://github.com/brave/brave-ios/pull/5278/files#diff-fd9d159a48f7dccb1d98842de313c00f0837495b9cc86fffac7c55b2ec936bb9R32)) prevent the code you're concerned about running on non-amp pages

thypon:
Screenshot 2022-05-11 at 00 56 35

NVM, it seems we already have a second safeguard by the Google AMP validator

To clarify my concern: google AMP pages are pages that may be hosted by google on a google domain. The same-site policy allows code executed in the context of that page to have access to the cookies/session/credentials of that page (google domain).

If a `javascript:...` instruction or `<script>` tag is executed in that context, It will be able to interact with the Google CDN page context.

I wonder if we can execute that script in a bookmarklet context to safeguard this issue further. But given at least two failsafe are in place (one by google/one by us), I feel we are good to go.

thypon (approved):

cuba:
@thypon could I then get the sec issue closed? I need that to be able to merge this one. Also wondering since someone can in theory create an amp page but not add the google amp scripts which means that google won't be doing any safeguards. but we will still navigate as if it were an amp page as we don't check any presence of scripts. Anything we need to worry about there? Doesn't seem like a likely scenario to me and not sure if that could mean anything security wise.

Client/Frontend/Browser/UserScriptManager.swift
```diff
@@ -72,6 +72,14 @@ class UserScriptManager {
       reloadUserScripts()
     }
   }
+  
+  /// Whether night mode is enabled for webview
```
iccub: update comment to mention de-amp

iccub (changes requested):

iccub (approved):

PR #5279, opened 2022-04-22
Fix #3222: Add ability to zoom pages

### LINKED ISSUE #3222: Add support for increasing website's font size in Brave
### Description: 
Slack: https://bravesoftware.slack.com/archives/C06UXF3KJ/p1611102899052600

Pease add support for allowing users to increase/decrease to size of the font in Brave

- [x] Complete design - where will the UI go, how will it work
- [x] Implement design

Brandon-T:
## Summary of Changes
- Added global zoom level.
- Allow site level preference to take priority over global preference
- Update pages zoom-levels when Zoom changes from the UI in real-time.

This pull request fixes #3222

## Screenshots:

image image image image

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
```diff
@@ -0,0 +1,172 @@
...
+    do {
+      let domains = try context.fetch(request)
+      return domains.sorted(by: { $0.host < $1.host })
+    } catch {
+      return []
+    }
+  }
+}
+
+// #if DEBUG
```
StephenHeaps: I think this should be uncommented?

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
```diff
@@ -0,0 +1,172 @@
...
+  @State private var defaultZoomLevel = Preferences.General.defaultPageZoomLevel.value
+  
+  var body: some View {
+    List {
+      if !domains.isEmpty {
+        Section(
+          header: Text(Strings.PageZoom.specificWebsitesZoomLevelsSectionTitle)
+            .font(.footnote)
+            .foregroundColor(Color(.secondaryBraveLabel))
+            .multilineTextAlignment(.leading)
```
StephenHeaps: Default alignment will be `.leading`, this shouldn't be needed (same for 5 line below in this file it's used I think; you may want `.trailing` for the right side aligned `Text`’s)

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
```diff
@@ -0,0 +1,172 @@
...
+            HStack {
+              let url = domain.url ?? ""
+              
+              let zoomLevel = domain.zoom_level?.doubleValue ?? 1.0
+              
+              let percentageLevel = Int(zoomLevel * 100.0)
+              
+              Text(url)
+                .font(.headline)
+                .foregroundColor(.white)
```
StephenHeaps: The foreground color's aren't updating in light mode 🤔. 
first open settings -  zoom level
after visiting a website

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,189 @@
...
+  }
+  
+  private func decrement() {
+    guard let index = PageZoomView.steps.firstIndex(of: currentValue) else { return }
+    currentValue = PageZoomView.steps[index - 1]
+    storeChanges()
+  }
+}
+
+//#if DEBUG
```
StephenHeaps: Same here, I think this should be uncommented?

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
StephenHeaps: Oh I am tired. This should be using `bravePrimary` or `.braveLabel`* color I think 🙂

iccub: No foregroundColor set at all would fix as well, it should adapt

BraveShared/BraveStrings.swift
```diff
@@ -4330,3 +4330,78 @@ extension Strings {
...
 }
+
+// Page Zoom
+extension Strings {
+  public struct PageZoom {
+    public static let settingsMenuTitle = NSLocalizedString(
+      "pagezoom.settings.menu-title",
+      bundle: .braveShared,
+      value: "Page Zoom",
+      comment: "Title of the Page Zoom settings menu"
```
kylehickinson: You should expand on the words, such as that "Page" refers to "Webpage" for translators sake (This applies to all localized string comments below).  All of these comments just simply explain where something is but not what its supposed to mean.

BraveShared/Preferences.swift
```diff
@@ -226,6 +226,13 @@ extension Preferences {
     public func reset() {
       value = defaultValue
     }
+    
+    public func notify() {
+      let key = self.key
+      observers.forEach {
+        $0.preferencesDidChange(for: key)
+      }
+    }
```
kylehickinson: This shouldn't be something that an outside API should call. Its not up to the user to notify that things have changed, its up to the receiver to subscribe to changes and the pref object itself to notify when the value has changed.

Is there a bug somewhere that's not notifying properly?

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1838,6 +1845,57 @@ class BrowserViewController: UIViewController, BrowserViewControllerDelegate {
...
+      updateViewConstraints()
+    }
+    
+    alertStackView.arrangedSubviews.forEach({
+      $0.removeFromSuperview()
+    })
+    alertStackView.addArrangedSubview(pageZoomBar.view)
+
+    pageZoomBar.view.snp.makeConstraints { make in
+      make.height.equalTo(UIConstants.toolbarHeight)
```
kylehickinson: We probably shouldn't height constrain this, maybe `greaterThanOrEqualTo` instead

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Menu.swift
```diff
@@ -294,6 +294,11 @@ extension BrowserViewController {
               browserViewController.openAddBookmark()
             }
           }
+          MenuItemButton(icon: UIImage(systemName: "binoculars")?.template ?? UIImage(), title: Strings.PageZoom.settingsMenuTitle) {
```
kylehickinson: This icon is changing to `aA` icon. The SF Symbol is `textformat.size` but we may use a Brave version here in the menu instead

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
```diff
@@ -0,0 +1,168 @@
...
+import SwiftUI
+import Shared
+import BraveShared
+import Data
+import CoreData
+
+extension Domain: Identifiable {
+  public var id: String {
+    url ?? UUID().uuidString
+  }
```
kylehickinson: Why wouldn't we use the Domain's `objectID` as its identifier? If we're not going to we shouldn't fallback to a UUID that's not stable

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
```diff
@@ -0,0 +1,168 @@
...
+              
+              let percentageLevel = Int(zoomLevel * 100.0)
+              
+              Text(url)
+                .font(.headline)
+                .foregroundColor(Color(UIColor.braveLabel))
+              
+              Spacer()
+              
+              Text(PageZoomView.percentFormatter.string(from: .init(value: zoomLevel)) ?? "\(percentageLevel)%")
```
kylehickinson: Use `Text(zoomLevel, formatter: PageZoomView.percentFormatter)` instead

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
```diff
@@ -0,0 +1,168 @@
...
+  
+  @State private var defaultZoomLevel = Preferences.General.defaultPageZoomLevel.value
+  
+  var body: some View {
+    List {
+      if !domains.isEmpty {
+        Section(
+          header: Text(Strings.PageZoom.specificWebsitesZoomLevelsSectionTitle)
+            .font(.footnote)
+            .foregroundColor(Color(UIColor.secondaryBraveLabel))
```
kylehickinson: All these `UIColor` prefixes are redundant

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
```diff
@@ -0,0 +1,170 @@
...
+          let preferredZoomLevel = Int(defaultZoomLevel * 100.0)
+         
+          Button(action: {
+            Preferences.General.defaultPageZoomLevel.value = step
+            defaultZoomLevel = step
+            
+            domains = PageZoomSettingsView.filteredDomains(context)
+          }, label: {
+            HStack(spacing: 0.0) {
+              Text(PageZoomView.percentFormatter.string(from: .init(value: step)) ?? "\(zoomLevel)%")
```
kylehickinson: Same as above

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
```diff
@@ -0,0 +1,170 @@
...
+                .foregroundColor(Color(UIColor.braveLabel))
+                .multilineTextAlignment(.trailing)
+              
+              Spacer()
+              
+              if preferredZoomLevel == zoomLevel {
+                Image(systemName: "checkmark")
```
kylehickinson, line 81: Probably needs a medium/semibold font weight to match standard table cell style checkmark

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
```diff
@@ -0,0 +1,170 @@
...
+            PageZoomSettingsView.allDomains(context).forEach({
+              $0.zoom_level = nil
+            })
+            
+            try? context.save()
+            
+            domains = PageZoomSettingsView.filteredDomains(context)
+            
+            Preferences.General.defaultPageZoomLevel.notify()
+          }
```
kylehickinson: Can we move this into a separate function the same way we have the private func for resetting item zoom level?

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
```diff
@@ -0,0 +1,170 @@
...
+    } catch {
+      return []
+    }
+  }
+  
+  private static func filteredDomains(_ context: NSManagedObjectContext) -> [Domain] {
+    let request = NSFetchRequest<Domain>().then {
+      $0.entity = Domain.entity(context)
+      $0.sortDescriptors = [NSSortDescriptor(key: "url", ascending: true)]
+      $0.predicate = NSPredicate(format: "zoom_level != nil AND zoom_level != %@", NSNumber(value: Preferences.General.defaultPageZoomLevel.value))
```
kylehickinson: Wouldn't you want to show overrides for all websites regardless of if it matches the default zoom level?

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+
+import SwiftUI
+import BraveUI
+import WebKit
+import Data
+import Shared
+import BraveShared
+
+private struct ZoomView: View {
+  private let color = Color(UIColor.braveLabel)
```
kylehickinson: Just use `Color(.braveLabel)` in each place individually, setting it like this without any doc or proper naming is confusing

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+  
+  var onDecrement: () -> Void
+  var onReset: () -> Void
+  var onIncrement: () -> Void
+  
+  var body: some View {
+    HStack(spacing: 0.0) {
+      Button(action: {
+        onDecrement()
+      }) {
```
kylehickinson: `Button(action: onDecrement) {`

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+        Image(systemName: "minus")
+          .font(.system(.footnote).weight(.medium))
+          .foregroundColor(color)
+          .frame(width: 44.0)
+          .padding(.vertical, 8.0)
+          .contentShape(Rectangle())
+      }
+      .fixedSize()
```
kylehickinson: No fixed size

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+  
+  var body: some View {
+    HStack(spacing: 0.0) {
+      Button(action: {
+        onDecrement()
+      }) {
+        Image(systemName: "minus")
+          .font(.system(.footnote).weight(.medium))
+          .foregroundColor(color)
+          .frame(width: 44.0)
```
kylehickinson: Use a ScaledMetric instead of hardcoding

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+          .padding(.vertical, 8.0)
+          .contentShape(Rectangle())
+      }
+      .fixedSize()
+      .disabled(value == minValue)
+      
+      Divider()
+        .padding(.vertical, 8.0)
+      
+      Text(PageZoomView.percentFormatter.string(from: .init(value: 0.999)) ?? "999%")
```
kylehickinson: Same comment as above for formatter/Text

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+        .font(.system(.footnote).weight(.medium))
+        .fixedSize()
```
kylehickinson: Only needs horizontal fixed size, not vertical

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+        onIncrement()
+      }) {
+        Image(systemName: "plus")
+          .font(.system(.footnote).weight(.medium))
+          .foregroundColor(color)
+          .frame(width: 44.0)
+          .padding(.vertical, 8.0)
+          .contentShape(Rectangle())
+      }
+      .fixedSize()
```
kylehickinson: Same comments as minus button

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+    HStack(spacing: 0.0) {
+      Button(action: {
+        onDecrement()
+      }) {
+        Image(systemName: "minus")
+          .font(.system(.footnote).weight(.medium))
+          .foregroundColor(color)
+          .frame(width: 44.0)
+          .padding(.vertical, 8.0)
+          .contentShape(Rectangle())
```
kylehickinson: Don't think `contentShape` is needed here anymore

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+      Text(PageZoomView.percentFormatter.string(from: .init(value: 0.999)) ?? "999%")
+        .font(.system(.footnote).weight(.medium))
+        .fixedSize()
+        .hidden()
+        .overlay(resetZoomButton)
+        .padding(.horizontal)
+        .padding(.vertical, 8.0)
+      
+      Divider()
+        .padding(.vertical, 8.0)
```
kylehickinson: You have 8pt's of vertical padding on every child, therefore you should remove it from all children and add it to the parent instead

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+    }
+    .fixedSize(horizontal: false, vertical: true)
+    .background(Color(UIColor.secondaryBraveBackground))
+    .clipShape(RoundedRectangle(cornerRadius: 10.0, style: .continuous))
+  }
+  
+  private var resetZoomButton: some View {
+    Button(action: {
+      onReset()
+    }) {
```
kylehickinson: `Button(action: onReset) {`

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+      Text(PageZoomView.percentFormatter.string(from: .init(value: value)) ?? "\(Int(value * 100.0))%")
```
kylehickinson: Same as above

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+  init(webView: WKWebView?) {
+    self.webView = webView
+    
+    if let url = webView?.url,
+       let domain = Domain.getPersistedDomain(for: url) {
+      
+      _currentValue = State(initialValue: domain.zoom_level?.doubleValue ?? Preferences.General.defaultPageZoomLevel.value)
+    } else {
+      _currentValue = State(initialValue: webView?.value(forKey: PageZoomView.propertyName) as? Double ?? Preferences.General.defaultPageZoomLevel.value)
+    }
```
kylehickinson, line 117: Maybe leave a comment here mentioning that this is OK because you are only presenting `PageZoomView` via `UIHostingController` so there's no chance of re-init.

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+  
+  public static let steps = [0.5, 0.75, 0.85,
+                             1.0, 1.15, 1.25,
+                             1.50, 1.75, 2.00,
+                             2.50, 3.0]
+  
+  var dismiss: (() -> Void)?
+  
+  init(webView: WKWebView?) {
+    self.webView = webView
```
kylehickinson, line 101: Do we need to store the WKWebView weakly?

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+  
+  var body: some View {
+    VStack(spacing: 0.0) {
+      Divider()
+      HStack(spacing: 0.0) {
+        Text(Strings.PageZoom.zoomViewText)
+          .font(.system(.subheadline))
+          .frame(maxWidth: .infinity, alignment: .leading)
+        
+        Spacer()
```
kylehickinson: Spacer is not needed

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+        ZoomView(
+          minValue: minValue,
+          maxValue: maxValue,
+          value: $currentValue,
+          onDecrement: decrement,
+          onReset: reset,
+          onIncrement: increment)
+          .frame(maxWidth: .infinity)
+        
+        Spacer()
```
kylehickinson: Spacer is not needed

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+        Button {
+          dismiss?()
+        } label: {
+          Image(systemName: "xmark")
+            .renderingMode(.template)
```
kylehickinson: renderingMode not needed

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+            .foregroundColor(Color(UIColor.braveBlurple))
+            .font(.system(.footnote).weight(.medium))
+        }
+        .frame(maxWidth: .infinity, alignment: .trailing)
+      }
+      .padding(.horizontal)
+      .padding(.vertical, 7.0)
```
kylehickinson: Use 6 or 8

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,208 @@
...
+class PageZoomViewController: UIHostingController<PageZoomView> {
+
+  init(webView: WKWebView) {
+    super.init(rootView: PageZoomView(webView: webView))
+  }
+
+  @objc required dynamic init?(coder aDecoder: NSCoder) {
+    fatalError("init(coder:) has not been implemented")
+  }
+}
```
kylehickinson: TBH this is probably not needed at all, no additional changes aside from a name change so you are better off just using `UIHostingController` directly at init point.

Client/Frontend/Settings/SettingsViewController.swift
```diff
@@ -329,6 +329,10 @@ class SettingsViewController: TableViewController {
         title: Strings.mediaAutoBackgrounding,
         option: Preferences.General.mediaAutoBackgrounding,
         image: #imageLiteral(resourceName: "background_play_settings_icon").template),
+      Row(text: "Zoom Level", selection: { [weak self] in
+        let controller = UIHostingController(rootView: PageZoomSettingsView().environment(\.managedObjectContext, DataController.swiftUIContext))
+        self?.navigationController?.pushViewController(controller, animated: true)
+      }, image: UIImage(systemName: "binoculars")?.template ?? UIImage(), accessory: .disclosureIndicator, cellClass: MultilineValue1Cell.self)
```
kylehickinson: Same icon change as menu

kylehickinson (changes requested):

BraveShared/Preferences.swift
Brandon-T: No but I have a rare case where I want to force a notification otherwise I have to write more notification code :(
Basically, I have an observer that when the preference changes it updates all the webViews with the new zoom level. Instead of duplicating this for when a `specific` web-page's zoom level changes, I just trigger the same observer. I'm really not sure if I should just write separate code for it or reuse what's already there and just call `notify` which I added.

Client/Frontend/Browser/BrowserViewController.swift
Brandon-T: Hmm but the `findInPage` bar is height constrained so I did it here as well. Also, if it's not this exact height, sometimes it goes `below` the safeArea insets.

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
Brandon-T: It will not compile without it. https://bravesoftware.slack.com/archives/GB9CJGQV6/p1650635341001139
It gives: 
image

Even after breaking up the entire view into separate parts, it still gives that error. So I fixed it by putting `UIColor.` in front of everything.

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
Brandon-T: No because if the default zoom level is 125%, it will show `125%` for all websites if they have that zoom level. The idea here is that if the two levels match, just don't show it.

So instead of having a bunch of websites with `125%` showing, just only show the ones that aren't the same as the default. I guess I could technically show all of them if they're not `nil` alone. But to me doesn't make much sense.

If you want it I will just remove the zoom level check there.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Menu.swift
Brandon-T: Changed to `textformat.size` does look a bit weird though, but I don't have a custom icon yet.

BraveShared/Preferences.swift
kylehickinson: This seems weird though, and maybe needs a different solution. You're notifying the observers that something has changed when nothing has, so I'd say this function should definitely not exist

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -23,70 +23,56 @@ private struct ZoomView: View {
...
+    Button(action: onReset) {
+      Text(NSNumber(value: value), formatter: PageZoomView.percentFormatter)
         .font(.system(.footnote).weight(.medium))
-        .foregroundColor(color)
+        .foregroundColor(Color(UIColor.braveLabel))
         .padding()
         .contentShape(Rectangle())
     }
-    .fixedSize()
+    .fixedSize(horizontal: true, vertical: false)
```
kylehickinson, line 69: Don't need this fixed size (the layout is determined by the parent already, this will always be overlayed on top)

Brandon-T: Doesn't render at all if I remove it. `overlay` doesn't seem to inherit the parent's fixedSize.

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,181 @@
...
+  private var resetZoomButton: some View {
+    Button(action: onReset) {
+      Text(NSNumber(value: value), formatter: PageZoomView.percentFormatter)
+        .font(.system(.footnote).weight(.medium))
+        .foregroundColor(Color(UIColor.braveLabel))
+        .padding()
+        .contentShape(Rectangle())
+    }
+    .fixedSize(horizontal: true, vertical: false)
+    .disabled(value == 1.0)
```
kylehickinson: Does reset == 1 always? Or does reset set to default in settings? If its default this should be checking that not 1

BraveShared/Preferences.swift
Brandon-T: Fixed. Used NotificationCenter.

Client/Frontend/Browser/PageZoom/PageZoomView.swift
Brandon-T: Changed it. It used to be just 1.

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,182 @@
...
+  }
+  
+  private func increment() {
+    guard let index = PageZoomView.steps.firstIndex(of: currentValue) else { return }
+    currentValue = PageZoomView.steps[index + 1]
+    storeChanges()
+  }
+  
+  private func reset() {
+    currentValue = 1.0
```
kylehickinson: This should be the default value, no?

Brandon-T: Actually had to think about that for a while. Whether or not resetting meant resetting the page to 100% or whether it reset to the default value. I changed it to default everywhere, so it'll reset to whatever the default value is.
Safari seems to reset to default as well, so it matches now.

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
```diff
@@ -0,0 +1,197 @@
...
+        }
+        .listRowBackground(Color(UIColor.secondaryBraveGroupedBackground))
+      }
+    }
+    .navigationTitle(Strings.PageZoom.settingsTitle)
+    .navigationBarTitleDisplayMode(.inline)
+    .listStyle(.insetGrouped)
+  }
+  
+  func normalizeZoomLevels() {
```
iccub, line 121: Can you add comment here explaining why do we need to normalize zoom levels?

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
```diff
@@ -0,0 +1,197 @@
...
+  }
+  
+  private static func allDomains(_ context: NSManagedObjectContext) -> [Domain] {
+    let request = NSFetchRequest<Domain>().then {
+      $0.entity = Domain.entity(context)
+      $0.sortDescriptors = [NSSortDescriptor(key: "url", ascending: true)]
+    }
+    
+    do {
+      let domains = try context.fetch(request)
```
iccub: Why not reuse one of `CRUD` methods `Domain.all(sortDescriptors: [sort..], context: context)`?

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
```diff
@@ -0,0 +1,197 @@
...
+      return domains.sorted(by: { $0.host < $1.host })
```
iccub: Why not add additional sort descriptor instead of sorting it here?

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
```diff
@@ -0,0 +1,197 @@
...
+  
+  private static func filteredDomains(_ context: NSManagedObjectContext) -> [Domain] {
+    let request = NSFetchRequest<Domain>().then {
+      $0.entity = Domain.entity(context)
+      $0.sortDescriptors = [NSSortDescriptor(key: "url", ascending: true)]
+      $0.predicate = NSPredicate(format: "zoom_level != nil")
+    }
+    
+    do {
+      let domains = try context.fetch(request)
```
iccub: Why not reuse one of `CRUD` methods `Domain.all(where: your_predicate, sortDescriptors: [sort..], context: context)`?

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
```diff
@@ -0,0 +1,197 @@
...
+      return domains.sorted(by: { $0.host < $1.host })
```
iccub: Why not add additional sort descriptor instead of sorting it here?

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,182 @@
...
+        Image(systemName: "minus")
+          .font(.system(.footnote).weight(.medium))
+          .foregroundColor(Color(UIColor.braveLabel))
+          .frame(width: buttonWidth)
+      }
+      .disabled(value == minValue)
+      
+      Divider()
+      
+      Text(NSNumber(value: 0.999), formatter: PageZoomView.percentFormatter)
```
iccub, line 38: why this specific number?

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,182 @@
...
+    let domain = Domain.getPersistedDomain(for: url)?.then {
+      $0.zoom_level = currentValue == $0.zoom_level?.doubleValue ? nil : NSNumber(value: currentValue)
+    }
+    
+    try? domain?.managedObjectContext?.save()
+  }
+  
+  private func increment() {
+    guard let index = PageZoomView.steps.firstIndex(of: currentValue) else { return }
+    currentValue = PageZoomView.steps[index + 1]
```
iccub, line 171: Add out of bounds safety check here?

Client/Frontend/Browser/PageZoom/PageZoomView.swift
```diff
@@ -0,0 +1,182 @@
...
+  }
+  
+  private func reset() {
+    currentValue = Preferences.General.defaultPageZoomLevel.value
+    storeChanges()
+  }
+  
+  private func decrement() {
+    guard let index = PageZoomView.steps.firstIndex(of: currentValue) else { return }
+    currentValue = PageZoomView.steps[index - 1]
```
iccub, line 183: Add out of bounds safety check here?

Client/Frontend/Settings/SettingsViewController.swift
```diff
@@ -329,6 +329,10 @@ class SettingsViewController: TableViewController {
         title: Strings.mediaAutoBackgrounding,
         option: Preferences.General.mediaAutoBackgrounding,
         image: #imageLiteral(resourceName: "background_play_settings_icon").template),
+      Row(text: "Zoom Level", selection: { [weak self] in
```
iccub: Missed localization

Data/models/Model.xcdatamodeld/Model15.xcdatamodel/contents
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
iccub: This needs conflict resolution after my Privacy Hub code was merged into dev

iccub (changes requested):
Few DB questions, will do functional review later

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
Brandon-T: Accessible only to `internal` and isn't public.

Client/Frontend/Browser/PageZoom/PageZoomView.swift
Brandon-T: Technically not necessary because the button is disabled if it gets to the maximum. But added anyway.

Client/Frontend/Browser/PageZoom/PageZoomView.swift
Brandon-T: Added a comment in the code. 

TLDR: `999` is the largest 3 digit number in terms of character widths. So if the percentage goes from 50 (2 characters) to 100 (3 characters) then to 300 (3 characters), it means 50 is smaller than 100 width-wise, and 100 is smaller than 300 width-wise because 3 takes up more space than 1. But if I put `999` a fixed size, no matter what the number is, the size of the text there doesn't shift and glitch.

Client/Frontend/Browser/PageZoom/PageZoomSettingsView.swift
Brandon-T: Fixed.

kylehickinson (approved):

iccub (approved):
🔥

PR #5444, opened 2022-06-03
Fix #5374: Add async request blocking via webkit message handlers

### LINKED ISSUE #5374: Block bad window.fetch() messages using the Rust engine
### Description: 
Block certain `window.fetch()` api calls by connecting to the rust engine asynchronously using apples built in javascript to swift mechanism `WKUserContentController`

### Test plan
1. You may test by building the debug version of the application and seeing if no javascript errors are thrown on the injected scripts.
2. Go to a few sites with known trackers and test that nothing is broken. (YouTube, Facebook, instagram, etc)

#### Test 2:
1. Switch languages to something that uses regional lists. (ex switch phone to the polish)
2. Go to twitter and make sure the login works

#### Test 3:
1. Go to  https://openocean.finance/ and make sure the page loads (should ideally run the application on debug mode to ensure no asserts are triggered)

cuba:
## Summary of Changes

This pull request fixes #5374 

## Submitter Checklist:

## Test Plan:

This is a tricky to test feature:
1. You may test by building the debug version of the application and seeing if no javascript errors are thrown on the injected scripts.
2. Go to a few sites with known trackers and test that nothing is broken. (YouTube, Facebook, instagram, etc)3. 

Test 2:
1. Switch languages to something that uses regional lists. (ex switch phone to the polish)
2. Go to twitter and make sure the login works

## Reviewer Checklist:

  - `QA/(Yes|No)` - test for performance degradation
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/User Scripts/UserScriptHelper.swift
+++ b/Client/Frontend/Browser/User Scripts/UserScriptHelper.swift
@@ -73,7 +73,7 @@ class UserScriptHelper {
     return userScriptTypes
   }
   
-  static func makeDeAmpScriptParamters() throws -> String {
+  static func makeDeAmpScriptParamters() throws -> String? {
     let arguments: [String: String] = [
       "handlerName": DeAmpHelper.scriptMessageHandlerName(),
       "securityToken": UserScriptManager.securityTokenString
@@ -81,6 +81,16 @@ class UserScriptHelper {
     
     let encoder = JSONEncoder()
     let data = try encoder.encode(arguments)
-    return String(data: data, encoding: .utf8)!
+    return String(data: data, encoding: .utf8)
+  }
+
+  static func makeRequestBlockingParams(securityToken: String) throws -> String? {
+    let params: [String: String] = [
+      "securityToken": securityToken,
+      "handlerName": RequestBlockingContentHelper.scriptMessageHandlerName()
+    ]
+    let encoder = JSONEncoder()
+    let data = try encoder.encode(params)
+    return String(data: data, encoding: .utf8)
   }
 }
--- a/Client/Frontend/UserContent/UserScripts/RequestBlocking.js
+++ b/Client/Frontend/UserContent/UserScripts/RequestBlocking.js
@@ -1,0 +1,84 @@
+// Copyright 2022 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+window.braveBlockRequests = (args) => {
+  'use strict'
+
+  const securityToken = args.securityToken
+  const sendMessage = (resourceURL) => {
+    return webkit.messageHandlers[args.handlerName].postMessage({
+      securityToken: securityToken,
+      data: {
+        resourceURL: resourceURL.href,
+        sourceURL: window.location.href,
+        resourceType: 'xmlhttprequest'
+      }
+    })
+  }
+  
+  const { fetch: originalFetch } = window
+  window.fetch = function () {
+    const [resource] = arguments
+    
+    // Extract the url
+    let urlString
+    if (typeof resource === 'string') {
+      // Simple send where resource is a string
+      urlString = resource
+    } else if (resource instanceof Request) {
+      // Assume it's a Request
+      urlString = resource.url
+    } else {
+      return originalFetch.apply(this, arguments)
+    }
+
+    const url = new URL(urlString, window.location.href)
+    return sendMessage(url).then(blocked => {
+      if (blocked) {
+        return new Promise(function (resolve, reject) {
+          reject(new TypeError('Load failed'))
+        })
+      } else {
+        return originalFetch.apply(this, arguments)
+      }
+    })
+  }
+  
+  const originalOpen = XMLHttpRequest.prototype.open
+  XMLHttpRequest.prototype.open = function () {
+    // Check if we're async
+    const isAsync = arguments[2]
+    if (isAsync !== undefined && !isAsync) {
+      // We can't handle non-async requests as they are not handled via callbacks
+      return originalOpen.apply(this, arguments)
+    }
+
+    // Extract the url
+    const urlString = arguments[1]
+    this._resourceURL = new URL(urlString, window.location.href)
+    return originalOpen.apply(this, arguments)
+  }
+  
+  const originalSend = XMLHttpRequest.prototype.send
+  XMLHttpRequest.prototype.send = function () {
+    if (this._resourceURL === undefined) {
+      return originalSend.apply(this, arguments)
+    }
+    
+    sendMessage(this._resourceURL).then(blocked => {
+      if (blocked) {
+        Object.defineProperties(this, {
+          readyState: { value: 4 }
+        })
+        this.dispatchEvent(new Event('readystatechange'))
+        this.dispatchEvent(new ProgressEvent('error'))
+      } else {
+        originalSend.apply(this, arguments)
+      }
+    })
+  }
+}
+
+// Invoke window.braveBlockRequests then delete the function
--- a/Client/WebFilters/AdblockResourceDownloader.swift
+++ b/Client/WebFilters/AdblockResourceDownloader.swift
@@ -25,7 +25,15 @@ public class AdblockResourceDownloader {
   private let servicesKeyName = "SERVICES_KEY"
   private let servicesKeyHeaderValue = "BraveServiceKey"
 
-  static let endpoint = "https://adblock-data.s3.brave.com/ios"
+  /// The base s3 environment url that hosts the debouncing (and other) files.
+  /// Cannot be used as-is and must be combined with a path
+  private lazy var baseResourceURL: URL = {
+    if AppConstants.buildChannel.isPublic {
+      return URL(string: "https://adblock-data.s3.brave.com")!
+    } else {
+      return URL(string: "https://adblock-data-staging.s3.bravesoftware.com")!
+    }
+  }()
 
   init(networkManager: NetworkManager = NetworkManager(), locale: String? = Locale.current.languageCode) {
     if locale == nil {
@@ -102,22 +110,19 @@ public class AdblockResourceDownloader {
       
       let fileExtension = fileType.rawValue
       let etagExtension = fileExtension + ".etag"
+      let cacheFileName = [fileName, etagExtension].joined(separator: ".")
 
-      guard let resourceName = type.resourceName(for: fileType),
-        var url = URL(string: AdblockResourceDownloader.endpoint)
-      else {
+      guard let resourcePath = type.resourcePath(for: fileType) else {
         return nil
       }
 
-      url.appendPathComponent(resourceName)
-      url.appendPathExtension(fileExtension)
-
+      let url = URL(string: resourcePath, relativeTo: baseResourceURL)!
       var headers = [String: String]()
       if let servicesKeyValue = Bundle.main.getPlistString(for: self.servicesKeyName) {
         headers[self.servicesKeyHeaderValue] = servicesKeyValue
       }
 
-      let etag = self.fileFromDocumentsAsString("\(fileName).\(etagExtension)", inFolder: folderName)
+      let etag = self.fileFromDocumentsAsString(cacheFileName, inFolder: folderName)
       return nm.downloadResource(
         with: url,
         resourceType: .cached(etag: etag),
@@ -229,8 +234,6 @@ public class AdblockResourceDownloader {
           return self.compileContentBlocker(resources: resources)
         }
         return nil
-      case .tgz:
-        return nil  // TODO: Add downloadable httpse list
       }
     }
     
--- a/Client/WebFilters/AdblockResourceDownloader.swift
+++ b/Client/WebFilters/AdblockResourceDownloader.swift
@@ -116,7 +116,11 @@ public class AdblockResourceDownloader {
         return nil
       }
 
-      let url = URL(string: resourcePath, relativeTo: baseResourceURL)!
+      guard let url = URL(string: resourcePath, relativeTo: baseResourceURL) else {
+        assertionFailure("This uses all hardcoded components so it should not fail unless we made a mistake somewhere")
+        return nil
+      }
+      
       var headers = [String: String]()
       if let servicesKeyValue = Bundle.main.getPlistString(for: self.servicesKeyName) {
         headers[self.servicesKeyHeaderValue] = servicesKeyValue
--- a/Client/WebFilters/AdblockRustEngine.swift
+++ b/Client/WebFilters/AdblockRustEngine.swift
@@ -7,6 +7,32 @@ import Foundation
 import BraveCore
 
 extension AdblockEngine {
+  public enum ResourceType: String {
+    case xmlhttprequest
+  }
+  
+  /// Check the rust engine if the request should be blocked given the `sourceURL` and `resourceType`.
+  ///
+  /// - Warning: You must provide a absolute URL (i.e. containing a host) fo r `requestURL` and `sourceURL`
+  public func shouldBlock(requestURL: URL, sourceURL: URL, resourceType: ResourceType) -> Bool {
+    // Compare the etld+1 of requestURL and sourceURL.
+    // Note: `baseDomain` returns etld+1
+    let isThirdParty = requestURL.baseDomain != sourceURL.baseDomain
+    
+    guard let requestHost = requestURL.host, let sourceHost = sourceURL.host else {
+      assertionFailure("You must provide absolute paths for `requestURL` and `sourceURL`")
+      return false
+    }
+    
+    return matches(
+      url: requestURL.absoluteString,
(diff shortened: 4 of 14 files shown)


Client/WebFilters/AdblockRustEngine.swift
```diff
@@ -7,6 +7,29 @@ import Foundation
...
 extension AdblockEngine {
+  public enum ResourceType: String {
+    case script, xhr
+  }
+  
+  /// Check the rust engine if the request should be blocked given the `sourceURL` and `resourceType`.
+  ///
+  /// - Warning: You must provide a absolute URL (i.e. containing a host) fo r `requestURL` and `sourceURL`
+  public func shouldBlock(requestURL: URL, sourceURL: URL, resourceType: ResourceType) -> Bool {
+    let isThirdParty = requestURL.baseDomain != sourceURL.baseDomain
```
pes10k, line 20: if `baseDomain` is eTLD+1, please leave a comment saying so, bc otherwise this will look like an origin check to readers and seem very wrong :)

Client/WebFilters/AdblockRustEngine.swift
```diff
@@ -7,6 +7,29 @@ import Foundation
...
+    
+    guard let requestHost = requestURL.host, let sourceHost = sourceURL.host else {
+      return false
```
pes10k, line 29: weren't you going to assert here?

Client/Frontend/UserContent/UserScripts/RequestBlocking.js
```diff
@@ -0,0 +1,100 @@
...
+    //[DEBUG]console.log(`[RB] URL: '${url}'`)
+    
+    const originalSend = this.send
+    this.send = function() {
+      sendMessage(url, 'xhr').then(blocked => {
+        this._blocked = blocked
+
+        if (blocked) {
+          //[DEBUG]console.log('[RB] BLOCKED: ' + url)
+          this.onerror(Error(`Content blocked for ${url}`))
```
pes10k: you should check that `onerror` is set.  You also should not call `Error` but create something like `new ProgressEvent('error')`

You also should handle traditional event handers.

So. instead of `this.onerror(Error...`, you should do `this.dispatchEvent(new ProgressEvent('error'))`

Client/Frontend/UserContent/UserScripts/RequestBlocking.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+  
+  const originalOpen = XMLHttpRequest.prototype.open
+  XMLHttpRequest.prototype.open = function() {
+    const sendArguments = arguments
+    originalOpen.apply(this, arguments)
+
+    // Extract the url string
+    let urlString = arguments[1]
+    if (!(typeof urlString === 'string' || urlString instanceof String)) {
```
pes10k: why the typechecking here?  What else are you expecting than a string?  And why check for string twice?

Client/Frontend/UserContent/UserScripts/RequestBlocking.js
```diff
@@ -0,0 +1,100 @@
...
+
+    let isAsync = arguments[2]
+    if (isAsync !== undefined && !isAsync) {
+      // We can't handle non-async requests as they are not handled via callbacks
+      return
+    }
+
+    // Resolve the url in case it's a relative path
+    const url = makeURLFromString(urlString)
+    if (url === undefined) { return }
```
pes10k: why return `undefined`?  This seems like invalid input.  `throw`ing seems right here, instead of silent failure

Client/Frontend/UserContent/UserScripts/RequestBlocking.js
```diff
@@ -0,0 +1,100 @@
...
+      //[DEBUG]console.log('[RB] Cannot extract url')
+      return originalFetch.apply(this, arguments)
+    }
+
+    // Resolve the url in case it's a relative path
+    const url = makeURLFromString(urlString)
+    if (url === undefined) { return }
+    //[DEBUG]console.log(`[RB] URL: '${url}'`)
+    
+    return sendMessage(url, 'script').then(blocked => {
```
pes10k: why `script`?  This should be `fetch` no?

Client/Frontend/UserContent/UserScripts/RequestBlocking.js
```diff
@@ -0,0 +1,100 @@
+// Copyright 2022 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+((args) => {
+  "use strict";
+  //[DEBUG]console.log('[RB] Loaded request blocking')
```
pes10k: please remove debugging code

Client/Frontend/UserContent/UserScripts/RequestBlocking.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+  }
+  
+  const { fetch: originalFetch } = window
+  window.fetch = function() {
+    const [resource, config] = arguments
+    
+    // Extract the url string
+    let urlString
+    if (typeof resource === 'string' || resource instanceof String) {
```
pes10k: Same question here, why all the type checking for all these values?  Why the uncertainty about data types?

Client/Frontend/UserContent/UserScripts/RequestBlocking.js
```diff
@@ -0,0 +1,100 @@
...
+      // Simple send where resource is a string
+      urlString = resource
+    } else if (resource !== undefined && (typeof resource.url === 'string' || resource.url instanceof String)) {
+      // Assume its a Request
+      urlString = resource.url
+    } else {
+      //[DEBUG]console.log('[RB] Cannot extract url')
+      return originalFetch.apply(this, arguments)
```
pes10k, line 34: Under what circumstances would this happen?

Client/Frontend/UserContent/UserScripts/RequestBlocking.js
pes10k: my error, should be `xmlhttprequest`

pes10k: https://help.eyeo.com/en/adblockplus/how-to-write-filters#options

Client/Frontend/UserContent/UserScripts/RequestBlocking.js
```diff
@@ -0,0 +1,75 @@
...
+      } else {
+        return originalFetch.apply(this, arguments)
+      }
+    })
+  }
+  
+  const originalOpen = XMLHttpRequest.prototype.open
+  XMLHttpRequest.prototype.open = function() {
+    const sendArguments = arguments
+    originalOpen.apply(this, arguments)
```
pes10k: Note that the return value here is needed. https://developer.mozilla.org/en-US/docs/Web/API/Window/open#return_value

Client/Frontend/UserContent/UserScripts/RequestBlocking.js
```diff
@@ -0,0 +1,75 @@
...
+
+    // Extract the url string
+    let urlString = arguments[1]
+    let isAsync = arguments[2]
+    if (isAsync !== undefined && !isAsync) {
+      // We can't handle non-async requests as they are not handled via callbacks
+      return
```
pes10k: Assert?  I dont think there are any cases where window.open returns undefined

Client/Frontend/UserContent/UserScripts/RequestBlocking.js
```diff
@@ -0,0 +1,75 @@
...
+    const url = new URL(urlString, window.location.href)
+    const originalSend = this.send
+    this.send = function() {
+      sendMessage(url).then(blocked => {
+        this._blocked = blocked
+
+        if (blocked) {
+          this.dispatchEvent(new ProgressEvent('error'))
+        } else {
+          originalSend.apply(this, arguments)
```
pes10k: This value too needs be returned to the original caller 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply#return_value

Client/WebFilters/CosmeticFiltersResourceDownloader.swift
```diff
@@ -308,8 +308,6 @@ public class CosmeticFiltersResourceDownloader {
           into: engine,
           data: $0.resource.data,
           id: $0.type.identifier)
-      case .tgz:
```
thypon, line 311: Is the desktop version supporting `.tgz` compressed files for cosmetic filters and rules?

Client/WebFilters/AdblockerType.swift
```diff
@@ -8,56 +8,55 @@ import Shared
...
-    default: return nil
-    }
-  }
-
   var associatedFiles: [FileType] { return [.json, fileForStatsLibrary] }
 
   private var fileForStatsLibrary: FileType {
     switch self {
     case .general, .regional: return .dat
-    case .httpse: return .tgz
```
thypon, line 31: here, too. Is it HTTPSEverywhere right? Is it dropped?

Client/WebFilters/AdblockResourceDownloader.swift
```diff
@@ -25,7 +25,15 @@ public class AdblockResourceDownloader {
...
   private let servicesKeyHeaderValue = "BraveServiceKey"
 
-  static let endpoint = "https://adblock-data.s3.brave.com/ios"
+  /// The base s3 environment url that hosts the debouncing (and other) files.
+  /// Cannot be used as-is and must be combined with a path
+  private lazy var baseResourceURL: URL = {
+    if AppConstants.buildChannel.isPublic {
+      return URL(string: "https://adblock-data.s3.brave.com")!
+    } else {
+      return URL(string: "https://adblock-data-staging.s3.bravesoftware.com")!
```
thypon, line 34: Isn't this string gonna be embed in the production build here?

Client/Frontend/UserContent/UserScripts/RequestBlocking.js
```diff
@@ -0,0 +1,77 @@
+// Copyright 2022 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+window.braveBlockRequests = (args) => {
+  "use strict";
+  const securityToken = args.securityToken
```
thypon, line 9: pls include `;`. Javascript parser is not too eager of missing `;`.

https://doc.casthighlight.com/alt_missingsemicolon-always-use-semicolons/

Client/Frontend/UserContent/UserScripts/RequestBlocking.js
```diff
@@ -0,0 +1,77 @@
...
+      securityToken: securityToken,
+      resourceURL: resourceURL.href,
+      sourceURL: window.location.href,
+      resourceType: 'xmlhttprequest'
+    })
+  }
+  
+  const { fetch: originalFetch } = window
+  window.fetch = function() {
+    const [resource] = arguments
```
thypon, line 23: here too, missing semicolon

Client/Frontend/UserContent/UserScripts/RequestBlocking.js
```diff
@@ -0,0 +1,77 @@
...
+    
+    // Extract the url
+    let urlString
```
thypon, line 26: and here

thypon (approved):

Client/WebFilters/AdblockerType.swift
iccub: this endpoint was never used for HTTPSE, we use a static list on iOS 14, and on iOS 15 we use Apple's built in httpse

Client/Frontend/Browser/User Scripts/UserScriptHelper.swift
```diff
@@ -83,4 +83,14 @@ class UserScriptHelper {
...
   }
+
+  static func makeRequestBlockingParams(securityToken: String) throws -> String {
+    let params: [String: String] = [
+      "securityToken": securityToken,
+      "handlerName": RequestBlockingContentHelper.scriptMessageHandlerName()
+    ]
+    let encoder = JSONEncoder()
+    let data = try encoder.encode(params)
+    return String(data: data, encoding: .utf8)!
```
iccub: Better to unwrap it and do not continue if this function returns nil

Client/WebFilters/ShieldStats/Adblock/FifoDict.swift
```diff
@@ -3,13 +3,13 @@
 
 import Foundation
 
-class FifoDict {
+class FifoDict<Element> {
```
iccub, line 6: nice generics improvement

iccub (changes requested):

Client/Frontend/UserContent/UserScripts/RequestBlocking.js
```diff
@@ -0,0 +1,84 @@
...
+  const originalSend = XMLHttpRequest.prototype.send
+  XMLHttpRequest.prototype.send = function () {
+    if (this._resourceURL === undefined) {
+      return originalSend.apply(this, arguments)
+    }
+    
+    sendMessage(this._resourceURL).then(blocked => {
+      if (blocked) {
+        Object.defineProperties(this, {
+          readyState: { value: 4 }
```
Brandon-T, line 73: Why do we define this property on the promise or the send function (not sure which `this` it applies to)?

Client/WebFilters/AdblockResourceDownloader.swift
```diff
@@ -102,22 +110,19 @@ public class AdblockResourceDownloader {
...
-        var url = URL(string: AdblockResourceDownloader.endpoint)
-      else {
+      guard let resourcePath = type.resourcePath(for: fileType) else {
         return nil
       }
 
-      url.appendPathComponent(resourceName)
-      url.appendPathExtension(fileExtension)
-
+      let url = URL(string: resourcePath, relativeTo: baseResourceURL)!
```
Brandon-T: Guard here instead of force unwrapping.

Client/WebFilters/ContentBlocker/RequestBlockingContentHelper.swift
```diff
@@ -0,0 +1,86 @@
...
+    struct RequestBlockingDTOData: Decodable, Hashable {
+      enum CodingKeys: String, CodingKey {
+        case resourceURL, sourceURL, resourceType
+      }
+      
+      let resourceType: AdblockEngine.ResourceType
+      let resourceURL: URL
+      let sourceURL: URL
+      
+      init(from decoder: Decoder) throws {
```
Brandon-T: This method can be removed entirely if `AdblockEngine.ResourceType` is codable.

Client/WebFilters/ContentBlocker/RequestBlockingContentHelper.swift
```diff
@@ -0,0 +1,86 @@
...
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let resourceURLString = try container.decode(String.self, forKey: .resourceURL)
+        let sourceURLString = try container.decode(String.self, forKey: .sourceURL)
+        let resourceTypeString = try container.decode(String.self, forKey: .resourceType)
```
Brandon-T: There is no need to decode as a string first, then create a `URL` from it.
You can do: `resourceURL = try container.decode(URL.self, forKey: .resourceURL)` for example.
It will decode `URL` which is `Codable` :)
This way, you can get rid of all the `DecodingError.dataCorrupted` stuff.

```swift
import Foundation

let json = """
{
  "url": "https://google.ca"
}
"""

struct Foo: Codable {
  let url: URL
}

let data = json.data(using: .utf8)!
let foo = try JSONDecoder().decode(Foo.self, from: data)

print(foo.url)
```

If you want to test in a playground.

Client/WebFilters/ShieldStats/Adblock/AdBlockStats.swift
```diff
@@ -37,6 +37,40 @@ public class AdBlockStats: LocalAdblockResourceProtocol {
     parseBundledGeneralBlocklist()
     loadDownloadedDatFiles()
   }
+  
+  /// Checks the general and regional engines to see if the request should be blocked.
+  ///
+  /// - Note: This method is should not be synced on `AdBlockStatus.adblockSerialQueue` and the result is synced on the main thread.
+  func shouldBlock(requestURL: URL, sourceURL: URL, resourceType: AdblockEngine.ResourceType, callback: @escaping (Bool) -> Void) {
+    Self.adblockSerialQueue.async { [weak self] in
+      let shouldBlock = self?.shouldBlock(requestURL: requestURL, sourceURL: sourceURL, resourceType: resourceType)
```
Brandon-T: `let shouldBlock = self?.shouldBlock(....) == true`
This way you don't have to unwrap `?? false`

Same for all the calls to `regionalAdblockEngine?.shouldBlock`

Brandon-T (changes requested):

iccub (approved):

Brandon-T (approved):

PR #5552, opened 2022-06-20
Fix #5551: Update `window.ethereum` to be non-configurable

### LINKED ISSUE #5551: Update window.ethereum to be non-configurable
### Description: 

ref: https://github.com/brave/security/issues/887#issuecomment-1159332335

StephenHeaps:
## Summary of Changes
- Define `window.ethereum` as non-configurable (defaults to false)

This pull request fixes #5551

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/UserContent/UserScripts/WalletEthereumProvider.js
+++ b/Client/Frontend/UserContent/UserScripts/WalletEthereumProvider.js
@@ -24,7 +24,6 @@ if (window.isSecureContext) {
   }
   
   Object.defineProperty(window, 'ethereum', {
-    configurable: true,
     value: {
       chainId: undefined,
       networkVersion: undefined,


iccub (approved):

Brandon-T:
Does it also need to be `readonly` and `non-enumerable`? IE: `writable: false`
Asking because on Desktop it is configurable, writable, and enumerable: 
image

All other iOS injections are neither.

StephenHeaps:
> Does it also need to be `readonly` and `non-enumerable`? IE: `writable: false` Asking because on Desktop it is configurable, writable, and enumerable: image
> 
> All other iOS injections are neither.

Desktop defines each of `enumerable`, `configurable`, `writable` [to be true here](https://github.com/brave/brave-core/blob/0d066d53ebe590bf1b58ab9f676c279499e74106/components/brave_wallet/resources/ethereum_provider.js#L51) so MetaMask/other extensions can overwrite it. Since we don't support extensions on iOS, I believe we can keep these as disabled.

Brandon-T (approved):

nuo-xu (approved):

PR #5712, opened 2022-07-18
Fix #4805: Detect and store VPN state after sign-in

### LINKED ISSUE #4805: Detect and store VPN state after sign-in

iccub:
## Summary of Changes

This pull request fixes #4805 

Security review: https://github.com/brave/security/issues/1024

## Submitter Checklist:

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Client/Frontend/Browser/BraveSkusScriptHandler.swift
```diff
@@ -0,0 +1,144 @@
...
+    sku?.prepareCredentialsPresentation(domain, path: path) { [weak self] credential in
+      if !credential.isEmpty {
+        if let vpnCredential = BraveSkusWebHelper.fetchVPNCredential(credential, domain: domain) {
+          Preferences.VPN.skusCredential.value = credential
+          Preferences.VPN.skusCredentialDomain.value = domain
+          
+          BraveVPN.setCustomVPNCredential(vpnCredential.credential, environment: vpnCredential.environment)
+        }
+      } else {
+        //assertionFailure()
```
nuo-xu: maybe a log for this case?

Client/Frontend/Browser/BraveSkusScriptHandler.swift
```diff
@@ -0,0 +1,144 @@
...
+    }
+  }
+  
+  private func handleRefreshOrder(for orderId: String, domain: String, replyHandler: @escaping ReplyHandler) {
+    sku?.refreshOrder(domain, orderId: orderId) { [weak self] completion in
+      do {
+        guard let data = completion.data(using: .utf8) else { return }
+        let json = try JSONSerialization.jsonObject(with: data, options: .fragmentsAllowed)
+        log.debug("skus refreshOrder")
+        self?.callback(methodId: 1, result: json, replyHandler: replyHandler)
```
nuo-xu: is here all the method Id is the `Method`? maybe we should use enum's rawValue for code readability

Client/Frontend/Browser/BraveSkusWebHelper.swift
```diff
@@ -82,6 +82,50 @@ class BraveSkusWebHelper {
...
+  }
+  
+  static func environment(domain: String) -> String? {
+    switch domain {
+    case "account.brave.software", "vpn.brave.software": return "development"
+    case "account.bravesoftware.com", "vpn.bravesoftware.com": return "staging"
+    case "account.brave.com", "vpn.brave.com": return "production"
+    default:
+      return nil
+    }
```
nuo-xu, line 102: convert environment string to a string enum? its very easy to make mistakes typing strings. at least for me lol

Sources/BraveShared/BraveStrings.swift
```diff
@@ -2451,9 +2451,9 @@ extension Strings {
         comment: "Table cell title for cell that shows when the VPN subscription expires.")
 
     public static let settingsManageSubscription =
-      NSLocalizedString("vpn.settingsManageSubscription", tableName: "BraveShared", bundle: .strings,
-        value: "Manage Subscription",
-        comment: "Button to manage your VPN subscription")
+          NSLocalizedString("vpn.settingsManageSubscription", tableName: "BraveShared", bundle: .strings,
+            value: "Manage Subscription",
+            comment: "Button to manage your VPN subscription")
```
nuo-xu: looks like wrong indentation

iccub:
@nuo-xu all your comments should be resolved. I simplified the callbacks as well

Client/Frontend/Browser/BraveSkusScriptHandler.swift
nuo-xu: here still didnt change? can we do
self?.callback(methodId: Method.refreshOrder.rawValue, result: json, replyHandler: replyHandler)

Sources/BraveShared/BraveStrings.swift
```diff
@@ -2457,9 +2457,9 @@ extension Strings {
         comment: "Table cell title for cell that shows when the VPN subscription expires.")
 
     public static let settingsManageSubscription =
-      NSLocalizedString("vpn.settingsManageSubscription", tableName: "BraveShared", bundle: .strings,
-        value: "Manage Subscription",
-        comment: "Button to manage your VPN subscription")
+          NSLocalizedString("vpn.settingsManageSubscription", tableName: "BraveShared", bundle: .strings,
+          value: "Manage Subscription",
+          comment: "Button to manage your VPN subscription")
```
nuo-xu: The indentation still wrong tho

nuo-xu (approved):

Sources/BraveShared/BraveStrings.swift
iccub: will fix, thanks, xcode indents it completely different compare to other strings here

Client/Frontend/UserContent/UserScripts/BraveSkusHelper.js
```diff
@@ -0,0 +1,36 @@
...
+    return webkit.messageHandlers.braveSkusHelper.postMessage({'securitytoken': '$<security_token>' ,'method_id': method_id, data: data});
+  }
+
+  if (!window.chrome) {
+    window.chrome = {};
+  }
+
+Object.defineProperty(window.chrome, 'braveSkus', {
+enumerable: false,
+configurable: true,
```
Brandon-T: Any reason why this is `configurable`?
The reason I ask is because a page can do:
```javascript
const braveSkus = window.chrome.braveSkus;

Object.defineProperty(window.chrome, 'braveSkus', {
enumerable: false,
configurable: false,
writable: false, 
value: {
    refresh_order(orderId) {
        console.log(orderId);
        return braveSkus.refresh_order(orderId).then(result => {
            console.log(result)
        })
    },
}
});
```
Same for the other functions. Though, it seems this is injected only on `accounts.bravesoftware.com` anyway, so should be fine.
Just curious if it can be `non-configurable` anyway.

Brandon-T (changes requested):

Brandon-T (approved):

Client/Frontend/Browser/BraveSkusScriptHandler.swift
```diff
@@ -0,0 +1,141 @@
...
+  func scriptMessageHandlerName() -> String? { "braveSkusHelper" }
+  
+  private enum Method: Int {
+    case refreshOrder = 1
+    case fetchOrderCredentials = 2
+    case prepareCredentialsPresentation = 3
+    case credentialsSummary = 4
+  }
+  
+  func userContentController(_ userContentController: WKUserContentController, didReceiveScriptMessage message: WKScriptMessage, replyHandler: @escaping (Any?, String?) -> Void) {
```
stoletheminerals, line 38: We pass `securitytoken` but don't check it here. It's ok since we have `allowedHosts.contains(requestHost)` that prevents an arbitrary page from calling the handler, but for clarity, we should either not pass the token or implement relevant checks.

Client/Frontend/UserContent/UserScripts/BraveSkusHelper.js
```diff
@@ -0,0 +1,36 @@
+// Copyright 2022 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+'use strict';
+
+window.__firefox__.includeOnce("BraveSkusHelper", function() {
+  function sendMessage(method_id, data) {
+    return webkit.messageHandlers.braveSkusHelper.postMessage({'securitytoken': '$<security_token>' ,'method_id': method_id, data: data});
```
stoletheminerals, line 10: Not necessary, but we could use this value https://github.com/brave/brave-ios/blob/50bf289676a1bcbac25db84a4365fec9a5d4c2d4/Client/Frontend/Browser/User%20Scripts/ScriptFactory.swift#L120 to randomize the handler name for obscurity

stoletheminerals (commented):
Few comments/nitpicks. Overall lgtm

Client/Frontend/UserContent/UserScripts/BraveSkusHelper.js
iccub: @Brandon-T suggested to me it's fine to remove this part. Can you post your reasoning here Brandon?
I think with the new webkit callbacks we use it's not needed anymore

Client/Frontend/Browser/BraveSkusScriptHandler.swift
iccub: good catch, i missed it

iccub: Added here https://github.com/brave/brave-ios/pull/5712/commits/235c59c5f6a165e2bf5d66bf90e6f9eb0642a3cd

Client/Frontend/UserContent/UserScripts/BraveSkusHelper.js
Brandon-T: It won't make a difference because this script is only injected on `accounts.brave.com` which is a page owned by us.
Also, atm, pages can override `[[redacted]]` and so they can already retrieve the handler and the token.
I'm going to be fixing all of these already, but this one in particular is fine IMO because we own the page, and it will never be injected onto another page.

CC: @stoletheminerals

stoletheminerals: @Brandon-T we inject the script only on *.brave.com hosts, but the handler is available on any page (as I mentioned on Slack)
I can call `webkit.messageHandlers.braveSkusHelper.postMessage()` on localhost (or any host) and it will execute. This is also true for all webkit handlers we expose in `.page` world, the fact that we inject javascript into particular domains doesn't limit the accessibility of the webkit handlers.
In this particular case, this is not an issue because we have a host check here:
https://github.com/brave/brave-ios/blob/235c59c5f6a165e2bf5d66bf90e6f9eb0642a3cd/Client/Frontend/Browser/BraveSkusScriptHandler.swift#L42

But in general, we can't rely on where we inject .js files.

stoletheminerals: ^ This host check is actually a good example of how we should protect all host-related handlers, such as Search, proposed Jitsi SDK helper, etc. Instead of relying on JS obscurity, we should add security checks to the native code.

stoletheminerals (approved):

PR #5719, opened 2022-07-19
Fix 5716: Enabling Solana

### LINKED ISSUE #5716: Wallet: Enable Solana Support for iOS
### Description: 
Once Solana Support is enabled, users can create or import solana account, switch to solana networks, fetch solana account balance and make send transactions with a solana account

nuo-xu:
## Summary of Changes

This pull request fixes #5716 

## Submitter Checklist:

  - [ ] Light & dark mode

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/BraveWallet/WalletConstants.swift
+++ b/BraveWallet/WalletConstants.swift
@@ -40,10 +40,7 @@ struct WalletConstants {
   
   /// The currently supported coin types.
   static var supportedCoinTypes: OrderedSet<BraveWallet.CoinType> {
-    if WalletDebugFlags.isSolanaEnabled {
-      return [.eth, .sol]
-    }
-    return [.eth]
+    return [.eth, .sol]
   }
   
   /// The link for users to learn more about Solana SPL token account creation in transaction confirmation screen
@@ -56,8 +53,4 @@ struct WalletConstants {
   static let supportedBuyWithWyreNetworkChainIds: [String] = [
     BraveWallet.MainnetChainId
   ]
-}
-
-struct WalletDebugFlags {
-  static var isSolanaEnabled: Bool = true
 }
--- a/BraveWallet/WalletConstants.swift
+++ b/BraveWallet/WalletConstants.swift
@@ -59,5 +59,5 @@ struct WalletConstants {
 }
 
 struct WalletDebugFlags {
-  static var isSolanaEnabled: Bool = false
+  static var isSolanaEnabled: Bool = true
 }
--- a/Tests/BraveWalletTests/KeyringStoreTests.swift
+++ b/Tests/BraveWalletTests/KeyringStoreTests.swift
@@ -12,11 +12,6 @@ class KeyringStoreTests: XCTestCase {
   
   private var cancellables: Set<AnyCancellable> = .init()
   
-  override func setUp() {
-    super.setUp()
-    WalletDebugFlags.isSolanaEnabled = true
-  }
-  
   private func setupServices() -> (BraveWallet.TestKeyringService, BraveWallet.TestJsonRpcService, BraveWallet.TestBraveWalletService) {
     let currentNetwork: BraveWallet.NetworkInfo = .mockMainnet
     let currentChainId = currentNetwork.chainId
--- a/Tests/BraveWalletTests/NetworkSelectionStoreTests.swift
+++ b/Tests/BraveWalletTests/NetworkSelectionStoreTests.swift
@@ -18,12 +18,7 @@ import BraveShared
     .sol: [.mockSolana, .mockSolanaTestnet]
   ]
   
-  override func setUp() {
-    super.setUp()
-    WalletDebugFlags.isSolanaEnabled = true
-  }
-  
-  private func setupServices() -> (BraveWallet.TestKeyringService, BraveWallet.TestJsonRpcService, BraveWallet.TestBraveWalletService, BraveWallet.TestSwapService) {
+  private func setupServices() -> (BraveWallet.TestKeyringService, BraveWallet.TestJsonRpcService, BraveWallet.TestBraveWalletService) {
     let currentNetwork: BraveWallet.NetworkInfo = .mockMainnet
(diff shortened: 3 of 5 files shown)


iccub (approved):

BraveWallet/WalletConstants.swift
```diff
@@ -47,5 +47,5 @@ struct WalletConstants {
 }
 
 struct WalletDebugFlags {
-  static var isSolanaEnabled: Bool = false
+  static var isSolanaEnabled: Bool = true
```
iccub: We could remove this flag altogether maybe, sometime in the future

StephenHeaps: I agree, I would suggest dropping this flag entirely (and just include `.sol` in `supportedCoinTypes` array) would be good in this PR.

nuo-xu: removed this flag in [5a674dd](https://github.com/brave/brave-ios/pull/5719/commits/5a674ddab0e3e5b440cc259786a44e874b8fea3a)
re-request the team to review.

iccub (approved):

kdenhartog:
Spotted a minor UI bug
receiving address error bug

StephenHeaps (approved):

nuo-xu:
hey @kdenhartog, we created issue [5773](https://github.com/brave/brave-ios/issues/5773) and it has been fixed in `development`. It fixes the `same address` error message didn't update after selected account changed. We couldn't reproduce the `invalid SOL address` error with the account `AX1fumgqvHhXe2gRAJiQK2d759D4LPXaWjEVDRHeoG6B`. If you can reproduce, please provide us more information. thanks!

kdenhartog:
Ahh sorry @nuo-xu, yeah that was the only bug I was trying to report was the `same address` error. The `invalid SOL address` error was from me double pasting the address in. Looks like you got it all fixed

kdenhartog:
**Summary: I think if we can get these two issues fixed this would be good from my perspective. Will let @stoletheminerals comment on their findings as well**

I've tried the following:

- Rejecting transactions (works as expected)
- Submitting normal transaction (works as expected)
- Submitting to program address (submits to wallet, but chain errors and no funds lost)
- playing around with bad inputs for account porting

---

Issues found:
1. Doesn't appear to be properly detecting all assets in the wallet. When I manually added the asset as viewable then it shows up, but it doesn't show the value properly either.
Screen Shot 2022-08-04 at 6 30 38 PM

---

2. Can add the same account twice. Once as a primary account and once as a secondary account. Steps to repro:
  - copy a private key in a different wallet on your HD path (e.g. account 2) (I used slope finance to export the key)
  - import the key into the wallet by pasting the private key in
  - add an account again without pasting anything in

Screen Shot 2022-08-04 at 6 45 46 PM

stoletheminerals:
Should we put the SOL icon here instead of a basic grey circle? It would be more apparent to the user which asset is selected at the moment. Ideally a colour icon for the mainet and greyscale for test/dev.
image
image

nuo-xu:
@stoletheminerals the icon issue has been fixed in `development` ref: https://github.com/brave/brave-ios/pull/5772

darkdh:
>Can add the same account twice. Once as a primary account and once as a secondary account. Steps to repro:

@kdenhartog this is not iOS specific issue. The thing you are trying to do is import an account which hasn't been derived and we can't derived all possible accounts just to check if the key you are trying to import belongs to derived accounts. We only check what users have derived. (https://github.com/brave/brave-core/blob/2699c59acec157de3fd929b7d922b0c2d2e6beca/components/brave_wallet/browser/hd_keyring.cc#L63-L70)

The possible solution would be, when deriving account we check if the account being derived is in import accounts, if so delete the import account. But it will slow down derivation process when import accounts grew.

stoletheminerals:
Overall looks good to me!
Just a note: Since we don't support ENS / SOL domains on iOS I think we should remove the "URL" part from the "To" field on both Solana and Ethereum windows for now.

kdenhartog:
> @kdenhartog this is not iOS specific issue. The thing you are trying to do is import an account which hasn't been derived and we can't derived all possible accounts just to check if the key you are trying to import belongs to derived accounts. We only check what users have derived. ([brave/brave-core@2699c59/components/brave_wallet/browser/hd_keyring.cc#L63-L70](https://github.com/brave/brave-core/blob/2699c59acec157de3fd929b7d922b0c2d2e6beca/components/brave_wallet/browser/hd_keyring.cc#L63-L70))

Ok thanks for clarifying @darkdh that makes sense to me why this is possible then. I can see how that lookup would have tradeoffs here as the derivation gets larger so at this point, given I couldn't produce any unintended side effects through this, I'm good to leave this as is.

kdenhartog (approved):
Issue 1 from https://github.com/brave/brave-ios/pull/5719#issuecomment-1204830858 isn't a security vulnerability so not going to block the review on this. Approving I think this one is good to go from a sec perspective

stoletheminerals:
Just noticed: I can't choose my second Solana account on Send window:
image
Clicking the pointing down arrow should let me choose my second Solana account, but instead, I can only copy the address of my first account.
UPD: I can't switch Ethereum accounts as well

StephenHeaps:
> Just noticed: I can't choose my second Solana account on Send window:
> image
> Clicking the pointing down arrow should let me choose my second Solana account, but instead, I can only copy the address of my first account.
> UPD: I can't switch Ethereum accounts as well

@stoletheminerals are you on iOS 16 beta? If so it’s a known issue:
https://github.com/brave/brave-ios/issues/5618

Hopefully can have this fixed soon.

stoletheminerals:
@StephenHeaps yep, on 16 beta. Didn't see that issue, my bad!

nuo-xu:
fyi @kdenhartog issue 1* https://github.com/brave/brave-ios/issues/5779 is also resolved.

PR #5772, opened 2022-08-02
Fix #5769: Display correct native asset icon

### LINKED ISSUE #5769: Wallet: Display correct Solana native asset logo

nuo-xu:
Bundle some network's native assets' logos
For native token that is created by ourselves, we will insert the correct token logo.
For all token we will still try to load image using token's logo and token's symbol under `tokenLogoBaseURL`, if it fails we will use check if the token is native asset so we fetch the bundle native token icon. if it also fails, we will check if logo is a web image. if all fails, the monogram fallback image will be rendered. 

This pull request fixes #5769 

## Submitter Checklist:

  - [ ] Light & dark mode

## Test Plan:
Switch between different networks to see if network's native asset's icon is being displayed. 

## Screenshots:
Simulator Screen Shot - iPhone Xʀ - 2022-08-02 at 16 02 56

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/BraveWallet/Crypto/AssetIconView.swift
+++ b/BraveWallet/Crypto/AssetIconView.swift
@@ -41,8 +41,8 @@ struct AssetIconView: View {
     
     if network.isNativeAsset(token) {
       // check bundled images
-      if UIImage(named: network.nativeTokenLogo, in: .current, with: nil) != nil {
-        return Image(network.nativeTokenLogo, bundle: .current)
+      if let uiImage = UIImage(named: network.nativeTokenLogo, in: .current, with: nil) {
+        return Image(uiImage: uiImage)
       }
     }
     
--- a/BraveWallet/Crypto/AssetIconView.swift
+++ b/BraveWallet/Crypto/AssetIconView.swift
@@ -39,9 +39,9 @@ struct AssetIconView: View {
       }
     }
     
-    if network.isNativeAsset(token) {
+    if network.isNativeAsset(token), let logo = network.nativeTokenLogo {
       // check bundled images
-      if let uiImage = UIImage(named: network.nativeTokenLogo, in: .current, with: nil) {
+      if let uiImage = UIImage(named: logo, in: .current, with: nil) {
         return Image(uiImage: uiImage)
       }
     }
--- a/BraveWallet/Crypto/AssetIconView.swift
+++ b/BraveWallet/Crypto/AssetIconView.swift
@@ -17,6 +17,7 @@ import BraveCore
 ///
 struct AssetIconView: View {
   var token: BraveWallet.BlockchainToken
+  var network: BraveWallet.NetworkInfo
   @ScaledMetric var length: CGFloat = 40
 
   private var fallbackMonogram: some View {
@@ -37,6 +38,14 @@ struct AssetIconView: View {
         return Image(uiImage: image)
       }
     }
+    
+    if network.isNativeAsset(token) {
+      // check bundled images
+      if UIImage(named: network.nativeTokenLogo, in: .current, with: nil) != nil {
+        return Image(network.nativeTokenLogo, bundle: .current)
+      }
+    }
+    
     return nil
   }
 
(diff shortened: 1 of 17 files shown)


BraveWallet/Crypto/AssetIconView.swift
```diff
@@ -37,6 +38,14 @@ struct AssetIconView: View {
         return Image(uiImage: image)
       }
     }
+    
+    if network.isNativeAsset(token) {
+      // check bundled images
+      if UIImage(named: network.nativeTokenLogo, in: .current, with: nil) != nil {
+        return Image(network.nativeTokenLogo, bundle: .current)
```
StephenHeaps: I think this should just be an `if let` and use the `Image(uiImage:)` initializer for `Image` so we don't have to load the image twice

nuo-xu: fixed in [7c8ecca](https://github.com/brave/brave-ios/pull/5772/commits/7c8ecca2ff5e599880a10a90dff9cfadcaffc4d9)

BraveWallet/Extensions/BraveWalletSwiftUIExtensions.swift
```diff
@@ -42,6 +42,30 @@ extension BraveWallet.NetworkInfo: Identifiable {
...
+    } else if chainId.caseInsensitiveCompare(BraveWallet.BinanceSmartChainMainnetChainId) == .orderedSame {
+      return "bnb-asset-icon"
+    } else if chainId.caseInsensitiveCompare(BraveWallet.CeloMainnetChainId) == .orderedSame {
+      return "celo"
+    } else if chainId.caseInsensitiveCompare(BraveWallet.AvalancheMainnetChainId) == .orderedSame {
+      return "avax"
+    } else if chainId.caseInsensitiveCompare(BraveWallet.FantomMainnetChainId) == .orderedSame {
+      return "fantom"
+    } else if chainId.caseInsensitiveCompare(BraveWallet.OptimismMainnetChainId) == .orderedSame {
+      return "optimism" // should never happen since OptimismMainnet's symbol is ETH
```
StephenHeaps: If Optimism Mainnet's native token is ETH, should this be included here at all? The property is `nativeTokenLogo` and this is not the native token logo, could cause a mistake later on.

BraveWallet/Extensions/BraveWalletSwiftUIExtensions.swift
```diff
@@ -42,6 +42,30 @@ extension BraveWallet.NetworkInfo: Identifiable {
...
+    } else {
+      return iconUrls.first ?? ""
```
StephenHeaps: Kind of seems like we should use an optional String for this property, otherwise may have to check against `isEmpty` / empty string later?

StephenHeaps (approved):
Couple nits but resolves user assets missing logo / icons for CELO & SOL 👍

BraveWallet/Extensions/BraveWalletSwiftUIExtensions.swift
nuo-xu: i can remove this if condition since the first should already capture Optimism Mainnet's symbol is ETH. I put there to align the logic with desktop's but i think thats desktop's bug :p

BraveWallet/Extensions/BraveWalletSwiftUIExtensions.swift
nuo-xu: fixed in [1bc5c0f](https://github.com/brave/brave-ios/pull/5772/commits/1bc5c0fb0fc17be231c07cdd3804706b9c141410)

BraveWallet/Extensions/BraveWalletSwiftUIExtensions.swift
nuo-xu: fixed in [1bc5c0f](https://github.com/brave/brave-ios/pull/5772/commits/1bc5c0fb0fc17be231c07cdd3804706b9c141410)

PR #5839, opened 2022-08-11
Fix #1358: Cookie consent blocking

### LINKED ISSUE #1358: Cookie consent options
### Description: 
In the EU, almost every website displays a pop up asking users to agree to their cookies. As a user this is v annoying. 
There are many product that offer this consent management, one being Quantcast Choice (https://www.quantcast.com/gdpr/consent-management-solution/)

Can we have an option in Privacy that allows users to not see these pop-ups?
The text can be something like: 
Hide Cookie consent pop-ups

cc @johnnyryan and @tomlowenthal for approval

cuba:
## Summary of Changes
This will add a setting to Shields to enable cookie consent notices
It will also show a popup on second launch to allow the user to know this feature is available

This pull request fixes #1358 

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:

Make sure that 
### When the user clicks on "Yes block cookie consent notices" on the popup:
- [ ] The setting is enabled in shields
- [ ] The cookie notice blocking filter list is enabled
- [ ] The animation plays and then the popup disappears
### When the user clicks on "No thanks" on the popup
- [ ] The setting is not enabled in shields
- [ ] The cookie notice blocking filter list is not enabled
- [ ] The popup is dismissed
### When the setting is enabled:
- [ ] The popup doesn't appear on second launch (Note: you will need to enabled it on first launch)
- [ ] The cookie consent notices are blocked
### When the setting is not enabled:
- [ ] The popup appears on a second launch (note that depending on timing other full screen popups might appear in priority)
- [ ] Cookie consent notices are not blocked
### Persistence
- [ ] Test for synchronization between the cookie blocking setting in shields and the filter list is enabled in filter lists screen
- [ ] Test for persistence (setting is held) when the app is terminated

## Screenshots:

Simulator Screen Shot - iPhone 13 Pro - 2022-08-11 at 12 35 06
Simulator Screen Shot - iPhone 13 Pro - 2022-08-11 at 12 34 57
Simulator Screen Shot - iPhone 13 Pro - 2022-08-11 at 12 34 24

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


IronMaltese:
I feel like the realistic cookie on the pop up doesn't match with Brave's overall theme, and should instead be stylized differently

Client/Frontend/Browser/BrowserViewController/BrowserViewController+CookieNotificationBlocking.swift
```diff
@@ -0,0 +1,51 @@
...
+    // Don't show this if we already enabled the setting
+    guard !FilterListResourceDownloader.shared.isEnabled(filterListUUID: FilterListResourceDownloader.cookieConsentNoticesUUID) else { return }
+    // Don't show this if we are presenting another popup already
+    guard !isOnboardingOrFullScreenCalloutPresented else { return }
+    // We only show the popup on second launch
+    guard !Preferences.General.isFirstLaunch.value else { return }
+    
+    #if DEBUG
+    // We reset this value on debug. If you don't want this popup to appear on debug, just enable the setting and it won't appear
+    FullScreenCalloutManager.enable(for: .blockCookieConsentNotices)
```
soner-yuksel: Why do we reset this preference on debug? This will show all the time and like you mentioned we have to go toggle that setting to prevent this.

It might be better to use one of the `DebugFlag` in Preferences like we do in all the callouts and pop-ups or create a new skip flag and use ti same way.

Client/Frontend/Settings/BraveShieldsAndPrivacySettingsController.swift
```diff
@@ -238,6 +279,22 @@ class BraveShieldsAndPrivacySettingsController: TableViewController {
   }()
 
   // MARK: - Actions
+  
+  private func makeCookieConsentBlockingRow() -> Row {
+    return .boolRow(
+      uuid: self.cookieConsentNoticesRowUUID,
```
soner-yuksel, line 300: Do we need the UUID here we are enabling preference value with astatic UUID here anyways ?

Client/Frontend/Settings/BraveShieldsAndPrivacySettingsController.swift
```diff
@@ -238,6 +279,22 @@ class BraveShieldsAndPrivacySettingsController: TableViewController {
...
+      title: Strings.blockCookieConsentNotices,
+      detailText: nil,
```
soner-yuksel: nit: Default is nil can be removed

Client/Frontend/Widgets/GIFImage.swift
```diff
@@ -0,0 +1,179 @@
...
+      
+      return (0..<frameCount).map { index2 -> UIImage in
+        return frame
+      }
+    }
+  }
+  
+  private class func gcd(_ a: Int, _ b: Int) -> Int {
+    let absB = abs(b)
+    let r = abs(a) % absB
```
soner-yuksel: nit: Can we use one word variables instead of single letters ?

Client/WebFilters/FilterListResourceDownloader.swift
```diff
@@ -157,6 +167,22 @@ public class FilterListResourceDownloader: ObservableObject {
     _ = await (cachedFilterLists, cachedDefaultFilterList)
   }
   
+  /// Enables a filter list for the given uuid. Returns true if the filter list exists or not.
+  public func enableFilterList(forFilterListUUID uuid: String, isEnabled: Bool) -> Bool {
+    // Enable the setting
+    if let index = filterLists.firstIndex(where: { $0.uuid == Self.cookieConsentNoticesUUID }) {
```
soner-yuksel: nit: I believe we dont need to fetch the index first and get the value using the index and just use r

`if var filterList = filterLists.first(where: { $0.uuid == Self.cookieConsentNoticesUUID }) `

Client/Frontend/Browser/BrowserViewController/BrowserViewController+CookieNotificationBlocking.swift
cuba: Yah i'll change this or remove it. It was just easier for testing

Client/WebFilters/FilterListResourceDownloader.swift
cuba: I don't think that works. Filter lists are structs if I get a filter list I will be changing the copy of that struct not the one in the array.

Client/Frontend/Widgets/GIFImage.swift
cuba: sure. I stole this from the internet and didn't do a good job cleaning it up

Client/Frontend/Settings/BraveShieldsAndPrivacySettingsController.swift
cuba: well I need to find and reload the row when the setting changes so I need to pass in a static uuid. Nothing to do with the filter list UUID

Client/WebFilters/FilterListResourceDownloader.swift
soner-yuksel: oh yeah you are totally right, I though you are setting the modified value somewhere.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+CookieNotificationBlocking.swift
```diff
@@ -0,0 +1,51 @@
...
+import BraveUI
+import BraveShared
+import Data
+
+extension BrowserViewController {
+  func presentCookieNotificationBlockingCalloutIfNeeded() {
+    // Don't show this if we already enabled the setting
+    guard !FilterListResourceDownloader.shared.isEnabled(filterListUUID: FilterListResourceDownloader.cookieConsentNoticesUUID) else { return }
+    // Don't show this if we are presenting another popup already
+    guard !isOnboardingOrFullScreenCalloutPresented else { return }
```
soner-yuksel, line 17: Can you add the check for onboarding status here ?
`let basicOnboardingNotCompleted = Preferences.General.basicOnboardingProgress.value != OnboardingProgress.newTabPage.rawValue`

The reason is If the user kills the app in the middle of onboarding and open again. The behaviour should be showing NTP url calllouts but this cookie consent will take precedence there.

soner-yuksel (approved):
All comments are fixed and working as expected 🎉

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,100 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import SwiftUI
+import Strings
+import DesignSystem
+import BraveShared
+
+struct CookieNotificationBlockingConsentView: View {
+  public static let contentHeight: CGFloat = 480
+  private static let topSectionHeight: CGFloat = 328
```
iccub: CGFloat are auto bridged now, you can write it like this
`private static let topSectionHeight = 328.0`

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,100 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import SwiftUI
+import Strings
+import DesignSystem
+import BraveShared
+
+struct CookieNotificationBlockingConsentView: View {
+  public static let contentHeight: CGFloat = 480
```
iccub: Can be private maybe?

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,100 @@
...
+  private let animation = Animation.easeOut(duration: 0.5).delay(0)
+  private let transition = AnyTransition.scale(scale: 1.1).combined(with: .opacity)
+  private let textPadding: CGFloat = 16
+  
+  @Environment(\.presentationMode) var presentationMode
+  @State private var showAnimation = false
+  
+  var body: some View {
+    VStack(spacing: 0) {
+      ZStack {
```
iccub: You can remove this ZStack and use `.overlay` instead

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,100 @@
...
+        }
+      }
+      .frame(height: Self.bottomSectionHeight, alignment: .center)
+    }
+    .background(
+      Image("cookie-consent-background", bundle: .current),
+      alignment: .bottomLeading
+    )
+    .background(Color(UIColor.braveBackground))
+    .padding(0)
```
iccub: Is this zero padding needed?

Client/Frontend/Settings/BraveShieldsAndPrivacySettingsController.swift
```diff
@@ -45,6 +55,32 @@ class BraveShieldsAndPrivacySettingsController: TableViewController {
       manageWebsiteDataSection,
       otherSettingsSection,
     ]
+    
+    // Listen to changes on filter lists so we know we need to reload the section
+    filterListsSubscription = FilterListResourceDownloader.shared.$filterLists
+      .receive(on: DispatchQueue.main)
+      .sink { filterLists in
+        let filterList = filterLists.first(where: { $0.uuid == FilterListResourceDownloader.cookieConsentNoticesUUID })
+        let isEnabled = filterList?.isEnabled ?? false
```
iccub: `let isEnabled = filterList?.isEnabled == true` should be easier to read

Client/Frontend/Widgets/GIFImage.swift
```diff
@@ -0,0 +1,181 @@
...
+      imageView.image = lastFrame
+      imageView.startAnimating()
+    } else {
+      imageView.image = firstFrame
+      imageView.stopAnimating()
+    }
+  }
+  
+  func animationDidStop(_ anim: CAAnimation, finished: Bool) {
+    print(finished)
```
iccub: remove this print or replace with log

Client/Frontend/Widgets/GIFImage.swift
```diff
@@ -0,0 +1,181 @@
+// Copyright 2022 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import UIKit
+import SwiftUI
+
+struct GIFImage: UIViewRepresentable {
```
iccub, line 9: Interesting, we do not use lottie for animations anymore?

Client/WebFilters/FilterListResourceDownloader.swift
```diff
@@ -111,6 +111,16 @@ public class FilterListResourceDownloader: ObservableObject {
...
   }
+  
+  /// The uuid of the cookie consent notices filter list.
+  ///
+  /// This is a special filter list that has more accessible UI to control it
+  public static let cookieConsentNoticesUUID = "AC023D22-AE88-4060-A978-4FEEEC4221693"
+  /// The uuid of the "Remove Web annoyances" filter list
+  ///
+  /// This is a special filter list that has more accessible UI to control it
+  public static let removeWebAnnoyancesUUID = "67E792D4-AE03-4D1A-9EDE-80E01C81F9B8"
```
iccub: we can use `componentID` if possible, apparently the UUIDs are going to be deprecated

Client/WebFilters/FilterListResourceDownloader.swift
iccub: We should not look for the hardcoded `Self.cookieConsentNoticesUUID` but for a `uuid` function parameter
`filterLists.firstIndex(where: { $0.uuid == uuid })`

iccub (changes requested):

Client/Frontend/Widgets/GIFImage.swift
soner-yuksel: I wanted to mention this too but, it has already been designed as gif and implemented. We should use lottie animations, the gif images are holding **huge** space

soner-yuksel:
I have another question. are we supposed to show this pop-up for new users ?

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
cuba: No I use it outside in UIKit so I need this height to be public

Client/Frontend/Settings/BraveShieldsAndPrivacySettingsController.swift
cuba: habit :)

Client/WebFilters/FilterListResourceDownloader.swift
cuba: I can't remove all references of UUID yet but I believe here it will be ok.

Client/Frontend/Widgets/GIFImage.swift
cuba: Ooops...I didn't want to keep this

Client/Frontend/Widgets/GIFImage.swift
cuba: Unfortunately the animation is not compatible with Lottie. We would need to recreate it to probably use Lottie animations which might result in a different type of animation and more work for the design team. We agreed to use a gif for this case but future cases we will expect Lottie files

iccub: We decided to leave it as it is. Maybe something to improve later and for any future assets, gifs seem too filesize heavy for mobile apps

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,97 @@
...
+              self.showAnimation = true
+            }
+
+            if !FilterListResourceDownloader.shared.enableFilterList(for: FilterList.cookieConsentNoticesComponentID, isEnabled: true) {
+              assertionFailure("This filter list should exist or this UI is completely useless")
+            }
+            
+            Task {
+              try await Task.sleep(seconds: 3.5)
+              self.presentationMode.wrappedValue.dismiss()
```
iccub: This won't work on iOS 14, you can add a helper function with a custom handler something like

```
var onDismiss: (() -> Void)?

private func dismissView() {
    // Dismiss on presentation mode does not work on iOS 14
    // when using the UIHostingController is parent view.
    // As a workaround a completion handler is used instead.
    if #available(iOS 15, *) {
      presentationMode.dismiss()
    } else {
      onDismiss?()
    }
  }
```

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,97 @@
...
+    VStack {
+      VStack {
+        if !showAnimation {
+          VStack(alignment: .center, spacing: textPadding) {
+            Text(Strings.blockCookieConsentNoticesPopupTitle)
+              .font(.title)
+              .foregroundColor(Color(UIColor.braveLabel))
+              .multilineTextAlignment(.center)
+              .transition(transition)
+              .padding(.horizontal, textPadding)
```
iccub: These 3 modifiers can be removed from individual views and put in the root VStack

```
.multilineTextAlignment(.center)
.transition(transition)
.padding(.horizontal, textPadding)
```

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,97 @@
...
+  private let textPadding: CGFloat = 16
+  
+  @Environment(\.presentationMode) var presentationMode
+  @State private var showAnimation = false
+  
+  var body: some View {
+    VStack {
+      VStack {
+        if !showAnimation {
+          VStack(alignment: .center, spacing: textPadding) {
```
iccub: alignment should not be needed here, it's center by default

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,97 @@
...
+import BraveShared
+
+struct CookieNotificationBlockingConsentView: View {
+  public static let contentHeight = 480.0
+  private static let gifHeight = 328.0
+  private static let bottomSectionHeight = contentHeight - gifHeight
+  
+  private let animation = Animation.easeOut(duration: 0.5).delay(0)
+  private let transition = AnyTransition.scale(scale: 1.1).combined(with: .opacity)
+  private let textPadding: CGFloat = 16
```
iccub: You use this property not only for padding but for spacing too, either remove it and use hardcoded values or rename maybe

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,97 @@
...
+              .transition(transition)
+              .padding(.horizontal, textPadding)
+          }
+        }
+      }
+      .padding(.top, 80)
+      .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .center)
+      .background(
+        GIFImage(asset: "cookie-consent-animation", animate: showAnimation)
+          .frame(height: Self.gifHeight, alignment: Alignment.center)
```
iccub: `alignment` should not be needed

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,97 @@
...
+              .font(.body)
+              .foregroundColor(Color(UIColor.braveLabel))
+              .multilineTextAlignment(.center)
+              .transition(transition)
+              .padding(.horizontal, textPadding)
+          }
+        }
+      }
+      .padding(.top, 80)
+      .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .center)
```
iccub: `alignment` should not be needed

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,97 @@
...
+      .background(
+        GIFImage(asset: "cookie-consent-animation", animate: showAnimation)
+          .frame(height: Self.gifHeight, alignment: Alignment.center)
+      )
+      
+      VStack(alignment: .center, spacing: textPadding) {
```
iccub: `alignment` should not be needed

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,97 @@
...
+  
+  private let animation = Animation.easeOut(duration: 0.5).delay(0)
+  private let transition = AnyTransition.scale(scale: 1.1).combined(with: .opacity)
+  private let textPadding: CGFloat = 16
+  
+  @Environment(\.presentationMode) var presentationMode
+  @State private var showAnimation = false
+  
+  var body: some View {
+    VStack {
```
iccub: Has to be embedded in ScrollView most likely, to work with dynamic font sizes

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,97 @@
...
+      VStack {
+        if !showAnimation {
+          VStack(alignment: .center, spacing: textPadding) {
+            Text(Strings.blockCookieConsentNoticesPopupTitle)
```
iccub: Those labels probably need `.fixedSize(horizontal: false, vertical: true)`
so they look correctly on all devices and font sizes

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,97 @@
...
+            self.presentationMode.wrappedValue.dismiss()
+          }
+          .font(Font.body.weight(.semibold))
+          .foregroundColor(.accentColor)
+          .multilineTextAlignment(.center)
+          .transition(transition)
+          .padding(.horizontal, textPadding)
+        }
+      }
+      .frame(height: Self.bottomSectionHeight, alignment: .center)
```
iccub: `alignment` should not be needed

iccub (changes requested):
SwiftUI changes, and the new swiftUI does not support accessibility, it has to be fixed

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,107 @@
...
+
+struct CookieNotificationBlockingConsentView: View {
+  public static let contentHeight = 480.0
+  public static let contentWidth = 344.0
+  private static let gifHeight = 328.0
+  private static let bottomSectionHeight = contentHeight - gifHeight
+  
+  private let animation = Animation.easeOut(duration: 0.5).delay(0)
+  private let transition = AnyTransition.scale(scale: 1.1).combined(with: .opacity)
+  private let textPadding: CGFloat = 16
```
iccub: remove CGFloat from this one too

Sources/DesignSystem/Views/BraveButtonStyle.swift
```diff
@@ -31,6 +31,7 @@ public struct BraveButtonSize {
 public struct BraveFilledButtonStyle: ButtonStyle {
   @Environment(\.isEnabled) private var isEnabled
 
+  private let clipShape = RoundedRectangle(cornerRadius: 48, style: .continuous)
```
iccub, line 34: Need @kylehickinson for this line, to make sure we do not break this button in other views

iccub:
Most of my comments are still unaddressed

Sources/DesignSystem/Views/BraveButtonStyle.swift
kylehickinson: We do not want to use 48 instead of `Capsule`, the background shape will max out then. What's the goal with this change?

iccub: Dynamic font support

Zrzut ekranu 2022-10-3 o 18 32 56

cuba: We want it to max out so we don't get a circle on large font

Client/Frontend/Popup/CookieNotificationBlockingConsentView.swift
cuba: well the containing box is fixed width and height

iccub (changes requested):
Code looks good to me, only thing left is the BraveButton style changes

iccub (approved):

soner-yuksel (approved):

PR #5864, opened 2022-08-15
No Bug: Remove legacy code

iccub:
## Submitter Checklist:

  - [x] Light & dark mode

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Client/Frontend/Browser/New Tab Page/NewTabPageViewController.swift
```diff
@@ -109,6 +109,9 @@ class NewTabPageViewController: UIViewController {
   private var preventReloadOnBraveNewsEnabledChange = false
 
   private let notifications: NewTabPageNotifications
+  
+  /// Internal url to access the new tab page.
+  static let ntpInteralURL = URL(string: "\(InternalURL.baseUrl)/\(AboutHomeHandler.path)#panel=0")
```
cuba: Just a personal preference but I would use a ! here. Everything is hard coded here. It either fails or it doesn't

iccub: I never use force unwraps even for impossible cases :o

soner-yuksel: I think this better be inside TabManager as constant instead of static property inside NTPController

soner-yuksel (approved):
Just one comment, rest seems good, nice to remove some unused stuff.

Client/Frontend/Browser/New Tab Page/NewTabPageViewController.swift
iccub: thanks, was not really sure where to place it

PR #6038, opened 2022-09-21
Ref #6036 Improved code readability on BrowserViewController+Menu

### LINKED ISSUE #6036: Improve code BrowserViewController+Menu
### Description: 

On BrowserViewController+Menu file we have two methods responsible for creating the menu, `privacyFeaturesMenuSection` and `destinationMenuSection`. Those functions are too large with a bunch of UI declaration that is making the code a bit hard to read. PageActionsMenuSection, on its body function, is very similar as well. 

What I'm proposing is moving those codes to separate functions so that way it improves code readability and also we can reuse some of them.

mauriciomaniglia:
## Summary of Changes

This pull request fixes #6036

Moved some UI declaration to separate methods to improve readability. Those changes were apply on this functions: `privacyFeaturesMenuSection(:)`
`destinationMenuSection(:isShownOnWebPage:)`
Also on PageActionsMenuSection->body 

## Submitter Checklist:

  - [ ] Light & dark mode

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Menu.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Menu.swift
@@ -32,7 +32,7 @@ extension BrowserViewController {
     }
   }
 
-  func privacyFeaturesMenuSection(_ menuController: MenuViewController) -> some View {
+  func privacyFeaturesMenuSection(_ menuController: MenuViewController) -> some View {v
     VStack(alignment: .leading, spacing: 5) {
       Text(Strings.OptionsMenu.menuSectionTitle.capitalized)
         .font(.callout.weight(.semibold))
@@ -40,68 +40,19 @@ extension BrowserViewController {
         .padding(.horizontal, 14)
         .padding(.bottom, 5)
 
-      VPNMenuButton(
-        vpnProductInfo: self.vpnProductInfo,
-        description: Strings.OptionsMenu.braveVPNItemDescription,
-        displayVPNDestination: { [unowned self] vc in
-          (self.presentedViewController as? MenuViewController)?
-            .pushInnerMenu(vc)
-        },
-        enableInstalledVPN: { [unowned menuController] in
-          /// Donate Enable VPN Activity for suggestions
-          let enableVPNActivity = ActivityShortcutManager.shared.createShortcutActivity(type: .enableBraveVPN)
-          menuController.userActivity = enableVPNActivity
-          enableVPNActivity.becomeCurrent()
-        }
-      )
-
-      MenuItemButton(
-        icon: UIImage(named: "playlist_menu", in: .current, compatibleWith: nil)!.template,
-        title: Strings.OptionsMenu.bravePlaylistItemTitle,
-        subtitle: Strings.OptionsMenu.bravePlaylistItemDescription
-      ) { [weak self] in
-        guard let self = self else { return }
-
-        self.presentPlaylistController()
-      }
-
+      vpnButton(menuController)
+      playlistButton()
+      
       // Add Brave Talk and News options only in normal browsing
       if !PrivateBrowsingManager.shared.isPrivateBrowsing {
         // Show Brave News if it is first launch and after first launch If the new is enabled
         if Preferences.General.isFirstLaunch.value || (!Preferences.General.isFirstLaunch.value && Preferences.BraveNews.isEnabled.value) {
-          MenuItemButton(
-            icon: UIImage(named: "menu_brave_news", in: .current, compatibleWith: nil)!.template,
-            title: Strings.OptionsMenu.braveNewsItemTitle,
-            subtitle: Strings.OptionsMenu.braveNewsItemDescription
-          ) { [weak self] in
-            guard let self = self, let newTabPageController = self.tabManager.selectedTab?.newTabPageViewController else {
-              return
-            }
-
-            self.popToBVC()
-            newTabPageController.scrollToBraveNews()
-          }
+          newsButton()
         }
-
-        MenuItemButton(
-          icon: UIImage(named: "menu-brave-talk", in: .current, compatibleWith: nil)!.template,
-          title: Strings.OptionsMenu.braveTalkItemTitle,
-          subtitle: Strings.OptionsMenu.braveTalkItemDescription
-        ) { [weak self] in
-          guard let self = self, let url = URL(string: "https://talk.brave.com/") else { return }
-
-          self.popToBVC()
-          self.finishEditingAndSubmit(url, visitType: .typed)
-        }
+        talkButton()
       }
 
-      MenuItemButton(
-        icon: UIImage(named: "menu-crypto", in: .current, compatibleWith: nil)!.template,
-        title: Strings.Wallet.wallet,
-        subtitle: Strings.OptionsMenu.braveWalletItemDescription
-      ) { [unowned self] in
-        self.presentWallet()
-      }
+      cryptoButton()
     }
     .fixedSize(horizontal: false, vertical: true)
     .padding(.top, 10)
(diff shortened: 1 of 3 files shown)


Client/Frontend/Browser/BrowserViewController/BrowserViewController+Menu.swift
```diff
@@ -40,68 +40,19 @@ extension BrowserViewController {
...
-        MenuItemButton(
-          icon: UIImage(named: "menu-brave-talk", in: .current, compatibleWith: nil)!.template,
-          title: Strings.OptionsMenu.braveTalkItemTitle,
-          subtitle: Strings.OptionsMenu.braveTalkItemDescription
-        ) { [weak self] in
-          guard let self = self, let url = URL(string: "https://talk.brave.com/") else { return }
-
-          self.popToBVC()
-          self.finishEditingAndSubmit(url, visitType: .typed)
+          newsButton()
```
soner-yuksel: I think instead of moving every button creation to a new function separately we should create a function which handles all the MenuItemButton creations. This will allow us to keep track what types of buttons we have and we can easily change/update the details for these UI elements.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Menu.swift
```diff
@@ -211,6 +169,74 @@ extension BrowserViewController {
       }
     }
   }
+  
+  private func bookmarksButton(_ menuController: MenuViewController) -> MenuItemButton {
```
soner-yuksel: Like I mentioned in the previous comment we should have one function which handles `MenuItemButton` creation and for helping this parametrized creation we have to create `enum` which can help us to determine the details of every button like

```
  // MARK: MenuButtonType
  
  enum MenuButtonType {
    case news
    case talk
    case wallet(subtitle: String)
    
      .
      .
      .

    var icon: UIImage {
      switch self {
        case .news:
          return UIImage(named: "menu_brave_news", in: .current, compatibleWith: nil)!.template
        case .talk:
          return UIImage(named: "menu-brave-talk", in: .current, compatibleWith: nil)!.template
        case .wallet(_):
          return UIImage(named: "menu-crypto", in: .current, compatibleWith: nil)!.template
      }
    }
    
    .
    .
    .
  }
```

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Menu.swift
```diff
@@ -211,6 +169,74 @@ extension BrowserViewController {
...
+      let vc = BookmarksViewController(
+        folder: bookmarkManager.lastVisitedFolder(),
+        bookmarkManager: bookmarkManager,
+        isPrivateBrowsing: PrivateBrowsingManager.shared.isPrivateBrowsing)
+      vc.toolbarUrlActionsDelegate = self
+      menuController.presentInnerMenu(vc)
+    }
+  }
+  
+  private func historyButton(_ menuController: MenuViewController) -> MenuItemButton {
```
soner-yuksel: Swap all these MenuButtonItem creations with a single function similar to 

```
  private func createMenuButton(for buttonType: MenuButtonType, completion: @escaping () -> Void) -> MenuItemButton {   
    MenuItemButton(icon: buttonType.icon, title: buttonType.title, subTitle: buttonType.subTitle) {
      completion()
    }
  }
```

soner-yuksel:
@mauriciomaniglia thanks for the contribution, I left some comments about how we should approach to this improvement. Please feel free to ask any question you have.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Menu.swift
mauriciomaniglia: Hi @soner-yuksel, thanks for the review! I like this idea of moving the button creation to some unified place. Instead of using a function for that, what do you think about moving this logic to a separate component? If we do that, we can still have all the benefits that you mention, but it will also be much cleaner the code for both this new component and the current file. More important, it will also be much easy to add unit tests. I thought something like that: 

```
class MenuButtonFactory {
    enum MenuButtonType {
      case news
      case talk
      case wallet(subtitle: String)
      
      var icon: UIImage {
        switch self {
          case .news:
            return UIImage(named: "menu_brave_news", in: .current, compatibleWith: nil)!.template
          case .talk:
            return UIImage(named: "menu-brave-talk", in: .current, compatibleWith: nil)!.template
          case .wallet(_):
            return UIImage(named: "menu-crypto", in: .current, compatibleWith: nil)!.template
        }
      }
      
      var title: String {
        switch self {
        case .news:
          return ""
        case .talk:
          return ""
        case .wallet:
          return ""
        }
      }
      
      var subtitle: String {
        switch self {
        case .news:
          return ""
        case .talk:
          return ""
        case .wallet:
          return ""
        }
      }
    }
    
    static func createMenuButton(for buttonType: MenuButtonType, completion: @escaping () -> Void) -> MenuItemButton {
      MenuItemButton(icon: buttonType.icon, title: buttonType.title, subtitle: buttonType.subtitle) {
        completion()
      }
    }
  }
```

```
 MenuButtonFactory.createMenuButton(for: .news)
```

What do you think?

soner-yuksel: @mauriciomaniglia  Sounds like a good idea. Please add the factory class and keep the completion handling where the button is created.

mauriciomaniglia: @soner-yuksel I've just updated the PR with the changes

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Menu.swift
```diff
@@ -211,6 +125,81 @@ extension BrowserViewController {
       }
     }
   }
+  
+  private func navigateToBraveNews() {
```
soner-yuksel: @mauriciomaniglia can you remove the functions for the actions and add all the logic directly inside the completion blocks here? We are not using these function more than once and it makes it irrelevant and hard to track this way.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Menu.swift
soner-yuksel: @mauriciomaniglia thanks you for all the changes, I have one more comment but after that it looks like it ready to go.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Menu.swift
mauriciomaniglia: @soner-yuksel  Yes. But before I do that, are you sure about that? I moved the actions to separate functions to make the code more readable, look: 

Screen Shot 2022-09-27 at 17 50 13

Even though those functions have not been used more than once, when you separate the code into blocks and glue those functions altogether, it looks nicer to read. 

If we keep all logic inside the completions blocks, it will look like this:

Screen Shot 2022-09-27 at 17 59 05
Screen Shot 2022-09-27 at 17 59 24

For me is fine anyway, is just a matter of taste 🙂

soner-yuksel: Yeap pretty sure

mauriciomaniglia: Done ✅

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Menu.swift
```diff
@@ -55,25 +55,16 @@ extension BrowserViewController {
...
-        icon: UIImage(named: "playlist_menu", in: .current, compatibleWith: nil)!.template,
-        title: Strings.OptionsMenu.bravePlaylistItemTitle,
-        subtitle: Strings.OptionsMenu.bravePlaylistItemDescription
-      ) { [weak self] in
+      MenuItemFactory.button(for: .playlist(subtitle: Strings.OptionsMenu.bravePlaylistItemDescription)) { [weak self] in
         guard let self = self else { return }
-
         self.presentPlaylistController()
       }
-
```
soner-yuksel, line 67: Please fix the unnecessary spacing

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Menu.swift
```diff
@@ -110,43 +92,40 @@ extension BrowserViewController {
...
     VStack(spacing: 0) {
-      MenuItemButton(icon: UIImage(named: "menu_bookmarks", in: .current, compatibleWith: nil)!.template, title: Strings.bookmarksMenuItem) { [unowned self, unowned menuController] in
-        let vc = BookmarksViewController(
-          folder: bookmarkManager.lastVisitedFolder(),
-          bookmarkManager: bookmarkManager,
-          isPrivateBrowsing: PrivateBrowsingManager.shared.isPrivateBrowsing)
-        vc.toolbarUrlActionsDelegate = self
-        menuController.presentInnerMenu(vc)
+      MenuItemFactory.button(for: .bookmarks) { [unowned self, unowned menuController] in
+          let vc = BookmarksViewController(
```
soner-yuksel: Indentation is off for this code block.

soner-yuksel:
@mauriciomaniglia well done 🎉 and thanks for the contribution again.

soner-yuksel (approved):

Issue #5944, opened 2022-08-30
Wallet: UI update on the `Connected` button in wallet panel for Solana Dapps connection


nuo-xu:
### Description: 
Once there at least one Solana Account has been granted permission to connect to a Solana Dapp.
The connected status button will be unhidden and update the UI to solana specific as shown in the screenshot.
Permitted and connected account will show `Connected` with a green dot on the left side.
Permitted but not connected account will show `Disconnected` with a red dot on the left side.
When user set to not allow site access to the solana provider apis will show `Blocked` with a red dot in the left side.

![Screen Shot 2022-08-30 at 4 35 45 PM](https://user-images.githubusercontent.com/1187676/187538204-516ca751-e875-46e7-ba2b-70093d62b289.png)

PR #6050, opened 2022-09-22
Fix #6049: Solana Dapps Javascript communication layer

### LINKED ISSUE #6049: Solana dapps javascript communication layer
### Description: 

Implement the javascript communication layer between each tab's web view and the `BraveWalletSolanaProvider`. 
Logic should remain behind the `isSolanaDappsEnabled` feature flag.

StephenHeaps:
## Summary of Changes
- PR is intended for initial security review of the javascript communication layer for Solana dapps.

Some features are currently unavailable:
- Creating a solana account automatically if one does not yet exist when connecting to a Solana dapp: https://github.com/brave/brave-ios/issues/6046
- Solana dapp transaction parsing (transactions will appear with some incorrect / unclear info): https://github.com/brave/brave-ios/issues/6004
- Solana Sign Transaction / Sign All Transactions panel: https://github.com/brave/brave-ios/issues/6005
- Connected button UI on wallet panel: https://github.com/brave/brave-ios/issues/5943 & https://github.com/brave/brave-ios/issues/5944

This pull request fixes #6049

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:
1. Enable `isSolanaDappsEnabled` feature flag in `WalletConstants.swift`
2. Build & run the project, visit Solana dapps websites.
    - Brave test site [hosted](https://codesandbox.io/s/github/darkdh/solana-provider-test) ([source](https://github.com/darkdh/solana-provider-test))
    - [MagicEden](https://magiceden.io/)
    - [Jupiter](https://jup.ag/)

## Screenshots:

https://user-images.githubusercontent.com/5314553/191633069-525fb167-55cc-4c2d-a9ef-8b59dade9825.mov

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
@@ -553,6 +553,7 @@ extension BrowserViewController: WKNavigationDelegate {
     
     // Need to evaluate Night mode script injection after url is set inside the Tab
     tab.nightMode = Preferences.General.nightModeEnabled.value
+    tab.clearSolanaConnectedAccounts()
 
     rewards.reportTabNavigation(tabId: tab.rewardsId)
 
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
@@ -80,6 +80,8 @@ extension BrowserViewController: WKNavigationDelegate {
        selectedTab.url?.origin != webView.url?.origin {
       // new site has a different origin, hide wallet icon.
       tabManager.selectedTab?.isWalletIconVisible = false
+      // new site, reset connected addresses
+      tabManager.selectedTab?.clearSolanaConnectedAccounts()
       // close wallet panel if it's open
       if let popoverController = self.presentedViewController as? PopoverController,
          popoverController.contentController is WalletPanelHostingController {
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
@@ -597,6 +597,7 @@ extension BrowserViewController: WKNavigationDelegate {
         )
       }
       tab.updateEthereumProperties()
+      tab.updateSolanaProperties()
       tab.reportPageLoad(to: rewards, redirectionURLs: tab.redirectURLs)
       tab.redirectURLs = []
       if webView.url?.isLocal == false {
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
@@ -600,7 +600,9 @@ extension BrowserViewController: WKNavigationDelegate {
         )
       }
       tab.updateEthereumProperties()
-      tab.updateSolanaProperties()
+      Task { @MainActor in
+        await tab.updateSolanaProperties()
+      }
       tab.reportPageLoad(to: rewards, redirectionURLs: tab.redirectURLs)
       tab.redirectURLs = []
       if webView.url?.isLocal == false {
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
@@ -600,7 +600,7 @@ extension BrowserViewController: WKNavigationDelegate {
         )
       }
       tab.updateEthereumProperties()
-      Task { @MainActor in
+      Task {
         await tab.updateSolanaProperties()
       }
       tab.reportPageLoad(to: rewards, redirectionURLs: tab.redirectURLs)
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
@@ -308,7 +308,9 @@ extension Tab: BraveWalletProviderDelegate {
   
   func clearSolanaConnectedAccounts() {
     tabDappStore.solConnectedAddresses = .init()
-    updateSolanaProperties()
+    Task { @MainActor in
+      await updateSolanaProperties()
+    }
   }
 }
 
@@ -406,7 +408,7 @@ extension Tab: BraveWalletSolanaEventsListener {
         let script = "window.solana.emit('accountChanged', _brave_solana.createPublickey('\(account)'))"
         await webView.evaluateSafeJavaScript(functionName: script, contentWorld: .page, asFunction: false)
       }
-      updateSolanaProperties()
+      await updateSolanaProperties()
     }
   }
 
@@ -428,30 +430,26 @@ extension Tab: BraveWalletSolanaEventsListener {
     }
   }
 
-  func updateSolanaProperties() {
-    guard Preferences.Wallet.defaultSolWallet.value == Preferences.Wallet.WalletType.brave.rawValue else {
+  @MainActor func updateSolanaProperties() async {
+    guard Preferences.Wallet.defaultSolWallet.value == Preferences.Wallet.WalletType.brave.rawValue,
+          let webView = webView,
+          let provider = walletSolProvider else {
       return
     }
-    Task { @MainActor in
-      guard let webView = webView,
-            let provider = walletSolProvider else {
-        return
-      }
-      let isConnected = await provider.isConnected()
+    let isConnected = await provider.isConnected()
+    await webView.evaluateSafeJavaScript(
+      functionName: "window.solana.isConnected = \(isConnected)",
+      contentWorld: .page,
+      asFunction: false
+    )
+    // publicKey
+    if let keyringService = walletKeyringService,
+       let publicKey = await keyringService.selectedAccount(.sol) {
       await webView.evaluateSafeJavaScript(
-        functionName: "window.solana.isConnected = \(isConnected)",
+        functionName: "if (window._brave_solana.createPublickey) { window.solana.publicKey = _brave_solana.createPublickey('\(publicKey)'); }",
         contentWorld: .page,
         asFunction: false
       )
-      // publicKey
-      if let keyringService = walletKeyringService,
-         let publicKey = await keyringService.selectedAccount(.sol) {
-        await webView.evaluateSafeJavaScript(
-          functionName: "window.solana.publicKey = _brave_solana.createPublickey('\(publicKey)');",
-          contentWorld: .page,
-          asFunction: false
-        )
-      }
     }
   }
 }
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
@@ -78,8 +78,7 @@ extension BrowserViewController {
   /// when the pending request is updated so we can update the wallet url bar button.
   func newWalletStore() -> WalletStore? {
     let privateMode = PrivateBrowsingManager.shared.isPrivateBrowsing
-    guard let walletStore = WalletStore.from(privateMode: privateMode)
-    else {
+    guard let walletStore = WalletStore.from(privateMode: privateMode) else {
       Logger.module.error("Failed to load wallet. One or more services were unavailable")
       return nil
     }
@@ -295,21 +294,6 @@ extension Tab: BraveWalletProviderDelegate {
     // TODO: can be called when Solana dapp tries to connect but user does not have Solana account
   }
   
-  func isSolanaAccountConnected(_ account: String) -> Bool {
-    walletSolConnectedAddresses.contains(account)
-  }
-  
-  func addSolanaConnectedAccount(_ account: String) {
-    walletSolConnectedAddresses.insert(account)
-  }
-  
-  func removeSolanaConnectedAccount(_ account: String) {
-    walletSolConnectedAddresses.remove(account)
-  }
-  
-  func clearSolanaConnectedAccounts() {
-    walletSolConnectedAddresses = .init()
-  }
   
   func isSolanaAccountConnected(_ account: String) -> Bool {
     tabDappStore.solConnectedAddresses.contains(account)
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
@@ -265,7 +265,14 @@ extension Tab: BraveWalletProviderDelegate {
   }
   
   func isPermissionDenied(_ type: BraveWallet.CoinType) -> Bool {
-    return type != .eth
+    switch type {
+    case .eth, .sol:
+      return false
+    case .fil:
+      return true
+    @unknown default:
+      return true
+    }
   }
   
   func showAccountCreation(_ type: BraveWallet.CoinType) {
@@ -371,6 +378,50 @@ extension Tab: BraveWalletEventsListener {
         asFunction: false,
         completion: nil
       )
+    }
+  }
+}
+
+extension Tab: BraveWalletSolanaEventsListener {
+  func accountChangedEvent(_ account: String?) {
+    emitSolanaEvent(.solanaAccountChanged(account: account ?? ""))
+    updateSolanaProperties()
+  }
+
+  func emitSolanaEvent(_ event: Web3ProviderEvent) {
+    guard Preferences.Wallet.defaultSolWallet.value == Preferences.Wallet.WalletType.brave.rawValue,
+          let webView = webView else {
+      return
+    }
+    Task { @MainActor in
+      var arguments: [Any] = [event.name]
+      if let eventArgs = event.arguments {
+        arguments.append(eventArgs)
+      }
+      await webView.evaluateSafeJavaScript(
+        functionName: "window.solana.emit", // TODO: `window.braveSolana` ?
+        args: arguments,
+        contentWorld: .page
+      )
+    }
+  }
+
+  func updateSolanaProperties() {
+    guard Preferences.Wallet.defaultSolWallet.value == Preferences.Wallet.WalletType.brave.rawValue else {
+      return
+    }
+    Task { @MainActor in
+      guard let webView = webView,
+            let provider = walletSolProvider else {
+        return
+      }
+      let isConnected = await provider.isConnected()
+      await webView.evaluateSafeJavaScript(
+        functionName: "window.solana.isConnected = \(isConnected)",
+        contentWorld: .page,
+        asFunction: false
+      )
+      // TODO: publicKey
     }
   }
 }
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
@@ -212,7 +212,10 @@ extension Tab: BraveWalletProviderDelegate {
           return
         }
       case .sol:
-        if success, accounts.allSatisfy({ allowedAccounts.contains($0) }) {
+        if success,
+           accounts.count == 1, // only 1 Solana account connect at a time
+           let account = accounts.first,
+           allowedAccounts.contains(account) {
           completion(.none, allowedAccounts)
           return
         }
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
@@ -444,7 +444,8 @@ extension Tab: BraveWalletSolanaEventsListener {
     )
     // publicKey
     if let keyringService = walletKeyringService,
-       let publicKey = await keyringService.selectedAccount(.sol) {
+       let publicKey = await keyringService.selectedAccount(.sol),
+       self.isSolanaAccountConnected(publicKey) {
       await webView.evaluateSafeJavaScript(
         functionName: "if (window._brave_solana.createPublickey) { window.solana.publicKey = _brave_solana.createPublickey('\(publicKey)'); }",
         contentWorld: .page,
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
@@ -401,8 +401,13 @@ extension Tab: BraveWalletEventsListener {
 
 extension Tab: BraveWalletSolanaEventsListener {
   func accountChangedEvent(_ account: String?) {
-    emitSolanaEvent(.solanaAccountChanged(account: account ?? ""))
-    updateSolanaProperties()
+    Task { @MainActor in
+      if let webView = webView, let account = account {
+        let script = "window.solana.emit('accountChanged', _brave_solana.createPublickey('\(account)'))"
+        await webView.evaluateSafeJavaScript(functionName: script, contentWorld: .page, asFunction: false)
+      }
+      updateSolanaProperties()
+    }
   }
 
   func emitSolanaEvent(_ event: Web3ProviderEvent) {
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
@@ -176,9 +176,25 @@ extension Tab: BraveWalletProviderDelegate {
       // Check if eth permissions already exist for this origin and if they don't, ensure the user allows
       // ethereum/solana provider access
       let walletPermissions = origin.url.map { Domain.walletPermissions(forUrl: $0, coin: coinType) ?? [] } ?? []
-      if walletPermissions.isEmpty, !Preferences.Wallet.allowEthProviderAccess.value {
-        completion(.internal, nil)
-        return
+      if walletPermissions.isEmpty {
+        switch coinType {
+        case .eth:
+          if !Preferences.Wallet.allowEthProviderAccess.value {
+            completion(.internal, nil)
+            return
+          }
+        case .sol:
+          if !Preferences.Wallet.allowSolProviderAccess.value {
+            completion(.internal, nil)
+            return
+          }
+        case .fil:
+          // not supported
+          fallthrough
+        @unknown default:
+          completion(.internal, nil)
+          return
+        }
       }
       
       guard WalletStore.from(privateMode: isPrivate) != nil else {
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
@@ -431,7 +431,8 @@ extension Tab: BraveWalletSolanaEventsListener {
   }
 
   @MainActor func updateSolanaProperties() async {
-    guard Preferences.Wallet.defaultSolWallet.value == Preferences.Wallet.WalletType.brave.rawValue,
+    guard WalletDebugFlags.isSolanaDappsEnabled,
+          Preferences.Wallet.defaultSolWallet.value == Preferences.Wallet.WalletType.brave.rawValue,
           let webView = webView,
           let provider = walletSolProvider else {
       return
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
@@ -440,9 +440,7 @@ extension Tab: BraveWalletSolanaEventsListener {
       )
       // publicKey
       if let keyringService = walletKeyringService,
-         let userScriptManager = userScriptManager,
          let publicKey = await keyringService.selectedAccount(.sol) {
-        await userScriptManager.injectSolanaInternalScript()
         let (value, _) = await webView.evaluateSafeJavaScript(
           functionName: "_brave_solana.createPublickey",
           args: [publicKey],
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
@@ -444,7 +444,7 @@ extension Tab: BraveWalletSolanaEventsListener {
         let (value, _) = await webView.evaluateSafeJavaScript(
           functionName: "_brave_solana.createPublickey",
           args: [publicKey],
-          contentWorld: .page
+          contentWorld: .walletSandbox
         )
         guard let dict = value as? [String: Any],
               let createdPublicKey = dict["publicKey"],
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
@@ -294,7 +294,6 @@ extension Tab: BraveWalletProviderDelegate {
     // TODO: can be called when Solana dapp tries to connect but user does not have Solana account
   }
   
-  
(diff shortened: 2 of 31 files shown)


Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
```diff
@@ -377,6 +416,66 @@ extension Tab: BraveWalletEventsListener {
...
+  func accountChangedEvent(_ account: String?) {
+    Task { @MainActor in
+      if let webView = webView {
+        let script: String
+        if let account = account {
+          script = "window.solana.emit('accountChanged', _brave_solana.createPublickey('\(account)'))"
+        } else {
+          script = "window.solana.emit('accountChanged')"
+        }
+        await webView.evaluateSafeJavaScript(functionName: script, contentWorld: .page, asFunction: false)
```
stoletheminerals, line 430: nit: Here and in a few other places. I think it's not secure to eval scripts like this (passing both a function and parameter in `functionName`), the `evaluateSafeJavaScript` is supposed to be called with `args` parameter. However, I don't see how it could be exploited even if I have control over injected variable (`account` in this case). We would need to recheck this if or when we enable dapps support in iframes.

Client/Frontend/Browser/UserScriptManager.swift
```diff
@@ -569,6 +597,42 @@ class UserScriptManager {
...
+      
+      if WalletDebugFlags.isSolanaDappsEnabled,
+         let script = walletSolProviderScript,
+         tab?.isPrivate == false,
+         Preferences.Wallet.WalletType(rawValue: Preferences.Wallet.defaultSolWallet.value) == .brave {
+        $0.addUserScript(script)
+        if var providerJS = walletSolProviderScripts[.solana] {
+          providerJS = """
+            (function($Object) {
+              if (window.isSecureContext) {
```
kdenhartog: Is this a way to check that window.isSecureContext is in use before injecting here? Looks like we may need to update the ETH provider with this same method as well as we're not doing it in the same way and this looks to more accurately reflect what should be done.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
StephenHeaps: Most of these were done to preserve the [solanaWeb3.PublicKey](https://solana-labs.github.io/solana-web3.js/classes/PublicKey.html) type (or the [solanaWeb3.Transaction](https://solana-labs.github.io/solana-web3.js/classes/Transaction.html) type). 

In [v1.44.x of core](https://github.com/brave/brave-core/blob/1.44.x/components/brave_wallet/resources/solana_provider_internal.js#L13), the `solanaWeb3.PublicKey` object was being returned in an object/dictionary which we'd have to extract out and serialize which caused the object to lose it's methods so we were seeing `toBase58 is not a function` errors for example. A [prior commit](https://github.com/brave/brave-ios/pull/6050/commits/a0c8ebee6a8933a5a41a91cc3c22f1f80ceb3991#diff-2b6000bef9e917076bcc4d41ebcdf411adbbd29f19ca80ebbd01d6d2e4517957L446-R447) was trying to create these models using the correct `functionName` and `asFunction` values when using v1.44.x of core but `JSONSerialization` would strip the methods as methods don't exist in json. I'll do another check to see if this is possible with the `solanaWeb3.PublicKey` object being returned directly in [v1.45.x of core](https://github.com/brave/brave-core/blob/1.45.x/components/brave_wallet/resources/solana_provider_internal.js#L14).

In the case above for only emitting `accountChanged` event we should be able to update to use the function correctly, ex. `await webView.evaluateSafeJavaScript(functionName: "window.solana.emit, args: ["accountChanged"], contentWorld: .page, asFunction: true)`.

stoletheminerals: I'm more concerned about line 425:
https://github.com/brave/brave-ios/blob/91dd9e879176e52382b60421762c5f278ba8a24e/Client/Frontend/Browser/BrowserViewController/BrowserViewController%2BWallet.swift#L425

There are other examples where we pass `publicKey` parameter directly to `functionName`

StephenHeaps: Maybe I'm misunderstanding, is the concern that `publicKey` parameter is unsanitized? The `publicKey` value we use is always from our core apis so it should be sanitized already, so I believe we should be fine here.

Client/Frontend/UserContent/UserScripts/WalletSolanaProvider.js
```diff
@@ -0,0 +1,139 @@
...
+      })
+    }
+    /* <solanaWeb3.Transaction> ->
+      {transaction: <solanaWeb3.Transaction>,
+       serializedMessage: <base58 encoded string>,
+       signatures: [{publicKey: <base58 encoded string>, signature: <Buffer>}]} */
+    function convertTransaction(transaction) {
+      const serializedMessage = transaction.serializeMessage();
+      const signatures = transaction.signatures;
+      function convertSignaturePubkeyPair(signaturePubkeyPair) {
```
kdenhartog: non blocking nit: cause keypair can sometimes be read as the pair of public and private keys
```suggestion
      function convertSignaturePubkeyTuple(signaturePubkeyTuple) {
```

Client/Frontend/UserContent/UserScripts/WalletSolanaProvider.js
```diff
@@ -0,0 +1,139 @@
+// Copyright 2022 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+(function($Object) {
+  if (window.isSecureContext) {
+    function post(method, payload, completion) {
+      return new Promise((resolve, reject) => {
+        webkit.messageHandlers.$<handler>.postMessage({
```
kdenhartog: I might be abit confused here, but the usage of postMessage makes me think that there may be some unexpected behavior here if the frame directly calls this messageHandler via a postMessage rather than through our communication layer. Is that a possibility which the way this is written?

Client/Wallet/SolanaProviderHelper.swift
```diff
@@ -0,0 +1,336 @@
...
+    Task { @MainActor in
+      guard let tab = tab,
+            !tab.isPrivate,
+            let provider = tab.walletSolProvider,
+            !message.frameInfo.securityOrigin.host.isEmpty, // Fail if there is no last committed URL yet
+            message.frameInfo.isMainFrame, // Fail the request came from 3p origin
+            JSONSerialization.isValidJSONObject(message.body),
+            let messageData = try? JSONSerialization.data(withJSONObject: message.body, options: []),
+            let body = try? JSONDecoder().decode(MessageBody.self, from: messageData),
+            body.securityToken == UserScriptManager.securityTokenString
```
kdenhartog: What happens if the JSON object passed as the body already contains a property called securityToken?

Client/Wallet/SolanaProviderHelper.swift
```diff
@@ -0,0 +1,336 @@
...
+    }
+    guard let encodedSerializedTx = JSONSerialization.jsObject(withNative: serializedTx) else {
+      return (nil, buildErrorJson(status: .internalError, errorMessage: "Internal error"))
+    }
+    return (encodedSerializedTx, nil)
+  }
+  
+  /// Given args `[{serializedMessage: Buffer, signatures: {publicKey: String, signature: Buffer}}]`,
+  /// will encoded the response as a json object for success or provide an error dictionary for failures
+  @MainActor func signAllTransactions(args: String?) async -> (Any?, String?) {
```
kdenhartog: Is this a standard function that we use or an additional UX button we've added for this? I'd be concerned that a secondary transaction that's not been reviewed by the user gets injected to get a faked transaction.

@kdenhartog todo: when reviewing the UX verify what this function is being used for. (I see this is something exposed by the provider, so probably worth me double checking the solana docs here too)

Sources/Shared/Extensions/WKWebViewExtensions.swift
```diff
@@ -90,4 +90,23 @@ public extension WKWebView {
...
+  
+  @discardableResult @MainActor func evaluateSafeJavaScript(
+    functionName: String,
+    args: [Any] = [],
+    contentWorld: WKContentWorld,
+    escapeArgs: Bool = true,
+    asFunction: Bool = true
+  ) async -> (Any?, Error?) {
+    await withCheckedContinuation { continuation in
+      evaluateSafeJavaScript(
```
kdenhartog, line 102: This is probably a stupid question that I just glazed over and didn't figure out by myself, but what makes the JS safe in this case?

kdenhartog (commented):
initial review. So far things are looking pretty good

Client/Frontend/UserContent/UserScripts/WalletSolanaProvider.js
stoletheminerals: I can answer this, I think, since we just completed the JS security review.  It's technically possible to call postMessage directly, however, to do this, a webpage has to know the message handler's name and the security token, which are randomized every time the browser is started. And after the latest JS rework by @Brandon-T it's not possible (as far as we know) to leak these values.

Client/Wallet/SolanaProviderHelper.swift
StephenHeaps: Currently these are passed into the Solana provider in core and handled with the other pending requests in [CryptoStore](https://github.com/brave/brave-ios/blob/solana-dapps/Sources/BraveWallet/Crypto/Stores/CryptoStore.swift#L287-L288), which are shown as modals over the wallet panel or main wallet using [RequestContainerView](https://github.com/brave/brave-ios/blob/solana-dapps/Sources/BraveWallet/Panels/RequestContainerView.swift#L86). 

The UI isn't implemented yet for this request on iOS (Ticket for this UI [#6005](https://github.com/brave/brave-ios/issues/6005)), but the multiple transactions to be signed they are shown in a single panel within a list currently on desktop which UI we try to align with. ex. from the [test solana dapp site](https://github.com/darkdh/solana-provider-test) `Sign All Transactions (multiple)` button:
signAllTransactions
Details:
```
System Program - Transfer
From Address:
Solana Account 1
To Address:
Solana Account 1
Amount:
0.0000001 SOL

System Program - Transfer
From Address:
Solana Account 1
To Address:
Solana Account 1
Amount:
0.0000001 SOL
```

Client/Wallet/SolanaProviderHelper.swift
StephenHeaps: If a `securitytoken` is passed in the JSON object that wasn't added from our `postMessage` (which is randomized every session), it would fail the check here and return and log the ["Failed to handle solana provider communication"](https://github.com/brave/brave-ios/blob/solana-dapps/Client/Wallet/SolanaProviderHelper.swift#L86) error below and not proceed with any solana provider communication.

Sources/Shared/Extensions/WKWebViewExtensions.swift
StephenHeaps: This function is just a async wrapper around the existing function. The safe part of this function is the args are sanitized before being passed into the javascript in the [generateJSFunctionString function](https://github.com/brave/brave-ios/blob/solana-dapps/Sources/Shared/Extensions/WKWebViewExtensions.swift#L26). For this PR specific usage of this function, some [additional info](https://github.com/brave/brave-ios/pull/6050#discussion_r981247170) on why we weren't able to use the args param as often in this PR is in discussion above.

Client/Frontend/Browser/UserScriptManager.swift
StephenHeaps: Yeah this is done to check that we're in a secure context before injecting.
The Ethereum provider script is [being injected the same way above](https://github.com/brave/brave-ios/blob/solana-dapps/Client/Frontend/Browser/UserScriptManager.swift#L590-L597), just lost behind the fold on Github on this PR. Additionally there is a [hasOnlySecureContent check](https://github.com/brave/brave-ios/blob/solana-dapps/Client/Wallet/EthereumProviderHelper.swift#L85-L89) in the `EthereumProviderHelper` (similar to `SolanaProviderHelper`).

nuo-xu: eth provider is also injected behind this. there was a [discussion](https://bravesoftware.slack.com/archives/C023VS4HJ6Q/p1655313974617269?thread_ts=1655216153.799099&cid=C023VS4HJ6Q) for this and result in [here](https://github.com/brave/brave-ios/blob/3f59b86e76503dc1b9497913d7fa37766ccefaa7/Client/Frontend/Browser/UserScriptManager.swift#L565)

Client/Wallet/SolanaProviderHelper.swift
Brandon-T: It shouldn't be an issue @kdenhartog because it is passed like:

```javascript
postMessage({"securityToken": OurInjectedToken, message: actualJSObjectContainingSecurityTokenAlready})
```

instead of:
```javascript
postMessage(actualJSObjectContainingSecurityTokenAlready)
```

So the iOS side would receive:
```swift
{
    securityToken: "injectedToken",
    message: {"securityToken": ..., field: "...", otherField: "..."}
}
```

But also, my own PR will merge before this one, and the parsing here will HAVE to change (it cannot use `body.securityToken == UserScriptManager.securityToken` anymore, as there is no such thing.

Client/Frontend/UserContent/UserScripts/WalletSolanaProvider.js
StephenHeaps: Thanks, fixed in https://github.com/brave/brave-ios/pull/6050/commits/f5f36788b12a788ab3539b56d677a4610ecdd54f

Client/Wallet/SolanaProviderHelper.swift
kdenhartog: Sounds safe to me, thanks for the context resolving this one.

Client/Frontend/UserContent/UserScripts/WalletSolanaProvider.js
kdenhartog: Ok good to know. That makes sense and isn't that much different from most other programming language designs then. Sounds like basically an object capability model with the slight difference being that this assumption can be broken with the ability to directly inspect memory (rather than write/use memory like most overflow style attacks) which is still quite hard.

Thanks for the better context on how these work as well!

Client/Frontend/Browser/UserScriptManager.swift
kdenhartog: Thanks found it. I've been working on an EIP for this lately and had another wallet developer ask how this can be done on iOS. When I originally looked for it I didn't find this part of the code. I'll link them to this so they can see an example.

Sources/Shared/Extensions/WKWebViewExtensions.swift
kdenhartog: Ok, gonna resolve this one then and we can consolidate the discussion into that comment above.

Client/Wallet/SolanaProviderHelper.swift
kdenhartog: Thanks for the comment, I got some time to look into this a bit more and realized this is more a design flaw of the API IMO than something specific to the way we've implemented it.

This seems like something where the UX could be improved here to highlight the separate transactions better, but AFAICT there's nothing specifically that's a problem here because we're still displaying all the relevant information. My biggest worry would be that an additional transaction gets injected here during the request, but as long as the wallet hasn't been compromised then the user would still be able to detect this additional transaction.

I think this is fine from my perspective, so going to resolve it.

Client/Frontend/Browser/UserScriptManager.swift
```diff
@@ -202,4 +234,23 @@ class UserScriptManager {
...
+  
+  @MainActor func injectSolanaInternalScript(tab: Tab, solanaInternalScript: String?) async {
+    guard let webView = tab.webView,
+          let solanaInternalScript = solanaInternalScript else {
+      return
+    }
+    // inject the internal solana script
+    let script = """
+    window.__firefox__.execute(function($, $Object) {
+      \(solanaInternalScript)
```
kdenhartog: Where is `solanaInternalScript` coming from here? This looks odd that we're just randomly injecting a script that executes and without santizing what's actually executing here

StephenHeaps: This script comes from BraveCore. This `injectSolanaInternalScript(tab:solanaInternalScript:)` function is called from the [SolanaProviderScriptHandler](https://github.com/brave/brave-ios/blob/solana-dapps/Client/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Paged/SolanaProviderScriptHandler.swift#L165) with the `walletSolProviderScripts` value set on each Tab in [TabManager.swift](https://github.com/brave/brave-ios/blob/solana-dapps/Client/Frontend/Browser/TabManager.swift#L509). The script is retrieved from BraveCore in a closure in [BrowserViewController.swift](https://github.com/brave/brave-ios/blob/solana-dapps/Client/Frontend/Browser/BrowserViewController.swift#L427). 

The solana internal script itself is this [internal script](https://github.com/brave/brave-core/blob/1.45.x/components/brave_wallet/resources/solana_provider_internal.js), which is being [removed in the next release (v1.46)](https://github.com/brave/brave-core/pull/15247).

stoletheminerals: @StephenHeaps we don't inject this if the wallet is locked, so a webpage can overwrite `_brave_solana.createPublickey` and `_brave_solana.createTransaction`. I'm not sure about the security implications, but we freeze the object here:
https://github.com/brave/brave-core/blob/803186400aa12f0d29016ad2f95b958c0f880eda/components/brave_wallet/resources/solana_provider_internal.js#L28
probably for a good reason. Is there a reason we don't inject at the document's start?

StephenHeaps: Core team advised to lazy-load this script due to performance concerns when loading the Solana Web3 library, some additional info in the [brave-core issue](https://github.com/brave/brave-browser/issues/22468).

stoletheminerals: Ok, so I checked on Desktop, and there is the same issue. I'll ask Anthony (can't mention him here for some reason) in the PR you linked

kdenhartog: Ok, so in this case it sounds like the script is trusted, but we've still got a few issues in terms of the object getting overwritten. In the case of `createPublicKey` that could present issues with the account being returned causing issues. I don't see a large impact here, but this is highly contextual to the dApp in question so it certainly could create a larger impact in the future.

In the case of `createTransaction` this seems like it could present a legitimate concern to modify a transaction before it's signed such that the user could be tricked into signing it later when believing it came as a legitimate dApp request. In this case, I can see legitimate harm coming from this if it were able to be overwritten while the wallet is locked and I think this should get fixed. It seems @stoletheminerals suggestion of injecting at the start seems like a safe bet, but given there's a performance impact here that's going to have to be worked around as well.

StephenHeaps: v1.46 of core removes the `window._brave_solana` property entirely in favour of injecting the minified js solanaWeb3 library (I have a [wip branch available here against 1.46](https://github.com/brave/brave-ios/compare/solana-dapps...solana-dapps-solanaWeb3Library#diff-6d62f27259cf433b454bdcd9bc43c8f83a8345487158bf3d635015164226cefeR245)). I _believe_ this change was made to address the mentioned issue, but we'll have to investigate if this is possible to bring in v1.45 of iOS without an uplift of that change.

stoletheminerals: @StephenHeaps, are we still going to lazy-load the minified solanaWeb3 lib? If so, I'm afraid the issue could persist.
@kdenhartog, do you have any attack vectors in mind for `createTransaction`?

StephenHeaps: Yeah the plan was to continue to lazy-load the minified solanaWeb3 library on 1.46 for the performance concerns (cc @darkdh). The current changes are calling the constructors directly through `solanaWeb3` instead of through `createPublickey` / `createTransaction` functions.

darkdh: The requiring webpacked solanaWeb3.js code will be evaluated on every page load because we use it when defining properties. We used to load it on start and get flagged by automatic performance monitoring (https://github.com/brave/brave-browser/issues/22468).
Now we don't pollute the global namespace with `_brave_solana` which means the website has no access to where we construct public key or transaction. Desktop and Android are doing that in v8 and I suggested iOS construct those in a way that website won't be able to interfere. It is to prevent various attack vectors toward  `window._brave_solana` while our SafeBuiltins can only protect it from prototype polluted. (Here are some examples of how to do that https://bravesoftware.slack.com/archives/C0414R6NUKT/p1663952287601889?thread_ts=1663886636.872799&cid=C0414R6NUKT).
Let me know if we want brave-core's [`window._brave_solana` removal](https://github.com/brave/brave-core/pull/15247) in 1.45 so iOS can release Solana dapps with it.

thypon: ^^ As @darkdh suggested, we should load in a way that `_brave_solana` is not globally accessible

StephenHeaps: Thanks @darkdh.
The iOS code changes for removing `_brave_solana` from global namespace are available here:
https://github.com/brave/brave-ios/compare/solana-dapps...solana-dapps-solanaWeb3Library
This branch was tested using a local build of `brave-core`, so will likely fail to build or run without a local build against `brave-core`'s `master` branch, but at least code can be viewed.

kdenhartog (approved):
I think this one is good to move forward for now as mentioned in DM. We've got a secondary issue https://github.com/brave/brave-ios/issues/6154 that will double check all this has been closed out properly and this will let the front end work progress forward so we can still hopefully get this into 1.45.x iOS release. See the [security issue](https://github.com/brave/security/issues/1040) for this for additional details.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
```diff
@@ -377,6 +419,66 @@ extension Tab: BraveWalletEventsListener {
   }
 }
 
+extension Tab: BraveWalletSolanaEventsListener {
+  func accountChangedEvent(_ account: String?) {
+    Task { @MainActor in
+      if let webView = webView {
+        let script: String
+        if let account = account {
+          script = "if (solanaWeb3) { window.solana.emit('accountChanged', new solanaWeb3.PublicKey('\(account)')) }"
```
iccub: should we sanitize account string here or unnecessary?

Client/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Paged/SolanaProviderScriptHandler.swift
```diff
@@ -0,0 +1,346 @@
...
+    return .init(encodedSerializedMsg: encodedSerializedMsg, signatures: signatures)
+  }
+  
+  private func buildErrorJson(status: BraveWallet.SolanaProviderError, errorMessage: String) -> String? {
+    JSONSerialization.jsObject(withNative: [Keys.code.rawValue: status.rawValue, Keys.message.rawValue: errorMessage])
+  }
+  
+  @MainActor private func emitConnectEvent(publicKey: String) async {
+    if let webView = tab?.webView {
+      let script = "window.solana.emit('connect', new solanaWeb3.PublicKey('\(publicKey)'))"
```
iccub: same question here whether public address sanitization would make sense here or not

Sources/BraveWallet/Panels/Sign Transaction/SignTransactionView.swift
```diff
@@ -0,0 +1,52 @@
...
+  var request: Request
+  var cryptoStore: CryptoStore
+  var onDismiss: () -> Void
+  
+  var navigationTitle: String {
+    switch request {
+    case .signTransaction:
+      return "Sign Transaction"
+    case .signAllTransactions:
+      return "Sign All Transactions"
```
iccub: Do we want these translated?

Sources/BraveWallet/Panels/Sign Transaction/SignTransactionView.swift
```diff
@@ -0,0 +1,52 @@
...
+              .signTransaction(approved: false, id: request.id, signature: nil, error: "User rejected the request")
+            )
+          case let .signAllTransactions(request):
+            cryptoStore.handleWebpageRequestResponse(
+              .signAllTransactions(approved: false, id: request.id, signatures: nil, error: "User rejected the request")
+            )
+          }
+          onDismiss()
+        }) {
+          Text("Reject")
```
iccub: Localized string

iccub (commented):
code looks good to me. only concerns i had were about public address sanitization in some methods

Sources/BraveWallet/Panels/Sign Transaction/SignTransactionView.swift
StephenHeaps: Yeah I will localize these here, but otherwise would be handled in #6005

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
StephenHeaps: The address here and in `SolanaProviderScriptHandler` (comment below) comes from core so it _should_ be fine, but given other comments on this piece I think it would help to use the `htmlEntityEncodedString` to be safe.

Sources/BraveWallet/Panels/Sign Transaction/SignTransactionView.swift
StephenHeaps: https://github.com/brave/brave-ios/pull/6050/commits/87993ce8e4d3855c9e0ac2f522d60c999d6d337f

Sources/BraveWallet/Panels/Sign Transaction/SignTransactionView.swift
StephenHeaps: https://github.com/brave/brave-ios/pull/6050/commits/87993ce8e4d3855c9e0ac2f522d60c999d6d337f

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
StephenHeaps: https://github.com/brave/brave-ios/pull/6050/commits/87993ce8e4d3855c9e0ac2f522d60c999d6d337f

iccub (approved):

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -589,6 +592,9 @@ extension BrowserViewController: WKNavigationDelegate {
         )
       }
       tab.updateEthereumProperties()
+      Task { @MainActor in
```
kylehickinson: `@MainActor` here is redundant

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
```diff
@@ -285,6 +324,9 @@ extension Tab: BraveWalletProviderDelegate {
   
   func clearSolanaConnectedAccounts() {
     tabDappStore.solConnectedAddresses = .init()
+    Task { @MainActor in
```
kylehickinson: `@MainActor` here is redundant

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
```diff
@@ -375,6 +417,66 @@ extension Tab: BraveWalletEventsListener {
   }
 }
 
+extension Tab: BraveWalletSolanaEventsListener {
+  func accountChangedEvent(_ account: String?) {
+    Task { @MainActor in
```
kylehickinson: `@MainActor` here is redundant

Client/Frontend/Browser/BrowserViewController/BrowserViewController+Wallet.swift
```diff
@@ -375,6 +417,66 @@ extension Tab: BraveWalletEventsListener {
...
+      await updateSolanaProperties()
+    }
+  }
+
+  func emitSolanaEvent(_ event: Web3ProviderEvent) {
+    guard Preferences.Wallet.defaultSolWallet.value == Preferences.Wallet.WalletType.brave.rawValue,
+          let webView = webView else {
+      return
+    }
+    Task { @MainActor in
```
kylehickinson: `@MainActor` here is redundant

Client/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Paged/SolanaProviderScriptHandler.swift
```diff
@@ -0,0 +1,344 @@
...
+  }()
+  
+  private weak var tab: Tab?
+  
+  init(tab: Tab) {
+    self.tab = tab
+  }
+  
+  static func shouldInjectWalletProvider(_ completion: @escaping (Bool) -> Void) {
+    BraveWallet.KeyringServiceFactory.get(privateMode: false)?
```
kylehickinson: 1. We probably want to pass in the correct private mode here
2. We will want to guard against the service and complete with false. Currently if `KeyringServiceFactory.get` returns nil you will never call completion

Client/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Paged/SolanaProviderScriptHandler.swift
```diff
@@ -0,0 +1,344 @@
...
+    var args: String?
+    
+    private enum CodingKeys: String, CodingKey {
+      case method
+      case args
+    }
+  }
+  
+  func userContentController(_ userContentController: WKUserContentController, didReceiveScriptMessage message: WKScriptMessage, replyHandler: @escaping (Any?, String?) -> Void) {
+    Task { @MainActor in
```
kylehickinson, line 104: Task is created for the entire body but the first await doesnt happen until you execute on the body method. To avoid more `Sendable` issues down the road, the Task should only surround the async required parts

Sources/BraveWallet/Crypto/Transactions/TransactionParser.swift
```diff
@@ -543,6 +554,10 @@ struct Eth721TransferDetails: Equatable {
   let tokenId: String
 }
 
+struct SolanaDappTxDetails: Equatable {
+  
+}
```
kylehickinson: empty struct?

Client/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Paged/SolanaProviderScriptHandler.swift
StephenHeaps: Removed in [c42e657](https://github.com/brave/brave-ios/pull/6050/commits/c42e657f7488976f3293763b0d159d3d15abebf6) as this function was not being used, similar for `EthereumProviderScriptHandler`.

Client/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Paged/SolanaProviderScriptHandler.swift
StephenHeaps: Fixed in [c42e657](https://github.com/brave/brave-ios/pull/6050/commits/c42e657f7488976f3293763b0d159d3d15abebf6)

Sources/BraveWallet/Crypto/Transactions/TransactionParser.swift
StephenHeaps: Removed in [c42e657](https://github.com/brave/brave-ios/pull/6050/commits/c42e657f7488976f3293763b0d159d3d15abebf6). Will be re-added (with properties) in PR to support this transaction type.

kylehickinson (approved):

nuo-xu (approved):

PR #6054, opened 2022-09-23
Fix #6011: Suppress 'switch to app' annoyances by default.

### LINKED ISSUE #6011: Suppress Mobile “Switch to App” Annoyances
### Description: 
Spec
https://bravesoftware.slack.com/archives/C03MMJ4CHEZ/p1661878378715489?thread_ts=1661878373.627779&cid=C03MMJ4CHEZ

Apply the switch to app suppression filter lists by default (this ticket).
Add a toggle to enable/disable the lists (future work - needs some UX thought).

iccub:
## Summary of Changes

This pull request fixes #6011 

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:

1. Fresh install
2. Go to settings -> Brave Shields & Privacy, verify that 'Block 'Switch to App' Notices` is enabled by default.
3. Go to reddit.com, verify that there's no 'switch to mobiel app' showing
4. Disable the setting, go to reddit again and see if the popup is visible

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Client/WebFilters/FilterList.swift
```diff
@@ -11,6 +11,9 @@ struct FilterList: Decodable, Identifiable {
     case uuid, title, componentId, description = "desc", urlString = "url"
   }
   
+  public static let mobileAnnoyancesComponentID = "ldbgldhcahahpffloggbbmjllggnkenk"
+  public static let mobileAnnoyancesUUID = "2F3DCE16-A19A-493C-A88F-2E110FBD37D6"
```
cuba: This is a good approach but I decided on a slightly different one. Perhaps we can put all the static uuids in one place. Mine are in `FilterListResourceDownloader` starting on line 118 here:

```
/// The uuid of the cookie consent notices filter list.
  ///
  /// This is a special filter list that has more accessible UI to control it
  public static let cookieConsentNoticesUUID = "AC023D22-AE88-4060-A978-4FEEEC4221693"
  /// The uuid of the "Remove Web annoyances" filter list
  ///
  /// This is a special filter list that has more accessible UI to control it
  public static let removeWebAnnoyancesUUID = "67E792D4-AE03-4D1A-9EDE-80E01C81F9B8"
```

If you prefer your approach maybe you can move my UUIDs to `FilterList`

Also I thought we were going to enable the web annoyances filter list so I added the `removeWebAnnoyancesUUID` but if we're not going to use it perhaps we should remove that one too. I can do that but if you will touch the code perhaps you don't mind doing it

Client/WebFilters/FilterListResourceDownloader.swift
```diff
@@ -265,11 +277,22 @@ public class FilterListResourceDownloader: ObservableObject {
     )
   }
   
+  /// This method allows us to enable selected lists by default for new users.
+  /// Make sure you use componentID to identify the filter list, as `uuid` will be deprecated in the future.
```
cuba, line 291: I should then make this change for my PR?

Client/WebFilters/FilterListResourceDownloader.swift
```diff
@@ -265,11 +277,22 @@ public class FilterListResourceDownloader: ObservableObject {
...
+  
   /// Load filter lists from the ad block service
   private func loadFilterLists(from regionalFilterLists: [AdblockFilterList], filterListSettings: [FilterListSetting]) -> [FilterList] {
     return regionalFilterLists.map { adBlockFilterList in
       let setting = filterListSettings.first(where: { $0.uuid == adBlockFilterList.uuid })
-      return FilterList(from: adBlockFilterList, isEnabled: setting?.isEnabled ?? false)
+      return FilterList(from: adBlockFilterList,
+                        isEnabled: setting?.isEnabled
+                        ?? newFilterListEnabledOverride(for: adBlockFilterList.componentId)
+                        ?? false)
```
cuba, line 305: I'm a little bit worried here: when I load filter lists for the first time I end up creating settings for all filter lists in the DB. I think what you do here is fine but just let's make sure this works fine.

cuba (changes requested):
Looks good...Mostly comments that are not strictly required except that we should remove the UUIDs and add a function that uses componentID

Client/WebFilters/FilterList.swift
cuba: As discussed this code will be available after this PR so its me that needs to sync up in my PR with this one. So Ignore this comment.

Client/WebFilters/FilterListResourceDownloader.swift
iccub: We decided to fix this separately or in the cookie-consent PR https://github.com/brave/brave-ios/pull/5839

cuba (approved):

PR #6094, opened 2022-10-03
Fix #5283: Option to switch to an open tab

### LINKED ISSUE #5283: Option to switch to an open tab
## Problem Description

Follow an external link to a site that opens in a new tab even if it is already open in another tab.

## Feature Overview

This option gives opportunity to follow internal and external links to an open tab instead of opening a new one while browsing in Normal Browsing Mode.

It will automatically open the in the existing tab, no settings required.

## Implementation Details 

The option for this settings will be under Settings-> Tabs.

- When enabled an _**external links**_ (**URL**) opened in Brave browser (Brave is default browser) if user is not on Private Browsing Mode will be first checked if that particular URL exists in any tab.

>  If it exists the tab will be automatically switched instead of being opened in a new tab.

soner-yuksel:
## Summary of Changes

This pull request fixes #5283

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:

For all the test enable Switch To an Open Tab setting.

Test Plan 1 External Link:

- Switch to Normal Browsing Mode
- Open an external link in Brave
- Open another link in another tab and leave it as active
- Close the browser
- Open the same external link outside the Brave
- Check if external link switches to the existing tab not open new tab with that url link

## Screenshots:

- Opening an external Link:

https://user-images.githubusercontent.com/6643505/193651030-cbea22a5-e0f6-488b-80e5-16281c7fdecf.MP4

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -849,11 +849,11 @@ class TabManager: NSObject {
     assert(Thread.isMainThread)
     
     let tab = allTabs.filter {
-      if let webViewURL = $0.webView?.url {
-        return  webViewURL.schemelessAbsoluteDisplayString == url.schemelessAbsoluteDisplayString
+      guard let webViewURL = $0.webView?.url else {
+        return  false
       }
       
-      return false
+      return webViewURL.schemelessAbsoluteDisplayString == url.schemelessAbsoluteDisplayString
     }.first
     
     return tab
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -847,8 +847,16 @@ class TabManager: NSObject {
 
   func getTabForURL(_ url: URL) -> Tab? {
     assert(Thread.isMainThread)
-
-    return allTabs.filter { $0.webView?.url == url }.first
+    
+    let tab = allTabs.filter {
+      if let webViewURL = $0.webView?.url {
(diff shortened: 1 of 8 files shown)


iccub (commented):
Code looks fine, but we need a design/product decision for this one

soner-yuksel:
> Code looks fine, but we need a design/product decision for this one

As requested removed the internal link criteria and only fixed the problem with external links and removed the setting entirely. I believe we can remove the label "blocked"

Client/Frontend/Browser/TabManager.swift
```diff
@@ -847,8 +847,16 @@ class TabManager: NSObject {
...
     assert(Thread.isMainThread)
-
-    return allTabs.filter { $0.webView?.url == url }.first
+    
+    let tab = allTabs.filter {
+      guard let webViewURL = $0.webView?.url else {
+        return  false
+      }
+      
+      return webViewURL.schemelessAbsoluteDisplayString == url.schemelessAbsoluteDisplayString
```
iccub, line 856: why is this change introduced, any side effects in other places?

soner-yuksel: This change is introduced because the older version (existing) doesnt work and the check for equality always fails.

So we fixed the conditional check here and also I added tests for it,

iccub (approved):

PR #6103, opened 2022-10-04
No Bug: Removing the condition of compiling filter list data

soner-yuksel:
Removing the condition of compiling AdBlockEngineManager and ContentBlockerManager before setup tabs can start

## Summary of Changes

## Submitter Checklist:

  - [ ] Light & dark mode

## Test Plan:

- Load the app and check no delay 

## Screenshots:

https://user-images.githubusercontent.com/6643505/193896111-464c6bd4-b170-4f41-ac4a-74cc422d05c9.MP4

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -532,13 +532,15 @@ public class BrowserViewController: UIViewController, BrowserViewControllerDeleg
       
       // Load cached data
       // This is done first because compileResources and compilePendingResource need their results
-      await FilterListResourceDownloader.shared.loadCachedData()
-      await AdblockResourceDownloader.shared.loadCachedData()
-      
+      async let filterListCache: Void = await FilterListResourceDownloader.shared.loadCachedData()
+      async let adblockResourceCache: Void = await AdblockResourceDownloader.shared.loadCachedData()
+      _ = await (filterListCache, adblockResourceCache)
+
       // Compile some ad-block data
-      await AdBlockEngineManager.shared.compileResources()
-      await ContentBlockerManager.shared.compilePendingResources()
-      
+      async let compiledResourcesCompile: Void = await AdBlockEngineManager.shared.compileResources()
+      async let pendingResourcesCompile: Void = await ContentBlockerManager.shared.compilePendingResources()
+      _ = await (compiledResourcesCompile, pendingResourcesCompile)
+  
       FilterListResourceDownloader.shared.start(with: self.braveCore.adblockService)
       await AdblockResourceDownloader.shared.startFetching()
       ContentBlockerManager.shared.startTimer()
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -162,7 +162,6 @@ public class BrowserViewController: UIViewController, BrowserViewControllerDeleg
 
   let downloadQueue = DownloadQueue()
 
-  fileprivate var contentBlockListCompiled: Bool = false
   private var cancellables: Set<AnyCancellable> = []
 
   let rewards: BraveRewards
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -526,22 +525,19 @@ public class BrowserViewController: UIViewController, BrowserViewControllerDeleg
     updateWidgetFavoritesData()
     
     Task { @MainActor in
+      self.setupTabs()
```
cuba, line 528: Since we are calling this first and on the main actor I think you can just call it outside the task now, no?

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -526,22 +525,19 @@ public class BrowserViewController: UIViewController, BrowserViewControllerDeleg
...
       await ContentBlockerManager.shared.loadBundledResources()
       await ContentBlockerManager.shared.loadCachedCompileResults()
       
       // Load cached data
       // This is done first because compileResources and compilePendingResource need their results
-      async let filterListCache: Void = await FilterListResourceDownloader.shared.loadCachedData()
-      async let adblockResourceCache: Void = await AdblockResourceDownloader.shared.loadCachedData()
-      _ = await (filterListCache, adblockResourceCache)
+      await FilterListResourceDownloader.shared.loadCachedData()
+      await AdblockResourceDownloader.shared.loadCachedData()
```
cuba: There may still be some benefit in doing this in parallel as we still want this all to happen as fast as possible so we can be faster than the user

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -526,22 +525,19 @@ public class BrowserViewController: UIViewController, BrowserViewControllerDeleg
...
       
       // Compile some ad-block data
-      async let compiledResourcesCompile: Void = await AdBlockEngineManager.shared.compileResources()
-      async let pendingResourcesCompile: Void = await ContentBlockerManager.shared.compilePendingResources()
-      _ = await (compiledResourcesCompile, pendingResourcesCompile)
-      
-      self.contentBlockListCompiled = true
-      self.setupTabs()
+      await AdBlockEngineManager.shared.compileResources()
+      await ContentBlockerManager.shared.compilePendingResources()
```
cuba: same here

cuba (changes requested):

Client/Frontend/Browser/BrowserViewController.swift
soner-yuksel: If I do that it is crashing at https://github.com/brave/brave-ios/blob/113cc3da2230fbffce13a7ad2884485fac389453/Client/Frontend/Browser/BrowserViewController.swift#L1425

Client/Frontend/Browser/BrowserViewController.swift
soner-yuksel: Fixed at https://github.com/brave/brave-ios/pull/6103/commits/c05dc1c0a9b1f7095e94c558b59137f1175a4038

Client/Frontend/Browser/BrowserViewController.swift
soner-yuksel: Fixed at https://github.com/brave/brave-ios/pull/6103/commits/c05dc1c0a9b1f7095e94c558b59137f1175a4038

Client/Frontend/Browser/BrowserViewController.swift
cuba: Maybe something wrong with how we load those views. Im ok with you keeping it this way then and we can fix this in another PR

cuba (approved):

PR #6144, opened 2022-10-11
No fix: Improve the quality of engine script injection on sub frames

cuba:
## Summary of Changes

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:

Pete has kindly provided the following test page: 
https://dev-pages.brave.software/filtering/scriptlets.html
Here you should see true for all values.

NOTE: Because filter lists and ad-block rules are loaded in the background when launching the app they may take a moment to kick in. This is a know and more general issue that requires a separate ticket. Please give a moment before testing these scripts.

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,7 +83,6 @@ node_modules
 !ThirdParty/**/node_modules
 
 # Brave
-scripts/
 BuildId.xcconfig
 
 BraveCore/BraveCore.xcframework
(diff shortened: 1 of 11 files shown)


.gitignore
```diff
@@ -81,7 +81,6 @@ MozBuildID.xcconfig
 node_modules
 
 # Brave
-scripts/
```
cuba, line 86: I'm not sure where this is needed but it interferes with the `Scripts_Danamic/Scripts` folder. I haven't seen any other script folder anywhere however if this folder is truly needed, we would just need to be more specific with the path (such as `/scripts`)

cuba:
So what is happening here:
Executing scripts is done so too late for the engine scripts to work effectively. The only way to effectively use the engine scripts is to inject them. However there are 2 problems with this:
1. We cannot target specific sub frames, only the main frame. So we wrap the scripts with a JS check to ensure they execute on the correct sub frame
2. Injecting sub-frame scripts causes a lot of script reloads as frames are loaded as the user scrolls, so we cached the subframe urls so we only need to reload them when a new subframe script is added.

cuba:
Merged as part of a larger PR

PR #6153, opened 2022-10-11
[No Bug]: Create shared directory in-case of a deletion

soner-yuksel:
Creating a shared container directory if somewhat the directory is removed.

This is added in case the shared directory will be deleted somehow, this should not be possible on a phone with no jailbreak applied.

Reported https://community.brave.com/t/ios-downloads-folder-gone/437019

## Summary of Changes

This pull request fixes #N/A

## Submitter Checklist:

  - [ ] Light & dark mode

## Test Plan:
N/A

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Providers/Profile.swift
+++ b/Client/Providers/Profile.swift
@@ -29,6 +29,16 @@ class ProfileFileAccessor: FileAccessor {
     var rootPath: String
     let sharedContainerIdentifier = AppInfo.sharedContainerIdentifier
     if let url = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: sharedContainerIdentifier) {
+      var isDirectory: ObjCBool = false
+
+      if !FileManager.default.fileExists(atPath: url.path, isDirectory: &isDirectory) {
+        do {
+          try FileManager.default.createDirectory(at: url, withIntermediateDirectories: true)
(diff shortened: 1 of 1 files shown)


Client/Providers/Profile.swift
```diff
@@ -29,6 +29,16 @@ class ProfileFileAccessor: FileAccessor {
...
+      var isDirectory: ObjCBool = false
+
+      if !FileManager.default.fileExists(atPath: url.path, isDirectory: &isDirectory) {
+        do {
+          try FileManager.default.createDirectory(at: url, withIntermediateDirectories: true)
+        } catch {
+          log.error("Unable to find the shared container directory and error while trying tpo create a new directory. ")
+        }
+      }
+
```
nuo-xu, line 41: how do we delete this directory?

soner-yuksel: You do not delete this directory. this directory should be automatically created (which it does) and should not be editable (which it is).

However for some users this directory doesn't exist, disappeared (no idea why and cant figure out), it is bandaid solution to this problem arising among the community.

Brandon-T (approved):

PR #6196, opened 2022-10-17
Fix #6195: Remove PlaylistDetectorScript and fix item Expiration

### LINKED ISSUE #6195: Playlist Items on Playlist Folders not loading properly and shows as Expired
### Description: 
- Sometimes when visiting a Playlist Folder page, the items are loading as expired.

### Steps to Reproduce 
  1. Visit https://playlist.bravesoftware.com/japan-sample/ for example
OR:
  1. Have a playlist item that is expired, and try to play it.

**Actual result:** 
- Item does not load or shows up as all expired.

**Expected result:**
- Item should load and not show as expired.

- Easily reproduced.

**Brave Version:** 
- 1.43

Brandon-T:
## Summary of Changes
- Remove `PlaylistDetector.js`, so only `Playlist.js` is ever loaded even for streams, expired streams, and playlist folders.
- Revert change to node's UUID-Tag. It should never be recreated.

This pull request fixes #6195

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:
- In ticket.

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Client/Frontend/Browser/Tab.swift
```diff
@@ -317,7 +317,6 @@ class Tab: NSObject {
       }
       let webView = TabWebView(frame: .zero, tab: self, configuration: configuration!, isPrivate: isPrivate)
       webView.delegate = self
-      configuration = nil
```
Brandon-T, line 320: **Note:** I do NOT know why we ever set the configuration here to `nil`. 
I have removed it as I tested and it has no negative impacts. Also Firefox either removed it or doesn't have it.
Plus it should NEVER be nil otherwise `createWebView` can be called TWICE which we do NOT want!

Client/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Sandboxed/PlaylistScript.js
```diff
@@ -228,18 +224,6 @@ window.__firefox__.includeOnce("Playlist", function($) {
         })
       }, 2000);  // Resolution of 1000ms is fine for us.
     }
-
-    function onReady(fn) {
```
Brandon-T, line 232: Function was not used anymore so removed it.

Client/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Sandboxed/PlaylistScript.js
```diff
@@ -214,10 +214,6 @@ window.__firefox__.includeOnce("Playlist", function($) {
     }
     
     function requestWhenIdleShim(fn) {
-      fn.toString = function() {
```
Brandon-T, line 217: Script is sandboxed. No need for this.

iccub (approved):
Works well for me :)

soner-yuksel:
Works well for me either, we should merge this solution and check(ask users) if this change will also solve problem with downloaded videos are showing expired.
If not we might need to change URL bookmark for that problem

soner-yuksel (approved):

StephenHeaps (approved):

PR #6249, opened 2022-10-26
Fix #6191: Display URL shows properly for Blobs and Session Restore of Blob URLs

### LINKED ISSUE #6191: [hackerone] Show correct location for `blob:` URLs
https://hackerone.com/reports/1735900
We show an incorrect location when a `blob:` URL is loaded.

Brandon-T:
## Security Review:
- https://github.com/brave/security/issues/1080

## Summary of Changes
- Changes to DisplayURL getter function which now returns the correct URL instead of blank, for display.
- Changes DisplayURL to return a proper URL when session restoring (otherwise it shows blank for blobs).

This pull request fixes #6191

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:
- Test session restore and make sure the URL in the bar is the exact same as before killing the app.
- Test hackerone blob restore and blob display in the URL bar.

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Tests/BraveSharedTests/NSURLExtensionsTests.swift
+++ b/Tests/BraveSharedTests/NSURLExtensionsTests.swift
@@ -725,12 +725,6 @@ class NSURLExtensionsTests: XCTestCase {
   }
 
   func testTypedDisplayString() {
-    let testURL1 = URL(string: "https://www.youtube.com")
-    let testURL2 = URL(string: "http://google.com")
-    let testURL3 = URL(string: "www.brave.com")
-    let testURL4 = URL(string: "http://brave.com/foo/")
-    let testURL5 = URL(string: "http://brave.com/foo")
-
     func checkDisplayURLString(testURL: URL?, displayString: String) {
       if let actual = testURL?.typedDisplayString {
         XCTAssertEqual(actual, displayString)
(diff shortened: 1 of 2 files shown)


soner-yuksel:
@Brandon-T can we have Test for this `blob` URLs? I think we have a test function already called `testTypedDisplayString` but it is missing cases for `file ` or `blob`

Tests/BraveSharedTests/NSURLExtensionsTests.swift
```diff
@@ -725,25 +725,55 @@ class NSURLExtensionsTests: XCTestCase {
...
+    let urlMap = [
+      URL(string: "https://www.youtube.com"): "https://www.youtube.com",
+      URL(string: "http://google.com"): "http://google.com",
+      URL(string: "www.brave.com"): "www.brave.com",
+      URL(string: "http://brave.com/foo/"): "http://brave.com/foo/",
+      URL(string: "http://brave.com/foo"): "http://brave.com/foo",
+      URL(string: "blob://http://brave.com/foo"): "blob://http//brave.com/foo",
+      URL(string: "blob://02C00302-CE62-4DAE-AD70-FDEE19594856"): "blob://02C00302-CE62-4DAE-AD70-FDEE19594856",
+      URL(string: "file:///Users/brave/documents/foo.txt"): "file://foo.txt",
+      URL(string: "file://http://brave.com/foo.txt"): "file://foo.txt",
```
stoletheminerals, line 770: Once you merge https://github.com/brave/brave-ios/pull/6250, will the `displayString` be different?

Brandon-T: No it will not be different. Once merged, it means the user cannot `type` file URLs into the URL bar.
But file URLs can still be displayed if the user tapped on a link to a file and it triggers a download or display of that file :)

stoletheminerals: But do we even support `file://` schemes on iOS? `http` to  `file` links won't work for example. How would a user navigate to `file://` urls in the first place?

Brandon-T: You are right. They can't. I forgot that `file://` is local :)

So for example, when we download files for the user, we do NOT expose the actual local file link, we only expose the file itself via the `Files` app on iOS (so we don't have code for this, and it's handled by the system).

In other words, it doesn't really matter if the user types a file URL into the URL bar, it'll do nothing and it will never load. That's because all files stored in the app is stored like:
`file:///var/mobile/applications/UUID-v4-here/data/Containers/Data/Application/application support/some/path/here/file.extension`

Real example (same format on device and simulator): `file:///Users/brandon/Library/Developer/CoreSimulator/Devices/24C9BFF4-BC8C-452E-A736-90E0824571CE/data/Containers/Data/Application/AC39D230-DC5A-48BD-AE55-7EEE10773851/Documents/Downloads/SomeFile.txt`

So unless the user knows that `UUID-v4`, it'll never work. It isn't possible for the user to ever get that file path with the UUID in Brave-iOS anyway, unless they can brute-force all UUID-v4s and get lucky.

I fixed this bug because it's super annoying that the browser attempts to load `file://` URLs, and it never loads, and time-out is too long, so the loader looks stuck. 

Also fixed it because `hackerone` :D

stoletheminerals (approved):

iccub (approved):

PR #6251, opened 2022-10-26
Fix #6219: Do not open external links in child tabs

### LINKED ISSUE #6219: [hackerone] Disable external schemes opening in child tabs
https://hackerone.com/reports/1744892

Brandon-T:
## Security Review
- https://github.com/brave/security/issues/1082

## Summary of Changes
- Changes the code to check if an external link such as `tel:` is opening in a child tab, and block it so it.

This pull request fixes #6219

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:
- Test the hacker one ticket and make sure external links such as `tel:` does not prompt the user from a second tab (child tab)

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
@@ -42,8 +42,17 @@ extension BrowserViewController {
   }
   
   fileprivate func handleExternalURL(_ url: URL,
-                                     isMainFrame: Bool,
+                                     tab: Tab?,
+                                     navigationAction: WKNavigationAction,
                                      openedURLCompletionHandler: ((Bool) -> Void)? = nil) {
+    
+    let isMainFrame = navigationAction.targetFrame?.isMainFrame == true
+    
+    // Do not open external links for child tabs automatically
+    // The user must tap on the link to open it.
+    if tab?.parent != nil && navigationAction.navigationType != .linkActivated {
+      return
+    }
     
     // We do not want certain schemes to be opened externally when called from subframes.
     if ["tel", "sms"].contains(url.scheme) && !isMainFrame {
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -179,7 +179,11 @@ extension BrowserViewController: WKNavigationDelegate {
     // First special case are some schemes that are about Calling. We prompt the user to confirm this action. This
     // gives us the exact same behaviour as Safari.
     if ["sms", "tel", "facetime", "facetime-audio"].contains(url.scheme) {
-      handleExternalURL(url, isMainFrame: navigationAction.targetFrame?.isMainFrame == true)
+      
+      // Do not allow opening external URLs from child tabs
+      if (webView as? TabWebView)?.tab?.parent == nil {
+        handleExternalURL(url, isMainFrame: navigationAction.targetFrame?.isMainFrame == true)
```
soner-yuksel: Since it is only used `WKNavigationDelegate`, maybe we can pass the tab instance to `handleExternalURL` function and do not add the check couple of times.

soner-yuksel (approved):

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
Brandon-T: Done.

iccub (approved):

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -42,9 +42,18 @@ extension BrowserViewController {
   }
   
   fileprivate func handleExternalURL(_ url: URL,
-                                     isMainFrame: Bool,
+                                     tab: Tab?,
```
stoletheminerals, line 45: Do we only call `handleExternalURL` for specific schemes? I tried some iOS built-in schemes, such as `shortcuts://` and `shareddocuments://`, those work in Safari but not Brave. Are we intentionally don't support it?

Brandon-T: I could fix this but I'm not sure if I should.
`handleExternalURL` is only called when a `page` tries to open an external URL or the user presses on a link and it attempts to open an external URL.

However, when you `type` into the URL-Bar/OmniBox, only `submitSearchText` is called which processes the query and determines if to send it to the default search engine, OR if to attempt to resolve it directly. Atm, this does NOT handle external URLs, and I'm not sure if it should. 

The reason I'm not sure is because Chrome and Firefox also do not handle this case. It only works in Safari.

stoletheminerals (approved):

PR #6300, opened 2022-11-01
Fix #6277 Detect unknown unicode in sign message requests and some improvement for misleading message 

### LINKED ISSUE #6277: Misleading Signing Request Message (BRA-Q322-7)
ref: 
https://github.com/brave/brave-browser/issues/24816

nuo-xu:
## Summary of Changes
We will detect unknown unicode in sign request message. (any unicode that is no inside [0, 127] would be considered as unknown)
When unknown unicode is detected, a corresponding warning will show up with a cta button that users can switch between original message and message in ascii-encoding.
We will also detect consecutive newline characters (2 or more consecutive newline characters). 
If those consecutive newline characters will push textview content size overflowing the textview frame size, another warning will show up and the message consecutive newline characters will be replaced with pilcrow sign with a number right after indicates how many newline characters are detected. 
When message contains both unknown characters and consecutive newline characters that will make message overflow the textbox, two warnings will show up. 
Note: the CTA button that to show message unknown unicodes will escape the pilcrow sign. 

This pull request fixes #6277

## Submitter Checklist:

  - [ ] Light & dark mode

## Test Plan:
1. Test message has unknown unicode (check warning message and CTA)
2. Test message has multiple consecutive newline characters that will overflow the textbox (check warning message and the pilcrow signs)
3. Test message has multiple consecutive newline characters that will NOT overflow the textbox (check there should not be a warning or pilcrow formatted)
4. Test message has both unknown unicode and consecutive newline characters that will overflow the textbox (check two warning messages; check the pilcrow sign; check warning CTA display correct message)
5. Test message has unknown unicode and consecutive newline characters that will NOT overflow the textbox (check one warning message and CTA)
6. Test a regular message (check no warning message)

## Screenshots:

https://user-images.githubusercontent.com/1187676/199269806-136751e4-9ef4-4212-8144-fc19ca67857e.mp4

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/BraveWallet/Panels/Signature Request/SignatureRequestView.swift
+++ b/Sources/BraveWallet/Panels/Signature Request/SignatureRequestView.swift
@@ -17,10 +17,13 @@ struct SignatureRequestView: View {
   var onDismiss: () -> Void
 
   @State private var requestIndex: Int = 0
+  @State private var renderUnknownUnicodes: Bool = false
+  @State private var needPilcrowFormatted: Bool = false
   @Environment(\.sizeCategory) private var sizeCategory
   @Environment(\.presentationMode) @Binding private var presentationMode
   @ScaledMetric private var blockieSize = 54
   private let maxBlockieSize: CGFloat = 108
+  private let staticTextViewHeight: CGFloat = 200
   
   private var currentRequest: BraveWallet.SignMessageRequest {
     requests[requestIndex]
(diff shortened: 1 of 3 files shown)


Sources/BraveWallet/Panels/Signature Request/SignatureRequestView.swift
```diff
@@ -180,6 +260,43 @@ struct SignatureRequestView: View {
...
+      if ci > 127 {
+        return true
+      }
+    }
+    return false
+  }
+  
+  var hasConsecutiveNewLines: Bool {
+    // return true if string has two or more consecutive newline chars
+    return range(of: "\\n{2,}", options: .regularExpression) != nil
```
stoletheminerals, line 297: If people start complain about pilcrows we can bump this to `3 or more`, but should be fine for now I think

StephenHeaps:
Small UI bug when showing 2+ different requests. One with the ASCII warning and one with the pilcrow warning view:
https://user-images.githubusercontent.com/5314553/199275325-4c064226-d186-48c2-b67b-9f6522372c1b.mov

I believe the ASCII should only show on one request, and the pilcrow on the other. Seems like the pilcrow warning message isn't being removed

thypon:
Should we have a `Show Original Message` if we mangle the spaces or the ASCII characters here?
Cc @stoletheminerals

stoletheminerals:
> Should we have a `Show Original Message` if we mangle the spaces or the ASCII characters here? Cc @stoletheminerals

We have this for non-ascii, probably a good idea to have a similar toggle for new lines

diracdeltas:
> > Should we have a `Show Original Message` if we mangle the spaces or the ASCII characters here? Cc @stoletheminerals
> 
> We have this for non-ascii, probably a good idea to have a similar toggle for new lines

need to be careful here so that if the original message actually contains pilcrows, they don't get replaced with newlines

stoletheminerals:
> > > Should we have a `Show Original Message` if we mangle the spaces or the ASCII characters here? Cc @stoletheminerals
> > 
> > 
> > We have this for non-ascii, probably a good idea to have a similar toggle for new lines
> 
> need to be careful here so that if the original message actually contains pilcrows, they don't get replaced with newlines

Yep. We don't need to replace pilcrows back to `/n`, we can just show the original message

stoletheminerals (approved):
Looks good to me. Thank you for the quick fix @nuo-xu

Sources/BraveWallet/Panels/Signature Request/SignatureRequestView.swift
```diff
@@ -128,6 +190,17 @@ struct SignatureRequestView: View {
     .navigationBarTitleDisplayMode(.inline)
     .foregroundColor(Color(.braveLabel))
     .background(Color(.braveGroupedBackground).edgesIgnoringSafeArea(.all))
+    .introspectTextView { textView in
+      // A flash to show users message is overflowing the text view (related to issue https://github.com/brave/brave-ios/issues/6277)
+      if showOrignalMessage[requestIndex] == true {
```
StephenHeaps, line 198: I believe this creates a tricky to repro UI bug where we show the warning when not needed, because `needPilcrowFormatted` is only updated when `showOriginalMessage[requestIndex]` is true.

STR:
1. Send request that requires warning message
2. Send request that requires no warning message
3. Open requests modal
4. Tap 'Show encoded message'
5. Tap 'Cancel' / 'Sign'

https://user-images.githubusercontent.com/5314553/199510393-df7756fa-a492-4c1a-b391-96e879413aa8.mov

I think resetting the 2 dictionaries when we approve/cancel (thus changing the indexes of the requests, which is used as the dictionary key) would resolve. Possibly can also use the request `id` as the dictionary key instead of the request index... they seem to be index based but don't change when cancelling/approving an earlier request.

Sources/BraveWallet/Panels/Signature Request/SignatureRequestView.swift
```diff
@@ -17,10 +17,14 @@ struct SignatureRequestView: View {
   var onDismiss: () -> Void
 
   @State private var requestIndex: Int = 0
+  @State private var needPilcrowFormatted: [Int: Bool] = [0: false]
+  @State private var showOrignalMessage: [Int: Bool] = [0: true]
```
StephenHeaps: nit: can we add a comment that the key here is the request index

StephenHeaps (approved):

PR #6302, opened 2022-11-02
Hold first tab load until sheilds are ready

### LINKED ISSUE #6261: Ensure shields are loaded before the first tab launches.
### Description: 
When launching the application (after it is terminated) shields may not be applied momentarily.

### Steps to Reproduce 
  1. Go to a website with visible blocking. For example https://dev-pages.brave.software/filtering/scriptlets.html or https://youtube.com (we block ads here)
  2. Kill the app
  3. Launch the app

**Actual result:** 
Shields not applied applied: 
1. https://dev-pages.brave.software/filtering/scriptlets.html does not show true for all values
2. Youtube shows ads when watching a video

**Expected result:**
Should apply shields right away (for that first tab):
1. https://dev-pages.brave.software/filtering/scriptlets.html should show true for all values
2. Youtube should not show ads when watching a video

Easily reproducible

**Brave Version:** 
1.45

**Device details:** 
N/A

### Additional Information
Caused by independently loading shields and the first tab. This is a race condition and usually the tab loads faster than the shields. Need to improve loading the shields (and block the first tab) from cache.
This is tricky because we don't want to block the app from launching the first tab (so it doesn't look slow) for the sake of loading shields

Youtube navigation doesn't solve the issue because it's a single page application. you need to refresh YouTube to solve the ads problem. Most people are not aware of this and stay and navigate YouTube which gives them ads all the time if this is the first page they go to after launching the app.

cuba:
## Summary of Changes

This pull request fixes #6261

Tab loading awaits the preparation of some ad block data. To achieve this I had to rework how the UI is loaded on the BVC so that the elements are loaded but only "enabled" once ABD is ready.

## Submitter Checklist:

  - [ ] Light & dark mode

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/App/iOS/Delegates/AppDelegate.swift
+++ b/App/iOS/Delegates/AppDelegate.swift
@@ -138,7 +138,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
       return true
     }
 
-    BrowserViewController.isFirstLaunch = Preferences.General.isFirstLaunch.value
     migration = Migration(braveCore: braveCore)
 
     // TODO: Downgrade to 14.5 once api becomes available.
--- a/App/iOS/Delegates/AppDelegate.swift
+++ b/App/iOS/Delegates/AppDelegate.swift
@@ -138,6 +138,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
       return true
     }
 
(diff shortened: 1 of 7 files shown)


cuba:
Merged here: https://github.com/brave/brave-ios/pull/6417

PR #6524, opened 2022-11-30
Fix #5789: Pin Certs with Chromium API

### LINKED ISSUE #5789: [security] Pin SSL certificates with Chromium API
### Description: 
SSL Pinning doesn't work for the domains listed in https://github.com/brave/brave-core/blob/master/chromium_src/net/tools/transport_security_state_generator/input_file_parsers.cc

### Steps to Reproduce 
  1. Navigate to https://ssl-pinning.someblog.org/

**Actual result:** 
Webpage loads without an SSL error
image

**Expected result:**
Webpage fails to load with an SSL error

Easily reproduced,

**Brave Version:** 

**Device details:** 

**Website problems only:**

Brandon-T:
## Security Review
- https://github.com/brave/security/issues/1127

## Summary of Changes
- Pin Certs with Chromium API
- Display appropriate error pages when pinning fails or certificate validation fails.

This pull request fixes #5789

## Screenshots
image

## Submitter Checklist:

  - [x] Light & dark mode

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
@@ -485,8 +485,7 @@ extension BrowserViewController: WKNavigationDelegate {
     return .allow
   }
 
-  // When this function is moved to Async-Await, do NOT forget to move the `TabManagerNavDelegate` to async-await!
-  public func webView(_ webView: WKWebView, didReceive challenge: URLAuthenticationChallenge, completionHandler: @escaping (URLSession.AuthChallengeDisposition, URLCredential?) -> Void) {
+  public func webView(_ webView: WKWebView, respondTo challenge: URLAuthenticationChallenge) async -> (URLSession.AuthChallengeDisposition, URLCredential?) {
 
     // If this is a certificate challenge, see if the certificate has previously been
     // accepted by the user.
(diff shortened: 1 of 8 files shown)


Brandon-T:
Added Unit tests.

stoletheminerals (changes requested):
We should not allow a user to dismiss the error and proceed

image

Brandon-T:
> We should not allow a user to dismiss the error and proceed
> 
> image

Fixed

image

stoletheminerals:
Sorry @Brandon-T I missed this, the error text is not correct as well. The cert IS trusted by the OS, it's just not trusted by Brave because Brave expects a specific cert. Chrome has NET::ERR_SSL_PINNED_KEY_NOT_IN_CERT_CHAIN error for this case, can we show something similar on iOS?

Brandon-T:
> Sorry @Brandon-T I missed this, the error text is not correct as well. The cert IS trusted by the OS, it's just not trusted by Brave because Brave expects a specific cert. Chrome has NET::ERR_SSL_PINNED_KEY_NOT_IN_CERT_CHAIN error for this case, can we show something similar on iOS?

Hmm I see. I can add a special error code like Chromium and the same error message as on desktop.

Will do it shortly.

Brandon-T:
image

@stoletheminerals

stoletheminerals (approved):

Client/Frontend/Browser/Authenticator.swift
```diff
@@ -11,6 +11,7 @@ import os.log
 class Authenticator {
   fileprivate static let maxAuthenticationAttempts = 3
 
+  @MainActor
```
iccub: why do we need main actor to handle it now?

Client/Frontend/Browser/Interstitial Pages/CertificateErrorPageHandler.swift
```diff
@@ -7,6 +7,10 @@ import Foundation
 import Shared
 import BraveShared
 
+extension CFNetworkErrors {
+  static let braveCertificatePinningFailed = CFNetworkErrors(rawValue: Int32.min)!
```
iccub, line 11: is this safe? can't we use an enum's value `CFNetworkErrors.kXXXX` or something

Client/Frontend/Browser/TabManagerNavDelegate.swift
```diff
@@ -41,18 +41,19 @@ class TabManagerNavDelegate: NSObject, WKNavigationDelegate {
...
     guard let firstAuthenticatingDelegate = authenticatingDelegates.first else {
-      return completionHandler(.performDefaultHandling, nil)
+      return (.performDefaultHandling, nil)
     }
 
-    firstAuthenticatingDelegate.webView?(webView, didReceive: challenge) { (disposition, credential) in
-      completionHandler(disposition, credential)
-    }
+    // Do NOT change to `delegate.webView?(....)` the optional operator makes async-await calls crash the compiler atm!
+    // It must be force-unwrapped at the time of writing `January 17th, 2023`.
```
iccub, line 54: this comment still valid?

Client/Frontend/Browser/Interstitial Pages/CertificateErrorPageHandler.swift
Brandon-T: No because we don't know what is the exact range of ALL network errors. Sometimes WebKit can throw errors outside of this range :(
So we use an extreme number so it doesn't clash.

Client/Frontend/Browser/TabManagerNavDelegate.swift
Brandon-T: Still valid as of January 17th. We have this comment on all other WebKit async delegate methods as well

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -529,45 +529,87 @@ extension BrowserViewController: WKNavigationDelegate {
     return .allow
   }
 
-  // When this function is moved to Async-Await, do NOT forget to move the `TabManagerNavDelegate` to async-await!
-  public func webView(_ webView: WKWebView, didReceive challenge: URLAuthenticationChallenge, completionHandler: @escaping (URLSession.AuthChallengeDisposition, URLCredential?) -> Void) {
+  @MainActor
+  public func webView(_ webView: WKWebView, respondTo challenge: URLAuthenticationChallenge) async -> (URLSession.AuthChallengeDisposition, URLCredential?) {
```
kylehickinson: Do we need to execute this on MainActor?

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -529,45 +529,87 @@ extension BrowserViewController: WKNavigationDelegate {
...
+      return (.useCredential, URLCredential(trust: trust))
+    }
+    
+    // Certificate Pinning
+    if challenge.protectionSpace.authenticationMethod == NSURLAuthenticationMethodServerTrust {
+      if let serverTrust = challenge.protectionSpace.serverTrust {
+        let host = challenge.protectionSpace.host
+        let port = challenge.protectionSpace.port
+        
+        return await Task.detached(priority: .userInitiated) {
```
kylehickinson: Why detach? Is `BraveCertificateUtility.verifyTrust` a heavy function?

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -529,45 +529,87 @@ extension BrowserViewController: WKNavigationDelegate {
...
+            return (.useCredential, URLCredential(trust: serverTrust))
+          }
+          
+          // Cert is valid and should not be pinned
+          // Let the system handle it and we'll show an error if the system cannot validate it
+          if result == Int32.min {
+            return (.performDefaultHandling, nil)
+          }
+          
+          return await Task { @MainActor in
```
kylehickinson: Why another Task inside? To access `tab(for:)`?

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -529,45 +529,87 @@ extension BrowserViewController: WKNavigationDelegate {
...
+            // Cert is invalid and cannot be pinned
+            Logger.module.error("CERTIFICATE_INVALID")
+            let errorCode = CFNetworkErrors.braveCertificatePinningFailed.rawValue
+            
+            let underlyingError = NSError(domain: kCFErrorDomainCFNetwork as String,
+                                          code: Int(errorCode),
+                                          userInfo: ["_kCFStreamErrorCodeKey": Int(errorCode)])
```
kylehickinson: Private API key?

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -529,45 +529,87 @@ extension BrowserViewController: WKNavigationDelegate {
...
+            
+            let error = NSError(domain: kCFErrorDomainCFNetwork as String,
+                                code: Int(errorCode),
+                                userInfo: [NSURLErrorFailingURLErrorKey: webView.url as Any,
+                                           "NSErrorPeerCertificateChainKey": certificateChain,
```
kylehickinson: Private API key?

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
Brandon-T: No but the old WebKit code with callbacks, calls it on `Main` like all the other WebKit delegates. So I made it the same as before and it matches all the other async-await delegates.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
Brandon-T: It has to be called on a non-main queue because `verifyTrust` also uses `SecTrustEvaluate` which is required to be called on a non-main queue.

If I don't, it will give the usual purple error in Xcode

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
Brandon-T: Yeah it's for the `tab(for:)` access.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
Brandon-T: Yeah. It is used in our certificate pinning. This is the only way known to get certificate error codes :(

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
Brandon-T: Yeah. It's the only way to get the certificate chain from `NSError`. Fortunately, I'm the one setting it here, so even if Apple removes this key, it won't break anything as I'm the one setting it, and reading it back elsewhere in the code (certificate viewer).

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
kylehickinson: Since we're changing a majority of the logic in here let's just dump the main actor isolation here and then dump the `Task.detached` part

Brandon-T: Fixed

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -529,46 +529,87 @@ extension BrowserViewController: WKNavigationDelegate {
...
+                            code: Int(errorCode),
+                            userInfo: [NSURLErrorFailingURLErrorKey: webView.url as Any,
+                                       "NSErrorPeerCertificateChainKey": certificateChain,
+                                               NSUnderlyingErrorKey: underlyingError])
+          
+        return await Task { @MainActor in
+          // Handle the error later in `didFailProvisionalNavigation`
+          self.tab(for: webView)?.sslPinningError = error
+          return (.cancelAuthenticationChallenge, nil)
+        }.value
```
kylehickinson: minor nit: since you aren't using a value inside the task in the return value you can just use `MainActor.run` to execute this code and return like normal, e.g.:

```swift
await MainActor.run {
  self.tab(for: webView)?.sslPinningError = error
}
return (.cancelAuthenticationChallenge, nil)
```

Brandon-T: Fixed

kylehickinson (approved):

Issue #6776, opened 2023-01-16
`Duplicate Active Tab` opens a blank page when in `Reader Mode` 


stoletheminerals:
### Description: 
When a web page opened in `Reader Mode` pressing `Duplicate Active Tab` opens a blank page instead of the original url

### Steps to Reproduce 
  1. Go to https://en.wikipedia.org/wiki/Software
  2. Activate Reader Mode
  3. Long press on toolbar and press Duplicate Active Tab

### Actual result: 
A blank page is opened

### Expected result:
A reader mode version of https://en.wikipedia.org/wiki/Software should be opened

### Reproduces how often: [Easily reproduced, Intermittent Issue]
Always (kinda?)

### Brave Version: 
The latest build from the AppStore (1.46.2 (23.1.10.20)) as well as a local `development` build 
- Can you reproduce this issue with the most recent build from TestFlight? 
- Can you reproduce this issue with the previous version of the current build from TestFlight? 
- Can you reproduce this issue with the current build from AppStore? 

### Device details: 
iPhone 14 Pro (iOS 16.2) / iPad Pro 

### Website problems only:
- did you check with Brave Shields down? 
- did you check in Safari/Firefox (WkWebView-based browsers)? 

### Additional Information
So last week I was checking one of the recent PRs and found that `Duplicate Active Tab` doesn't work for tabs in reader mode. I then tested it on AppStore build and it was working ok there, so I thought it's just a PR issue. 
I ran few tests today and it looks like I can reproduce it on all devices, even the old 1.45.* version.
When pressing Duplicate Active Tab Brave throws an error:
`2023-01-16 22:40:05.058660+0000 Client[50217:130482771] [Brave] Denying unprivileged request: http://localhost:6596/reader-mode/page?url=https://en.m.wikipedia.org/wiki/Software`
Seems like it's trying to open GCDWebServer's URL and fails here https://github.com/brave/brave-ios/blob/1434ecb55e07848f0d10ad6284854ddbaf852d70/Client/Frontend/Browser/BrowserViewController/BrowserViewController%2BWKNavigationDelegate.swift#L166
and here
https://github.com/brave/brave-ios/blob/1434ecb55e07848f0d10ad6284854ddbaf852d70/Sources/Shared/Extensions/URLExtensions.swift#L525

Brandon-T:
Fixed inadvertently in: https://github.com/brave/brave-ios/issues/6776

Uni-verse:
[...]

PR #6777, opened 2023-01-16
Fix #6796: Reader-Mode uses InternalURL and WKURLSchemeHandler

### LINKED ISSUE #6796: [Bug] - ReaderMode should not use GCDWebServer and instead use WKURLSchemeHandler
### Description: 
- Like all other interstitial pages, ReaderMode needs to use `WKURLSchemehandler` instead of `GCDWebServer`

### Steps to Reproduce 
  1. N/A See Internal Ticket: https://github.com/brave/internal/issues/919

### Actual result: 
- ReaderMode uses localhost URL

### Expected result:
- ReaderMode should not use localhost URL

- Easily Reproduced

### Brave Version: 

- ALL versions

### Device details: 
- All devices

Brandon-T:
## Summary of Changes
- Change Reader-mode to use `InternalURL` and `WKURLSchemeHandler` instead of `GCDWebServer`

This pull request fixes https://github.com/brave/internal/issues/919 and https://github.com/brave/brave-ios/issues/6796 and https://github.com/brave/brave-ios/issues/6776

## Submitter Checklist:

  - [x] Light & dark mode

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


stoletheminerals:
Thanks @Brandon-T, I'll take a look.
I guess it fixes https://github.com/brave/brave-ios/issues/6776 as well  in a way that there is no `Duplicate Active Tab` button anymore. But was it intentional?

iccub (approved):

soner-yuksel (approved):

Brandon-T:
> Thanks @Brandon-T, I'll take a look. I guess it fixes #6776 as well in a way that there is no `Duplicate Active Tab` button anymore. But was it intentional?

Inadvertently fixed it. Turns out that fixing `isWebPage` fixed it because we don't want the user to be able to visit `InternalURL` pages manually. These pages are generated via our UI or error handlers.
So you cannot duplicate a reader-mode page with this PR. We could of course add it back, but I don't think we should anyway. The previous code was allowing users to duplicate error pages too :(

stoletheminerals (approved):

PR #6857, opened 2023-02-01
Fix #6841: Add custom filter list urls

### LINKED ISSUE #6841: Add custom filter list URLs
### Description: 
Add the ability to add custom filter lists similar to desktop found in the following page:

Screenshot 2023-01-26 at 3 28 34 PM

cuba:
## Summary of Changes
This PR adds the ability to add custom filter list URLs. Also removed a lot unnecessary core for `ContentBlockerManager`
This should also improve the load time of the application as we don't compile rule lists unless they are necessary

This pull request fixes #6841 

Security ticket: https://github.com/brave/security/issues/1193

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:

You may use this filter list for testing: 
https://gist.githubusercontent.com/cuba/f13fce20b8fbd043ac8174e39ba16788/raw/af09db2fbda948dcf7a90781716c51f5e36f2237/TestAdBlockRules.txt
- It will remove a script from https://twitch.tv and will result in the blocked count to be 1 (testing script blocking)
- It will also remove the header in https://example.com (testing cosmetic filtering)

With shields enabled:
1. Test that all values return true for the following page: https://dev-pages.brave.software/filtering/scriptlets.html
2. Test that all values return true and blocked counts is 19 for the following page: https://test-pages.privacytests.org/tracking_content.html

## Screenshots:

Filter lists screen
Simulator Screen Shot - iPhone 14 Pro - 2023-02-10 at 13 32 17
Edit mode
Simulator Screen Shot - iPhone 14 Pro - 2023-02-10 at 13 32 21
Add filter list screen
Simulator Screen Shot - iPhone 14 Pro - 2023-02-10 at 13 32 26
Add filter list screen with keyboard open
Simulator Screen Shot - iPhone 14 Pro - 2023-02-10 at 13 35 34
Add filter list screen with entered url
Simulator Screen Shot - iPhone 14 Pro - 2023-02-10 at 13 36 18
Dark mode
Simulator Screen Shot - iPhone 14 Pro - 2023-02-10 at 14 42 01
iPad
Simulator Screen Shot - iPad Pro (12 9-inch) (6th generation) - 2023-02-10 at 14 44 18
Simulator Screen Shot - iPad Pro (12 9-inch) (6th generation) - 2023-02-10 at 14 44 22
Simulator Screen Shot - iPad Pro (12 9-inch) (6th generation) - 2023-02-10 at 14 44 33

Blocked content
Simulator Screen Shot - iPad Pro (12 9-inch) (6th generation) - 2023-02-10 at 14 44 45
Simulator Screen Shot - iPhone 14 Pro - 2023-02-10 at 14 47 25

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
```diff
@@ -0,0 +1,96 @@
...
+              presentationMode.wrappedValue.dismiss()
+            }
+          }
+        }
+    }
+  }
+  
+  private func handleOnSubmit() {
+    guard !newURLInput.isEmpty else { return }
+    guard let url = URL(string: newURLInput) else {
```
stoletheminerals, line 82: Do we accept non-https URLs on Desktop? I think it's a good idea to check for HTTPS URLs here

Sources/Brave/WebFilters/AdblockResourceDownloader.swift
```diff
@@ -71,36 +85,76 @@ public actor AdblockResourceDownloader: Sendable {
...
+        await handle(downloadResult: downloadResult, for: resource)
+      } else {
+        switch resource {
+        case .genericContentBlockingBehaviors:
+          do {
+            // This is a special case which we load a bundled file if the downloaded file is not present
+            try await ContentBlockerManager.shared.compileBundledRuleList(for: .blockAds)
+          } catch {
+            assertionFailure("A bundled file should not fail to compile")
+            ContentBlockerManager.log.error("Failed to compile bundled content blocker resource: \(error.localizedDescription)")
```
stoletheminerals: Won't `assertionFailure ` prevent the `log.error` call? Should we call `log.error` first?

Sources/Brave/WebFilters/ContentBlocker/ContentBlockerManager.swift
```diff
@@ -10,466 +10,307 @@ import Shared
...
-          "result: \(resultString)"
-        ].joined(separator: ", ")
-        
-        return ["{", resourcesDebugString, "}"].joined()
-      }.joined(separator: ", ")
-
-    Self.log.debug("Compiled \(resources.count, privacy: .public) additional block list resources: \(resourcesString, privacy: .public))")
+  private func punycodeConversion(domains: [String]) -> [String] {
+    return domains.compactMap { domain -> String? in
+      // TODO: Figure out how to convert this to punycode
```
stoletheminerals: Can we use Chromium's encoder? cc @Brandon-T

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
cuba: @kylehickinson do we have app transport security enabled for http? I don't do any specific http blocking but I thought we have it blocked globally but I'd be glad to add something specific to this.

Sources/Brave/WebFilters/AdblockResourceDownloader.swift
cuba: assertion failure is for debug only...I want it to crash in debug because the only way it will fail is if we messed up the bundled file and it should be fixed before going to production. In production this will definitely not crash and the log is there mostly in case someone changes the file without building or something pretty rare.

Sources/Brave/WebFilters/ContentBlocker/ContentBlockerManager.swift
cuba: ahh sorry this todo can be removed. I am using an encoder from brave core already

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
iccub: I don't think there's any http blocking, better to block manually here imo

cuba: Ok I checked for https. though I will have to still get the correct error copy here and display the message. for now its just blocked.

Sources/Brave/Frontend/Browser/LinkPreviewViewController.swift
```diff
@@ -44,14 +44,11 @@ class LinkPreviewViewController: UIViewController {
...
-    let domain = Domain.getOrCreate(forUrl: url, persistent: !PrivateBrowsingManager.shared.isPrivateBrowsing)
-    let compiledRuleTypes = ContentBlockerManager.shared.compiledRuleTypes(for: domain)
-    
-    Task(priority: .userInitiated) {
-      let ruleLists = await ContentBlockerManager.shared.loadRuleLists(for: Set(compiledRuleTypes.map({ $0.ruleType })))
-      for ruleList in ruleLists {
-        webView.configuration.userContentController.add(ruleList)
-      }
+    // Add rule lists for this page
+    Task { @MainActor in
```
iccub: MainActor not needed here I think, UIViewControllers are main actor by default

Sources/Brave/WebFilters/CustomFilterListStorage.swift
```diff
@@ -0,0 +1,50 @@
...
+    self.filterListsURLs = []
+  }
+  
+  @MainActor func loadCachedFilterLists() {
+    let settings = CustomFilterListSetting.loadAllSettings(fromMemory: !persistChanges)
+    
+    self.filterListsURLs = settings.map { setting in
+      let resource = setting.resource
+      let date = try? ResourceDownloader.creationDate(for: resource)
+      return FilterListCustomURL(setting: setting, downloadStatus: date != nil ? .downloaded(date!) : .pending)
```
iccub: can we avoid force unwrap here and maybe handle the error instead of convert it to nil with `try?`

Sources/Brave/WebFilters/FilterListCustomURL.swift
```diff
@@ -0,0 +1,42 @@
+// Copyright 2022 The Brave Authors. All rights reserved.
```
iccub: 2023

Sources/Brave/WebFilters/FilterListCustomURLDownloader.swift
```diff
@@ -0,0 +1,164 @@
+// Copyright 2022 The Brave Authors. All rights reserved.
```
iccub: 2023

Sources/Data/models/CustomFilterListSetting.swift
```diff
@@ -0,0 +1,75 @@
+// Copyright 2022 The Brave Authors. All rights reserved.
```
iccub: 2023

iccub (changes requested):
minor comments only

Sources/Brave/WebFilters/CustomFilterListStorage.swift
cuba: well the `!= nil` and `!` are on the same same line...But I can do this:

```
     if let date = date {
        return FilterListCustomURL(setting: setting, downloadStatus: .downloaded(date))
      } else {
        return FilterListCustomURL(setting: setting, .pending)
      }
```

iccub: you can catch the error instead

App/iOS/Delegates/AppDelegate.swift
```diff
@@ -338,7 +338,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
       UrpLog.log("Failed to initialize user referral program")
     }
 
-    DebouncingResourceDownloader.shared.startLoading()
```
iccub: Why is debouncing code removed in this PR?

App/iOS/Delegates/SceneDelegate.swift
```diff
@@ -217,11 +217,6 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
   }
 
   func sceneWillEnterForeground(_ scene: UIScene) {
-    // The reason we need to call this method here instead of `applicationDidBecomeActive`
-    // is that this method is only invoked whenever the application is entering the foreground where as
-    // `applicationDidBecomeActive` will get called whenever the Touch ID authentication overlay disappears.
-    DebouncingResourceDownloader.shared.startLoading()
```
iccub: same q here

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
```diff
@@ -0,0 +1,101 @@
...
+struct FilterListAddURLView: View {
+  @ObservedObject private var customFilterListStorage = CustomFilterListStorage.shared
+  @SwiftUI.Environment(\.presentationMode) private var presentationMode: Binding<PresentationMode>
+  @State private var newURLInput: String = ""
+  
+  private var textField: some View {
+    TextField(Strings.filterListsEnterFilterListURL, text: $newURLInput)
+      .keyboardType(.URL)
+      .textContentType(.URL)
+      .autocapitalization(.none)
```
iccub, line 25: We probably need to append `.disableAutocorrection(true)` as well

iccub: or `. autocorrectionDisabled()`

App/iOS/Delegates/AppDelegate.swift
cuba: Sorry I had some cleanup here that wasn't strictly necessary. I moved it to its own PR here: https://github.com/brave/brave-ios/pull/7102

App/iOS/Delegates/SceneDelegate.swift
cuba: Same: https://github.com/brave/brave-ios/pull/7102

Sources/Brave/WebFilters/CustomFilterListStorage.swift
cuba: Yes but the catch won't resolve the nil, it can still be nil.

```
  static func creationDate(for resource: Resource) throws -> Date? {
    guard let fileURL = downloadedFileURL(for: resource) else { return nil }
    let fileAttributes = try FileManager.default.attributesOfItem(atPath: fileURL.path)
    return fileAttributes[.creationDate] as? Date
  }
```

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
cuba: Done

App/iOS/Delegates/AppDelegate.swift
kylehickinson: This is still removed here despite the https://github.com/brave/brave-ios/pull/7102 being open

App/iOS/Delegates/SceneDelegate.swift
kylehickinson: Same here

Sources/Brave/Frontend/Browser/LinkPreviewViewController.swift
```diff
@@ -44,14 +44,11 @@ class LinkPreviewViewController: UIViewController {
...
+    // Add rule lists for this page
+    Task {
+      let domain = Domain.getOrCreate(forUrl: url, persistent: !PrivateBrowsingManager.shared.isPrivateBrowsing)
       let ruleLists = await ContentBlockerManager.shared.ruleLists(for: domain)
-      
-      for ruleList in ruleLists {
-        webView.configuration.userContentController.add(ruleList)
-      }
+      ruleLists.forEach { webView.configuration.userContentController.add($0) }
     }
```
kylehickinson: This change seems unnecessary

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
```diff
@@ -0,0 +1,102 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import SwiftUI
+import Strings
+import DesignSystem
+import BraveUI
+
+struct FilterListAddURLView: View {
+  @ObservedObject private var customFilterListStorage = CustomFilterListStorage.shared
+  @SwiftUI.Environment(\.presentationMode) private var presentationMode: Binding<PresentationMode>
```
kylehickinson: Remove `SwiftUI` prefix, change it to `@Binding`

```swift
@Environment(\.presentationMode) @Binding private var presentationMode
```

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
```diff
@@ -0,0 +1,102 @@
...
+  }
+  
+  var body: some View {
+    NavigationView {
+      List {
+        Section(content: {
+          VStack {
+            textField
+              .submitLabel(SubmitLabel.done)
+          }.listRowBackground(Color(.secondaryBraveGroupedBackground))
```
kylehickinson: No reason to be embedded inside a `VStack`

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
```diff
@@ -0,0 +1,102 @@
...
+        }, header: {
+          Text(Strings.customFilterListURL)
+            .textCase(.uppercase)
```
kylehickinson: Isn't uppercase the default?

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
```diff
@@ -0,0 +1,102 @@
...
+        }, footer: {
+          VStack(alignment: .leading, spacing: 8) {
+            Text(Strings.addCustomFilterListDescription)
+            Text(Strings.addCustomFilterListWarning).bold() + Text(Strings.addCustomFilterListWarning2)
```
kylehickinson: No space between the text, but now that we're iOS 15+ just combine these into 1 Text and use markdown. Just make sure to do `LocalizedStringKey(Strings. addCustomFilterListWarning)` so it is parsed as markdown

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
```diff
@@ -0,0 +1,102 @@
...
+              handleOnSubmit()
+            }.disabled(newURLInput.isEmpty)
+          }
+          
+          ToolbarItemGroup(placement: .cancellationAction) {
+            Button(Strings.CancelString) {
+              presentationMode.wrappedValue.dismiss()
+            }
+          }
+        }
```
kylehickinson: Seems like its tabbed in 1 tab too much

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
```diff
@@ -0,0 +1,102 @@
...
+            Text(Strings.addCustomFilterListDescription)
+            Text(Strings.addCustomFilterListWarning).bold() + Text(Strings.addCustomFilterListWarning2)
+          }.padding(.top, 16)
+        })
+      }
+        .listBackgroundColor(Color(UIColor.braveGroupedBackground))
+        .navigationTitle(Strings.customFilterList)
+        .navigationBarTitleDisplayMode(.inline)
+        .toolbar {
+          ToolbarItemGroup(placement: .primaryAction) {
```
kylehickinson: `primaryAction` → `confirmationAction`

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
```diff
@@ -0,0 +1,102 @@
...
+  }
+  
+  private func handleOnSubmit() {
+    guard !newURLInput.isEmpty else { return }
+    guard let url = URL(string: newURLInput) else {
+      // Show invalid URL error
+      return
+    }
+    guard url.scheme == "https" else {
+      // Show invalid scheme error
```
kylehickinson: Are these supposed to be TODOs? They don't do anything, just return

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
```diff
@@ -0,0 +1,102 @@
...
+      let customURL = FilterListCustomURL(
+        externalURL: url, isEnabled: true,
+        inMemory: !customFilterListStorage.persistChanges
+      )
+      
+      customFilterListStorage.filterListsURLs.append(customURL)
+      
+      await FilterListCustomURLDownloader.shared.startFetching(
+        filterListCustomURL: customURL
+      )
```
kylehickinson, line 108: No loader while it downloads?

Sources/Brave/Frontend/Settings/FilterListsView.swift
```diff
@@ -11,31 +11,143 @@ import BraveUI
...
+                  .foregroundColor(Color(.bravePrimary))
+                  .truncationMode(.middle)
+                  .lineLimit(1)
+                
+                switch filterListURL.downloadStatus {
+                case .downloaded(let downloadDate):
+                  Text([
+                    Strings.filterListsLastUpdatedLabel,
+                    dateFormatter.localizedString(for: downloadDate, relativeTo: Date())].joined(separator: ": "))
+                  .font(.caption)
```
kylehickinson: 1. `Last Updated: in 0 seconds` when you first add it looks like a bug. Use [`Text(_:style:)`](https://developer.apple.com/documentation/swiftui/text/init(_:style:)) instead so it updates live 
2. Remove the colon

image

Sources/Brave/Frontend/Settings/FilterListsView.swift
```diff
@@ -11,31 +11,143 @@ import BraveUI
...
   var body: some View {
     List {
+      // `onSubmit` doesn't work with iOS 14 so we don't support custom urls on iOS 14
       Section {
-        ForEach($downloader.filterLists) { $filterList in
+        ForEach($customFilterListStorage.filterListsURLs) { $filterListURL in
+          VStack(alignment: .leading, spacing: 4) {
+            Toggle(isOn: $filterListURL.setting.isEnabled) {
+              VStack(alignment: .leading, spacing: 4) {
+                Text(filterListURL.title ?? filterListURL.setting.externalURL.absoluteDisplayString)
```
kylehickinson: Lets made the fallback the `origin` instead of the full URL, since the full URL is displayed below regardless

Sources/Brave/Frontend/Settings/FilterListsView.swift
```diff
@@ -11,31 +11,143 @@ import BraveUI
...
                 .font(.caption)
                 .foregroundColor(Color(.secondaryBraveLabel))
             }
           }.toggleStyle(SwitchToggleStyle(tint: .accentColor))
             .listRowBackground(Color(.secondaryBraveGroupedBackground))
         }
       } header: {
-        Text(Strings.filterListsDescription)
-          .textCase(.none)
+        VStack(alignment: .leading, spacing: 8) {
```
kylehickinson: Spacing here looks weird, maybe reduce to 4 and add a small bottom padding

Sources/Brave/Frontend/Settings/FilterListsView.swift
```diff
@@ -11,31 +11,143 @@ import BraveUI
...
+                  "Failed to remove file for resource \(removedURL.setting.uuid)"
+                )
+              }
+              
+              // 4. Remove the setting.
+              // This should always happen in the end
+              // because we need to access properties on the setting until then
+              removedURL.setting.delete(inMemory: !customFilterListStorage.persistChanges)
+            }
+          }
```
kylehickinson: Move this to a separate function since its this large

Sources/Brave/Frontend/Settings/FilterListsView.swift
```diff
@@ -11,31 +11,143 @@ import BraveUI
...
+        } label: {
+          Text(Strings.addCustomFilterList)
+            .foregroundColor(Color(.braveBlurple))
+        }
+          .buttonStyle(PlainButtonStyle())
+          .disabled(editMode?.wrappedValue.isEditing == true)
+          .listRowBackground(Color(.secondaryBraveGroupedBackground))
+          .popover(isPresented: $showingAddSheet, content: {
+            FilterListAddURLView()
+          })
```
kylehickinson, line 78: This is broken on iPad:

image

Sources/Brave/WebFilters/AdblockResourceDownloader.swift
```diff
@@ -70,37 +83,70 @@ public actor AdblockResourceDownloader: Sendable {
...
+      if let downloadResult = try ResourceDownloaderStream.downloadResult(for: resource) {
+        await handle(downloadResult: downloadResult, for: resource)
+      } else {
+        switch resource {
+        case .genericContentBlockingBehaviors:
+          do {
+            // This is a special case which we load a bundled file if the downloaded file is not present
+            try await ContentBlockerManager.shared.compileBundledRuleList(for: .blockAds)
+          } catch {
+            assertionFailure("A bundled file should not fail to compile")
```
kylehickinson: Cached data not loading should not crash the app even in debug mode

Sources/Brave/WebFilters/AdblockResourceDownloader.swift
```diff
@@ -70,37 +83,70 @@ public actor AdblockResourceDownloader: Sendable {
...
+            ContentBlockerManager.log.error("Failed to compile bundled content blocker resource: \(error.localizedDescription)")
+          }
+          
+        default:
+          // No special handling if these files are not downloaded
+          // But ensure that this is only triggered for handled resource
+          // We should have never triggered this method for resources that are outside
+          // of the handled resources list
+          assert(Self.handledResources.contains(resource))
```
kylehickinson: Same as above

Sources/Brave/WebFilters/AdblockResourceDownloader.swift
```diff
@@ -70,37 +83,70 @@ public actor AdblockResourceDownloader: Sendable {
...
+        }
+      }
+    } catch {
+      assertionFailure()
```
kylehickinson: Same as above

Sources/Brave/WebFilters/BraveS3Resource.swift
```diff
@@ -0,0 +1,77 @@
+// Copyright 2023 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import Shared
+
+enum BraveS3Resource: Hashable, DownloadResourceInterface {
```
kylehickinson: Couldn't these changes be separated off into a different PR?

Sources/Brave/WebFilters/ContentBlocker/ContentBlockerManager.swift
```diff
@@ -10,422 +10,312 @@ import Shared
...
-  public enum BlocklistSourceType: Hashable, Sendable {
-    /// A block list that is bundled with this application
-    case bundled
-    /// A block list that is downloaded from a server
-    case downloaded(version: String?)
+  enum CompileError: Error {
+    case noRuleListReturned(identifier: String)
+    case couldNotPunycodeDomains(identifier: String)
+    case couldNotStripCosmeticFilters(identifier: String)
   }
```
kylehickinson: Change this into a struct that takes an identifier, create static definitions for each case like you have now

Sources/Brave/WebFilters/ContentBlocker/ContentBlockerManager.swift
```diff
@@ -10,422 +10,312 @@ import Shared
...
-  /// An object representing a rule type and a source type.
-  public struct RuleTypeWithSourceType: Hashable {
-    let ruleType: BlocklistRuleType
-    let sourceType: BlocklistSourceType
+  struct CompileOptions: OptionSet {
+    let rawValue: Int
+    
+    static let stripContentBlockers = CompileOptions(rawValue: 1 << 0)
+    static let punycodeDomains = CompileOptions(rawValue: 1 << 1)
+    static let `default`: CompileOptions = []
```
kylehickinson: Not needed, revisit if default ever has any options

Sources/Brave/WebFilters/ContentBlocker/ContentBlockerManager.swift
```diff
@@ -10,422 +10,312 @@ import Shared
...
   
-  /// An object representing the type of block list
-  public enum GeneralBlocklistTypes: String, CaseIterable {
-    case blockAds = "block-ads"
-    case blockCookies = "block-cookies"
-    case blockTrackers = "block-trackers"
+  public enum GenericBlocklistType: Hashable, CaseIterable {
+    case blockAds
+    case blockCookies
+    case blockTrackers
```
kylehickinson: Same as above, switch for a struct, bundled file name as var, static definitions for each

Sources/Brave/WebFilters/ContentBlocker/ContentBlockerManager.swift
```diff
@@ -10,422 +10,312 @@ import Shared
...
-    return await withTaskGroup(of: Void.self) { group in
-      for identifier in identifiers {
-        group.addTask {
-          do {
-            try await self.ruleStore.removeContentRuleList(forIdentifier: identifier)
-          } catch {
-            Self.log.error("\(error.localizedDescription)")
-          }
-        }
+    for identifier in availableIdentifiers ?? [] {
```
kylehickinson: Move the `?? []` to the assignment

Sources/Brave/WebFilters/ContentBlocker/ContentBlockerManager.swift
```diff
@@ -10,422 +10,312 @@ import Shared
...
+      do {
+        let ruleList = try await ruleStore.contentRuleList(forIdentifier: identifier)
+        cachedRuleLists[identifier] = ruleList
+      } catch {
+        // assertionFailure()
```
kylehickinson: Log or delete the line

Sources/Brave/WebFilters/ContentBlocker/ContentBlockerManager.swift
```diff
@@ -10,422 +10,312 @@ import Shared
...
+  /// Should be used only as a cleanup once during launch to get rid of unecessary/old data.
+  /// This is mostly for custom filter lists a user may have added.
+  public func cleaupInvalidRuleLists(validTypes: Set<BlocklistType>) async {
+    await cachedRuleLists.asyncConcurrentForEach { identifier, ruleList in
+      guard !validTypes.contains(where: { $0.identifier == identifier }) else { return }
+      
+      do {
+        try await self.removeRuleList(forIdentifier: identifier)
+      } catch {
+        assertionFailure()
```
kylehickinson: Do we need to crash in debug when cleanup fails?

Sources/Brave/WebFilters/ContentBlocker/ContentBlockerManager.swift
```diff
@@ -10,422 +10,312 @@ import Shared
...
       }
     }
   }
   
-  /// Start the timer that will continually compile new resources
-  public func startTimer() {
-    guard endlessCompileTask == nil else { return }
+  /// Compile the given resource and store it in cache for the given blocklist type
+  func compile(encodedContentRuleList: String, for type: BlocklistType, options: CompileOptions = .default) async throws {
```
kylehickinson: Just make options default `[]`

Sources/Brave/WebFilters/ContentBlocker/ContentBlockerManager.swift
```diff
@@ -10,422 +10,312 @@ import Shared
...
-          } catch {
-            self.cachedCompileResults[identifier] = (resource.sourceType, .failure(error))
-          }
-        }
-      }
-    }
+  /// Remove the rule list for the blocklist type
+  func removeRuleList(for type: BlocklistType) async throws {
+    try await removeRuleList(forIdentifier: type.identifier)
   }
```
kylehickinson: Given this method is only ever used in tests its fair to say it can be removed

Sources/Brave/WebFilters/ContentBlocker/ContentBlockerManager.swift
```diff
@@ -10,422 +10,312 @@ import Shared
...
+  /// Checks if the rule list for this type is cached
+  func hasCache(for type: BlocklistType) -> Bool {
+    return cachedRuleLists[type.identifier] != nil
+  }
+  
+  /// Load a rule list from the rule store and return it. Will use cached results if they exist
+  func ruleList(for type: BlocklistType) async throws -> WKContentRuleList? {
+    if let ruleList = cachedRuleLists[type.identifier] { return ruleList }
+    let ruleList = try await ruleStore.contentRuleList(forIdentifier: type.identifier)
+    self.cachedRuleLists[type.identifier] = ruleList
```
kylehickinson: This will blow away the cached rule list if for some reason it disappears from WebKit (`ruleList` being nil), is that fine?

Sources/Brave/WebFilters/ContentBlocker/ContentBlockerManager.swift
```diff
@@ -10,422 +10,312 @@ import Shared
...
+    return ruleList
+  }
+  
+  /// Load a rule list from the rule store and return it. Will not use cached results
+  func loadRuleList(for type: BlocklistType) async throws -> WKContentRuleList? {
+    let ruleList = try await ruleStore.contentRuleList(forIdentifier: type.identifier)
+    self.cachedRuleLists[type.identifier] = ruleList
```
kylehickinson: Same as above

Sources/Brave/WebFilters/ContentBlocker/ContentBlockerManager.swift
```diff
@@ -10,422 +10,312 @@ import Shared
...
   
-  /// Get the cached rule list for this rule type. Returns nil if it's either not compiled or there were errors during compilation
-  private func cachedRuleList(for ruleType: BlocklistRuleType) -> WKContentRuleList? {
-    guard let compileResult = cachedCompileResults[ruleType.identifier] else { return nil }
+  /// Convert all domain in the `if-domain` and `unless-domain` fields.
+  ///
+  /// Sometimes we get non-punycoded domans in our JSON and apple does not allow non-punycoded domains to be passed to the rule store.
+  private func punycodeConversion(encodedContentRuleList: String) async throws -> String? {
+    guard let blocklistData = encodedContentRuleList.data(using: .utf8) else {
+      assertionFailure()
```
kylehickinson: Is this necessary?

Sources/Brave/WebFilters/CustomFilterListStorage.swift
```diff
@@ -0,0 +1,55 @@
+// Copyright 2023 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import Data
+
+class CustomFilterListStorage: ObservableObject {
```
kylehickinson: May as well make the whole thing `MainActor` isolated since this class is essentially just a wrapper of `CustomFilterListSetting` which looks like its also is 99% main actor isolated

Sources/Brave/WebFilters/FilterListCustomURLDownloader.swift
```diff
@@ -0,0 +1,188 @@
...
+  @MainActor func handleUpdate(to filterListCustomURL: FilterListCustomURL, isEnabled: Bool) async {
+    if isEnabled {
+      let resource = filterListCustomURL.setting.resource
+      
+      do {
+        if let downloadResult = try ResourceDownloaderStream.downloadResult(for: resource) {
+          await self.handle(downloadResult: downloadResult, for: filterListCustomURL)
+        }
+      } catch {
+        assertionFailure()
```
kylehickinson: Not necessary, switch with a log or runtime error

Sources/Brave/WebFilters/FilterListCustomURLDownloader.swift
```diff
@@ -0,0 +1,188 @@
...
+  }
+  
+  /// Handle the update to a filter list enabled status which will will add or remove it from the engine.
+  /// This will not compile anything to iOS rule lists as we always do this on the download regardless of its enabled status.
+  @MainActor func handleUpdate(to filterListCustomURL: FilterListCustomURL, isEnabled: Bool) async {
+    if isEnabled {
+      let resource = filterListCustomURL.setting.resource
+      
+      do {
+        if let downloadResult = try ResourceDownloaderStream.downloadResult(for: resource) {
```
kylehickinson: This naming is really confusing, it seems like you're actively downloading a file or something, instead of getting a result

Sources/Brave/WebFilters/FilterListResourceDownloader.swift
```diff
@@ -168,41 +144,31 @@ public class FilterListResourceDownloader: ObservableObject {
...
-            for resource in settingInfo.resources {
-              group.addTask {
-                guard let fileURL = ResourceDownloader.downloadedFileURL(for: resource) else { return }
-                let date = try? ResourceDownloader.creationDate(for: resource)
-                await self.handle(downloadedFileURL: fileURL, for: resource, filterListUUID: settingInfo.uuid, date: date, index: settingInfo.index)
-              }
-            }
-          }
+    await filterListSettings.asyncConcurrentForEach { setting in
+      guard setting.isEnabled == true else { return }
```
kylehickinson: redundant `== true`, also for these types of checks its better to just use if instead of guard: e.g. `if !setting.isEnabled { return }`

Sources/Brave/WebFilters/FilterListResourceDownloader.swift
```diff
@@ -168,41 +144,31 @@ public class FilterListResourceDownloader: ObservableObject {
...
+        guard let componentId = setting.componentId else { return }
+        let resource = setting.makeResource(componentId: componentId)
+        
+        guard let downloadResult = try ResourceDownloaderStream.downloadResult(for: resource) else {
+          return
+        }
+        
+        await self.handle(downloadResult: downloadResult, for: setting)
+      } catch {
+        assertionFailure()
```
kylehickinson: Not necessary

Sources/Brave/WebFilters/FilterListResourceDownloader.swift
```diff
@@ -421,53 +388,53 @@ public class FilterListResourceDownloader: ObservableObject {
...
+  private func handle(downloadResult: ResourceDownloaderStream<BraveS3Resource>.DownloadResult, for filterList: FilterListInterface) async {
+    do {
+      if !downloadResult.isModified {
+        // if the file is not modified first we need to see if we already have a cached value loaded
+        guard try await ContentBlockerManager.shared.loadRuleList(
+          for: .filterList(uuid: filterList.uuid)
+        ) == nil else {
+          // We don't want to recompile something that we alrady have loaded
+          return
+        }
```
kylehickinson: `if let _ = ... { return }` here would read much better

Sources/Brave/WebFilters/ResourceDownloader.swift
```diff
@@ -265,14 +175,15 @@ actor ResourceDownloader: Sendable {
     }
   }
   
-  /// Get an existing etag for the given `Resource`
+  /// Get a creation date for the file downloaded for the given `Resource`.
+  /// If the file is not downloaded a nil is returned.
   static func creationDate(for resource: Resource) throws -> Date? {
     guard let fileURL = downloadedFileURL(for: resource) else { return nil }
```
kylehickinson: Why is this method static? Why wouldn't it be a method on `Resource`?

Sources/Brave/WebFilters/ResourceDownloaderStream.swift
```diff
@@ -50,21 +50,34 @@ struct ResourceDownloaderStream: Sendable, AsyncSequence, AsyncIteratorProtocol
...
+    case .notModified(let url, let date):
+      return DownloadResult(date: date, fileURL: url, isModified: false)
+    }
+  }
+  
   func makeAsyncIterator() -> ResourceDownloaderStream {
     return self
   }
+  
+  static func downloadResult(for resource: Resource) throws -> DownloadResult? {
```
kylehickinson: As talked about before, this name is super confusing. `DownloadedResult` would be better as a type, `result` would be better for the name of the method

Sources/Brave/WebFilters/ResourceDownloaderStream.swift
```diff
@@ -50,21 +50,34 @@ struct ResourceDownloaderStream: Sendable, AsyncSequence, AsyncIteratorProtocol
...
+      } catch let error as URLError {
+        // Soft fail these errors
         return .failure(error)
+      } catch {
+        throw error
       }
     }
   }
   
+  private func downloadResult() async throws -> DownloadResult {
```
kylehickinson: Same as below

Sources/Data/models/CustomFilterListSetting.swift
```diff
@@ -0,0 +1,75 @@
...
+
+import Foundation
+import CoreData
+import Shared
+import os.log
+
+public final class CustomFilterListSetting: NSManagedObject, CRUD, Identifiable {
+  @MainActor @NSManaged public var uuid: String
+  @MainActor @NSManaged public var isEnabled: Bool
+  @MainActor @NSManaged public var externalURL: URL
```
kylehickinson: If you're going to add `@MainActor` to every function & property, just add it to the class itself instead

Sources/Data/models/Model.xcdatamodeld/Model17.xcdatamodel/contents
```diff
@@ -60,6 +60,11 @@
             </uniquenessConstraint>
         </uniquenessConstraints>
     </entity>
+    <entity name="CustomFilterListSetting" representedClassName="Data.CustomFilterListSetting" syncable="YES">
+        <attribute name="externalURL" optional="YES" attributeType="URI"/>
+        <attribute name="isEnabled" optional="YES" attributeType="Boolean" usesScalarValueType="YES"/>
+        <attribute name="uuid" optional="YES" attributeType="String"/>
+    </entity>
```
kylehickinson: Do we want a new model version for this?

kylehickinson (changes requested):

App/iOS/Delegates/AppDelegate.swift
cuba: Nice catch!

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
cuba: It doesn't await the download. It awaits the actor. Like desktop the download is done on the background. Otherwise I would have added some download error handling

Sources/Brave/Frontend/Settings/FilterListsView.swift
cuba: Just so I'm sure I get you. The two strings are concatenated. If I simply remove the colon it will look good in English but in other languages there is no guarantee that it will. I guess I should change the translation to contain the %@ placeholder for the last updated time?

Sources/Brave/Frontend/Settings/FilterListsView.swift
cuba: Strange I tested this in iPad since iPad was the reason for the change. I'll take another look.

Sources/Brave/WebFilters/AdblockResourceDownloader.swift
cuba: It's not cached. It's bundled. Not sure how a bundled file should fail to load unless it's missing and needs to fixed right away.

Sources/Brave/WebFilters/BraveS3Resource.swift
cuba: Maybe but then I would have to add custom filter lists here and make changes to the resource file anyways because I don't want to pass the security key to non s3 resources. I felt this was was more secure in the long term when we have a clear separation of what is going to s3 and what is not.

Sources/Brave/WebFilters/ContentBlocker/ContentBlockerManager.swift
cuba: Can I ask why? An enum gives me a definite switch. Whereas a struct forces me to place a default. When I add new cases it's risky with a default that I forget to handle it in some place.

Sources/Brave/Frontend/Browser/LinkPreviewViewController.swift
cuba: Yes sorry must have been something I changed for something I needed but then changed the approach or something. Not sure why I touched this.

Sources/Brave/WebFilters/ResourceDownloader.swift
cuba: I could move it but now I'll be cleaning up things I did way back when. Can I make a small pr moving these methods after this?

Sources/Brave/WebFilters/ResourceDownloaderStream.swift
cuba: If you don't mind id like to do this in the proposed cleanup above. Because changing this will result in several places in code changing. And technically its a cached result so what about `cachedResult`?

Sources/Data/models/Model.xcdatamodeld/Model17.xcdatamodel/contents
cuba: I forgot what the rule is here. I'm adding a new entry so I thought its not necessary but I can add one if its best

App/iOS/Delegates/AppDelegate.swift
cuba: Fixed

App/iOS/Delegates/SceneDelegate.swift
cuba: Fixed

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
cuba: Fixed

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
cuba: Sorry yah ill add some error handling. Just need to get copy from design

Sources/Brave/Frontend/Settings/FilterListsView.swift
cuba: If you want me to use %@ placeholders, do you know of an example where we do this? `BraveStrings` don't use any of that

Sources/Brave/Frontend/Settings/FilterListsView.swift
cuba: Hmm, I put a bottom padding of 8. Now it looks strange to me. I thought we should be consistent here and leave the default bottom padding we have everywhere else. 4 looks to be the default.

Sources/Brave/WebFilters/AdblockResourceDownloader.swift
cuba: Well this is an assertion that if triggered needs to be fixed right away. It means I added a resource to be downloaded but never actually handled the download.

Sources/Brave/WebFilters/FilterListCustomURLDownloader.swift
cuba: Yes good point. I think that ill call it `cachedResult`

Sources/Brave/WebFilters/ResourceDownloaderStream.swift
cuba: I did change the method name to `cachedResult`

cuba:
@kylehickinson I fixed most of the comments. Some things I wasn't really sure what to do and so I put comments below your comments

Sources/Brave/WebFilters/AdblockResourceDownloader.swift
kylehickinson: The function name is literally `loadCachedData`…

Sources/Brave/WebFilters/ContentBlocker/ContentBlockerManager.swift
kylehickinson: You don't switch on this enum at all, and that is big indication that you don't actually need it to be an enum.

Sources/Brave/WebFilters/ResourceDownloader.swift
kylehickinson: At this point this PR is like 75% cleanup 25% add custom filter lists, so may as well do it here

Sources/Brave/WebFilters/ResourceDownloaderStream.swift
kylehickinson: I don't see this changed

kylehickinson (changes requested):
A vast majority of my requested changes are still not fixed

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
```diff
@@ -0,0 +1,111 @@
...
+                .foregroundColor(.red)
+            }
+          }
+        }, header: {
+          Text(Strings.customFilterListURL)
+        }, footer: {
+          VStack(alignment: .leading, spacing: 8) {
+            Text(Strings.addCustomFilterListDescription)
+            Text(LocalizedStringKey(Strings.addCustomFilterListWarning))
+          }.padding(.top, 16)
```
kylehickinson: 16 pts of padding is too much, lets reduce this to 8

Sources/Brave/Frontend/Settings/FilterListsView.swift
kylehickinson: There are 86 instances of `%@` used in BraveStrings

kylehickinson: If we plan on keeping the colon it still needs to be in the translated string for it to be localized correctly

Sources/Brave/WebFilters/ResourceDownloaderStream.swift
cuba: Sorry I changed only one of the methods. But just so you know this method actually downloads. I was thinking you were talking about the other method which returns cached data. This one actually does the download. So I made it `result()`

cuba: Ok moved

Sources/Brave/WebFilters/ResourceDownloader.swift
cuba: ok, moved

cuba: dbb03c7b4658a9cd278157d0425c0b988de7f9de

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
cuba: Ok no problem, but I'm just following the design I was given.

Sources/Data/models/Model.xcdatamodeld/Model17.xcdatamodel/contents
cuba: Done

Sources/Brave/WebFilters/ResourceDownloaderStream.swift
cuba: Done

Sources/Brave/WebFilters/ResourceDownloaderStream.swift
cuba: I didn't change DownloadResult yet. Not exactly sure what you want me to change `DownloadResult` to

cuba:
@kylehickinson Ok changed more stuff

Sources/Brave/Frontend/Settings/FilterListsView.swift
cuba: ahh if we have examples of %@ ill try that. It will look better I agree.

Sources/Brave/WebFilters/AdblockResourceDownloader.swift
cuba: Yes but it's doing this in the else (i.e. if there is no cache). So if you want I can call it `loadCachedOrBundledData`

Sources/Brave/Frontend/Settings/FilterListsView.swift
cuba: The thing is that I can't use `Text(style:)` with using `%@` :(

Sources/Brave/Frontend/Settings/FilterListsView.swift
cuba: I can't get this reproduced. Could this maybe be on iOS 14?

cuba: Ahh I got it. It happens when going through the kebab menu.

cuba: @kylehickinson so this should be solved. Anyway you can resolve this?

kylehickinson (changes requested):
- Most of the button for Add Custom Filter List is not tappable (If you are using `buttonStyle(.plain)` this should be removed) 
SCR-20230406-k2r

- Can we make the keyboard automatically focus the URL keyboard when the form appears (iOS 15+ so we can use `FocusState` now)
- Any chance for animations for errors/the custom list being added below the popup?
- Background color missing for form rows
SCR-20230406-k55
- After deleting the final filter list, edit mode should but disabled automatically
- Switch from `braveBlurple` to `braveBlurpleTint` on this button
SCR-20230406-k65

- Since we're iOS 15+ now we can also use `onSubmit` on the Form so that hitting done on the keyboard automatically submits the URL as if the user taps add

kjozwiak:
Removed from `1.49` as this wasn't merged into the `1.49` release. Please assign the appropriate milestone once the above is merged 👍

Sources/Data/models/Model.xcdatamodeld/Model19.xcdatamodel/contents
iccub, line 1: Brandon has model upgrades in his PR too, depending on which gets merged first you may need to update this model with Brandon's changes. Will let you know

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
```diff
@@ -9,10 +9,15 @@ import DesignSystem
 import BraveUI
 
 struct FilterListAddURLView: View {
+  enum FocusField: Hashable {
+    case urlInput
+  }
```
kylehickinson: No need to use an enum. Just use a bool

Sources/Brave/Frontend/Settings/FilterListsView.swift
```diff
@@ -8,13 +8,78 @@ import Strings
 import Data
 import DesignSystem
 import BraveUI
+import BraveCore
 
 /// A view showing enabled and disabled community filter lists
 struct FilterListsView: View {
   @ObservedObject private var filterListDownloader = FilterListResourceDownloader.shared
+  @ObservedObject private var customFilterListStorage = CustomFilterListStorage.shared
+  @Environment(\.editMode) private var editMode
```
kylehickinson, line 17: Should be a Binding (`@Environment(\.editMode) @Binding private var editMode`)

cuba: That doesn't compile. An `@Environment` can't be also `@Binding`, can it?

I'm just following the example on this page: https://developer.apple.com/documentation/swiftui/editmode

Sources/Brave/Frontend/Settings/FilterListAddURLView.swift
cuba: Fixed

cuba:
@kylehickinson I fixed what I can. The `@Environment` suggestion doesn't compile  and the cell animation is still off. Not sure how to handle these things.

iccub (approved):

kylehickinson (approved):

PR #6937, opened 2023-02-13
Fix #6829: Add SKUs "session expired" state

### LINKED ISSUE #6829: Implement session expired state
Look for `VPN - “session expired” state` spec for more details

iccub:
## Summary of Changes

This pull request fixes #6829 

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:

This is probably something for premium-dev testers, please talk to them.

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/BraveSkus/BraveSkusManager.swift
+++ b/Sources/Brave/BraveSkus/BraveSkusManager.swift
@@ -106,15 +106,18 @@ public class BraveSkusManager {
         jsonDecoder.keyDecodingStrategy = .convertFromSnakeCase
         let credentialSummaryJson = try jsonDecoder.decode(CredentialSummary.self, from: data)
         
-        if credentialSummaryJson.isValid {
-          
+        switch credentialSummaryJson.state {
+        case .valid:
           if Preferences.VPN.skusCredential.value == nil {
             Logger.module.debug("The credential does NOT exists, calling prepareCredentialsPresentation")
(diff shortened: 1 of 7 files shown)


iccub:
The code can be reviewed. Blocked label is there because the spec has not been formally approved yet

Sources/Brave/BraveSkus/BraveSkusManager.swift
```diff
@@ -155,4 +166,19 @@ public class BraveSkusManager {
       self.expiresAt = expiresAt
     }
   }
+  
+  // MARK: - Session Expired state
+  public static var keepShowingSessionExpiredState = true
```
soner-yuksel: Any reason for making the sessionExpiredProperty Static ?

Is there a way we can add SKUSManager as a dependency in where we used ? Same question applies to method after this also.

iccub: This is a simple flag to hold one value in-memory without persisting it. BraveSkusManager is a bit heavier class and has entire SKUService setup in the initializer so I would prefer to not pass it around if possible

Why is this flag needed:
If it's on, we show an alert to the user every time they tap on the vpn button. We do it until the app is closed and reopened.
btw this is a bug and the value should be false, left it to true just for testing

soner-yuksel (approved):

PR #7131, opened 2023-03-22
Fix #7147: Cleanup ContentBlockerManager and all related files

### LINKED ISSUE #7147: Cleanup ContentBlockerManager and improve its speed.
### Description: 
The content blocker manager uses a timed sync mechanism. Because apple caches rule lists, this is actually not needed. We don't need to recompile anything on re-launches, just when settings and data changes. 

This caching way makes it very difficult to work with custom filter lists and adds extra overhead on background threads for no reason.

cuba:
## Summary of Changes

This pull request fixes #7147

1. Now I compile more blocklists. To prevent from adding extra delays, I no longer compile cached blocklists during re-launch (it's not needed). This also meant I can remove much of the syncing logic as it is completely unnecessary.
2. The code is increasingly more difficult to maintain as I load more content. I struggled not breaking bundled content blockers several times (and almost broke it couple of times). Simplifying the logic helps to maintain it.
3. Punycode domains that are not puny coded. Because this adds extra work on compiling blocklists, it's important that we don't compile blocklists during launch. While things from s3 or bundled blocklists should not require puny coding there is no guarantee of this for custom blocklists. However I still puny code domains for things coming from s3 because as we've seen this has been an issue recently.
4. Removed multiple states on filter lists. Instead of having copies of the fields on filter lists entries from core data, I just use the entries themselves to display the data in the UI. Helps with syncing things especially when I will have different sources of content blockers (brave core and custom urls)
5. Rule lists are loaded lazily from the rule store (except for the bundled ones which need to be compiled at first launch). This improves launch time, especially when a lot of filter lists are enabled.
6. Added some better debugging logs because it's difficult to test if ad-blocking is working properly. something being blocked doesn't mean everything is working properly as we have multiple ways of blocking content: CF, engine blocking and rule lists. Depending on the rule lists a mistake in loading Adblock content could only reveal itself later when a rule is added. Logs tell me that everything is loaded (or unloaded) properly so when I make code changes (or a webcompat issue occurs) I know that everything is loaded properly. 

#### iPhone 14 (sim)
Adblock loaded: 0.175008s (0.194461s on development)
With all filter lists enabled: 0.870092s

#### iPad mini (6th gen) (sim)
Adblock loaded: 0.219583s (0.259748s on development)
With all filter lists enabled: 0.903977s

#### iPhone 14 Pro (device):
Adblock loaded: 0.281288s (0.291682s on development)
With all filter lists enabled: 1.036126s

**Note:** these values have a bit of variation. Sometimes they are higher and sometimes lower but the variation is usually not that large. Perhaps ~0.04

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:

We need to test 
1. ad-blocking by going to https://test-pages.privacytests.org/tracking_content.html. We want to test first launch and re-launches. This page should have `passed: true` for all values.
2. Launch time for both first launch and subsequent launches. Try to enable more filter lists. This will surely add more delays to launch time because more engine dat files need to be loaded. This is unavoidable. But the 
3. Cosmetic filtering. Best way to do this is by enabling blocking in app notices and going to pages like reddit.com
4. Script loading. Best way to do this is by going to https://dev-pages.brave.software/filtering/scriptlets.html. All values should be true with shields on and nil with shields off. Note the values will be nil on first launch.

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Frontend/Browser/Helpers/LaunchHelper.swift
+++ b/Sources/Brave/Frontend/Browser/Helpers/LaunchHelper.swift
@@ -27,6 +27,10 @@ public actor LaunchHelper {
     
     // Otherwise prepare the services and await the task
     let task = Task {
+      #if DEBUG
+      let startTime = CFAbsoluteTimeGetCurrent()
+      #endif
+      
       // Load cached data
       // This is done first because compileResources need their results
       async let filterListCache: Void = FilterListResourceDownloader.shared.loadCachedData()
@@ -36,6 +40,10 @@ public actor LaunchHelper {
       // Compile some engines
       await AdBlockEngineManager.shared.compileResources()
       
+      #if DEBUG
+      let timeElapsed = CFAbsoluteTimeGetCurrent() - startTime
+      ContentBlockerManager.log.debug("Adblock loaded: \(timeElapsed)s")
+      #endif
       // This one is non-blocking
       performPostLoadTasks(adBlockService: adBlockService)
       areAdBlockServicesReady = true
--- a/Sources/Brave/Frontend/Browser/Helpers/LaunchHelper.swift
+++ b/Sources/Brave/Frontend/Browser/Helpers/LaunchHelper.swift
@@ -5,10 +5,13 @@
 
 import Foundation
 import BraveCore
+import os
 
 /// This class helps to prepare the browser during launch by ensuring the state of managers, resources and downloaders before performing additional tasks.
 public actor LaunchHelper {
   public static let shared = LaunchHelper()
+  static let signpost = OSSignposter(logger: ContentBlockerManager.log)
+  
   private var loadTask: Task<(), Never>?
   private var areAdBlockServicesReady = false
   
@@ -27,26 +30,24 @@ public actor LaunchHelper {
     
     // Otherwise prepare the services and await the task
     let task = Task {
-      #if DEBUG
-      let startTime = CFAbsoluteTimeGetCurrent()
-      #endif
+      let signpostID = Self.signpost.makeSignpostID()
+      let state = Self.signpost.beginInterval("blockingLaunchTask", id: signpostID)
       
       // Load cached data
       // This is done first because compileResources need their results
       async let filterListCache: Void = FilterListResourceDownloader.shared.loadCachedData()
       async let adblockResourceCache: Void = AdblockResourceDownloader.shared.loadCachedAndBundledDataIfNeeded()
       _ = await (filterListCache, adblockResourceCache)
+      Self.signpost.emitEvent("loadedCachedData", id: signpostID, "Loaded cached data")
       
       // Compile some engines
       await AdBlockEngineManager.shared.compileResources()
+      Self.signpost.emitEvent("compileResources", id: signpostID, "Compiled engine resources")
       
-      #if DEBUG
-      let timeElapsed = CFAbsoluteTimeGetCurrent() - startTime
-      ContentBlockerManager.log.debug("Adblock loaded: \(timeElapsed)s")
-      #endif
       // This one is non-blocking
       performPostLoadTasks(adBlockService: adBlockService)
       areAdBlockServicesReady = true
+      Self.signpost.endInterval("blockingLaunchTask", state)
     }
     
     // Await the task and wait for the results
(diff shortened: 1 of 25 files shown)


soner-yuksel (changes requested):
Please before we can review this, can you clarify in details what - why is changed? I believe we need the motivation behind it to understand what kind of benefit we aim from this change.

Additionally can you please provide  some test measurements related with App Launch, Is it improved is it changed, what is the rundown from various devices/simulators and iOS versions?

And please add proper Labels and an issue for this change if it is required.

cuba:
> Please before we can review this, can you clarify in details what - why is changed? I believe we need the motivation behind it to understand what kind of benefit we aim from this change.
> 
> Additionally can you please provide some test measurements related with App Launch, Is it improved is it changed, what is the rundown from various devices/simulators and iOS versions?
> 
> And please add proper Labels and an issue for this change if it is required.

Sure, added to the description

Brandon-T:
Fresh install crash: 
image

Happens 100% of the time.

cuba:
> Fresh install crash: image
> 
> Happens 100% of the time.

Oops that assertion should not be wrapping the loadRuleList part of the code. That can certainly fail if the rule is not yet compiled. Which is the case on new installs. Fixed it.

Brandon-T:
Crash is fixed, but it still hangs for me on iPad. Does not hang for me on iPhone though.
But if it hangs like this, it might hang for others as well.
We should probably investigate that. Might be due to File I/O. But not sure what's causing it. Debugger doesn't show me why as it also stalls.

soner-yuksel:
As far as I tested there is a big difference between this version and development. I have over 100 tabs and

This branch is giving me a big delay both on app launch and while loading website. Please check the video you can clearly see where the Hangs are happening since it is shown by the OS. Right now it is only measuring Hangs which is bigger than 500 ms


https://user-images.githubusercontent.com/6643505/228364457-dc51fa81-c038-409f-b4bc-77d4967a7715.MOV


The next video is actual development with same configuration and number of tabs.


https://user-images.githubusercontent.com/6643505/228364716-52223974-c37d-44f5-a68a-f8344c53d84a.mov

cuba:
@Brandon-T @soner-yuksel Best I can think of is the `await` on the delegate. It has to load x number of rule lists depending on how many filter lists you have enabled which takes some time. but still, it's not like it didn't do it before. unless it was broken before and it never guaranteed rule lists were loaded on time...I wonder if this is happening for you on a new install or an upgrade or both?

cuba:
And you're not running any custom version of brave core, are you?

soner-yuksel:
> And you're not running any custom version of brave core, are you?

No brave-core local build or anything, I am installing on top the same app, the same bundle we use in AppStore over and over again (this version-that version)so I can check proper settings and number of tabs etc. and make proper comparison, better not to do empty not tabs app tests, they are misleading for us

Brandon-T (approved):

Sources/Brave/Frontend/Browser/Helpers/LaunchHelper.swift
```diff
@@ -27,21 +27,26 @@ public actor LaunchHelper {
     
     // Otherwise prepare the services and await the task
     let task = Task {
+      #if DEBUG
+      let startTime = CFAbsoluteTimeGetCurrent()
+      #endif
```
kylehickinson: Please make a note to replace all of these debug timers with OSSignposter (which would not be behind a `DEBUG` check)

Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Paged/ContentBlockerScriptHandler.swift
```diff
@@ -66,7 +66,7 @@ extension ContentBlockerHelper: TabContentScript {
       Task { @MainActor in
         let isPrivateBrowsing = PrivateBrowsingManager.shared.isPrivateBrowsing
         let domain = Domain.getOrCreate(forUrl: currentTabURL, persistent: !isPrivateBrowsing)
-        if let shieldsAllOff = domain.shield_allOff, Bool(truncating: shieldsAllOff) {
+        guard !domain.areAllShieldsOff else {
```
kylehickinson: Typically easier to read simple boolean ops using if instead of guard

```swift
if domain.areAllShieldsOff {
  return
}
```

Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Sandboxed/SiteStateListenerScriptHandler.swift
```diff
@@ -65,7 +65,7 @@ class SiteStateListenerScriptHandler: TabContentScript {
         Task { @MainActor in
           let domain = pageData.domain(persistent: !tab.isPrivate)
           
-          guard domain.shield_allOff?.boolValue != true else {
+          guard !domain.areAllShieldsOff else {
```
kylehickinson: Same as above

soner-yuksel (approved):

Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Paged/ContentBlockerScriptHandler.swift
cuba: Ok I can change this. 

Though I kind of like guards because you know it will return if the condition is not met and you know that it doesn't continue execution. Whereas with an `if` you have to look at the content of the `if`. 
Though I agree that in a simple `if` is pretty easy to see the return and you don't need to negate the condition like I do here.

Sources/Brave/Frontend/Browser/Helpers/LaunchHelper.swift
cuba: I have changed this. I hope this is what you expected

kylehickinson (approved):

PR #7148, opened 2023-03-25
Fix #6254: Get cosmetic filters on serial queue

### LINKED ISSUE #6254: Adblock/Cosmetic filters crash
### Steps to Reproduce 
No way to reproduce found so far.

**Actual result:** 
Crash

**Expected result:**

**Brave Version:** 

**Device details:** 

**Website problems only:**

cuba:
## Summary of Changes

This pull request fixes #6254

This PR ensures that the cosmetic filters are retrieved on the serial queue. This should prevent any possible crashes due to threading. However the engine does allow multiple threads to be used so its more of a extra security than actual fix.

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:

1. Test cosmetic filters. For example. reddit.com should hide cookie consent notices when cookie consent blocking is enabled
2. Test launch performance (both initial launch and relaunch). Enable all filter lists for extra delays during launch. 
3. Test navigation performance. Just make sure it's not caused by a slow internet connection. best to compare before and after changes and see if there is any visible difference in website loading times that aren't present prior to the changes. Again enabling a lot of filter lists before and after change is a good test.
4. Ensure that passed: true is enabled for all values on the following site: https://test-pages.privacytests.org/tracking_content.html
5. Ensure that true is set on all frames on the following site: https://dev-pages.brave.software/filtering/scriptlets.html (note that this may be false on initial launch)

### Performance tests
0.918944s with all filter lists enabled. This is similar to what is found in current dev.

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
+++ b/Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
@@ -279,7 +279,7 @@ extension BrowserViewController: WKNavigationDelegate {
       // Check if custom user scripts must be added to or removed from the web view.
       if let targetFrame = navigationAction.targetFrame {
         tab?.currentPageData?.addSubframeURL(forRequestURL: url, isForMainFrame: targetFrame.isMainFrame)
-        let scriptTypes = tab?.currentPageData?.makeUserScriptTypes(domain: domainForMainFrame) ?? []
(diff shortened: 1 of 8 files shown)


Brandon-T (approved):

iccub (changes requested):
`AdBlockEngineManagerTests` fail now.
I tried to rebase it and merge, i hope i did not mess this up

let's fix then i will merge it

iccub (approved):

PR #7171, opened 2023-03-30
Fix #5481: Limit account name to 30 characters

### LINKED ISSUE #5481: Limit Account name to 30 characters

StephenHeaps:
## Summary of Changes
- Add 30 character limit to account name
- Updated name field to drop to multiline if needed on iOS 16+; minor UX tweak (nice with accessibility text sizes).

This pull request fixes #5481

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:
1. Open wallet, switch to Accounts tab and tap the `+` button
2. Select a coin type, then try entering a name > 30 characters long. 
3. Verify error message is shown and `add` button is disabled
4. Navigate back to Accounts tab and select an account
5. Tap `Rename`
6. Try entering an account name > 30 characters long
7. Verify error message is shown and `done` button is disabled

## Screenshots:

Add Account | Add Account multiline (iOS 16+) | Rename Account | Rename Account multiline (iOS 16+)
--|--|--|--
5481 - add account | 5481 - add account multiline | 5481 - rename account | 5481 - rename account multiline

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/BraveWallet/Crypto/Accounts/Details/AccountDetailsView.swift
+++ b/Sources/BraveWallet/Crypto/Accounts/Details/AccountDetailsView.swift
@@ -24,10 +24,13 @@ struct AccountDetailsView: View {
   @State private var isPresentingRemoveConfirmation: Bool = false
 
   @Environment(\.presentationMode) @Binding private var presentationMode
+  
+  private var isDoneDisabled: Bool {
+    name.isEmpty || !name.isValidAccountName
+  }
 
   private func renameAccountAndDismiss() {
-    if name.isEmpty {
-      // Show error?
+    guard !name.isEmpty && name.isValidAccountName else {
       return
     }
     keyringStore.renameAccount(account, name: name)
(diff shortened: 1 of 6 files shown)


Sources/BraveWallet/Crypto/Accounts/Details/AccountDetailsView.swift
```diff
@@ -24,10 +24,13 @@ struct AccountDetailsView: View {
   @State private var isPresentingRemoveConfirmation: Bool = false
 
   @Environment(\.presentationMode) @Binding private var presentationMode
+  
+  private var isDoneDisabled: Bool {
+    name.isEmpty || !name.isValidAccountName
```
soner-yuksel, line 29: Can we add isEmpty case inside isValidAccountName to satisfy this condition together? I believe in other place which uses isValidAccountName also does the isEmpty check.

soner-yuksel (approved):

Sources/BraveWallet/Crypto/Accounts/Details/AccountDetailsView.swift
StephenHeaps: An empty account name is permitted when adding an account because the default account name would be used. As a result, empty string is only invalid when renaming the account.

nuo-xu (approved):

PR #7184, opened 2023-03-31
Fix #7183: Use MutationObserver for JitsiSDK iframe's src attribute.

### LINKED ISSUE #7183: Use mutationobserver to fetch BraveTalk native iframe's src attribute

iccub:
## Summary of Changes

This pull request fixes #7183 

## Submitter Checklist:

  - [ ] Light & dark mode

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Brandon-T (commented):
Hmm maybe we could just override the `HTMLIFrameElement.setAttribute`

And check for "src" attribute being set. We do this in playlist atm. But not sure if we should for jitsi.

EDIT:
I guess this is fine since it only observes the frame and the attribute filter is also "src"

Brandon-T (approved):

iccub:
@Brandon-T any particular benefit in that approach over mutationobserver?

Brandon-T:
> @Brandon-T any particular benefit in that approach over mutationobserver? 

I was thinking about the performance of observers vs. a function that was made to handle attribute changes. But then I realized the observer actually is filtering it by that attribute anyway, so it should be similar.

So no difference for your approach.

For playlist since we observe the whole page it'd be very expensive. So instead we override src setter only for video and audio elements which is much better.

So TLDR: you're good. No worries. No need to change anything :)

iccub:
Closing this one for now

PR #7205, opened 2023-04-05
Fix #7034: fixing activation of tabs when moving them around

### LINKED ISSUE #7034: Multiple tabs can be shown as active
### Description: 

Dragging an inactive tab makes it active, but sometimes does not reset the previously active tab to show as inactive

### Steps to Reproduce 
  1. Have more than 1 tab open, have tab bar set to always visible
  2. Drag an inactive tab to the end of the tab bar
  3. Observe inactive tab now shows as active, but the old active tab does not update to show as inactive

### Actual result: 

https://user-images.githubusercontent.com/5314553/222179024-e5bee269-26cf-4823-813c-fdf34dbe505b.mov

### Reproduces how often: 

Not 100% of the time but consistent enough to reproduce

### Brave Version: 

### Website problems only:

### LINKED ISSUE #36102: [iOS] Multiple tabs can be shown as active
### Description: 

Dragging an inactive tab makes it active, but sometimes does not reset the previously active tab to show as inactive

### Steps to Reproduce 
  1. Have more than 1 tab open, have tab bar set to always visible
  2. Drag an inactive tab to the end of the tab bar
  3. Observe inactive tab now shows as active, but the old active tab does not update to show as inactive

### Actual result: 

https://user-images.githubusercontent.com/5314553/222179024-e5bee269-26cf-4823-813c-fdf34dbe505b.mov

### Reproduces how often: 

Not 100% of the time but consistent enough to reproduce

### Brave Version: 

### Website problems only:

davidcardd1:
## Summary of Changes

This pull request fixes brave/brave-browser#36102. I explicitly changed attributes of all unselected cells (like hiding the close button) inside the method that gets called when moving around a tab.

## Submitter Checklist:

- [X] New or updated UI has been tested across:
  - [X] Light & dark mode
  - [X] Different size classes (iPhone, landscape, iPad)
  - [X] Different dynamic type sizes

## Test Plan:

1. Open two or more tabs
2. Move one of the unselected tabs to another position
3. Move unselected tabs to all possible positions
4. Verify that only the moved tab gets selected. All the other tabs appear unselected (tab label font is not bold, close button is hidden, darker background color)

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Frontend/Browser/Tabs/TabBar/TabsBarViewController.swift
+++ b/Sources/Brave/Frontend/Browser/Tabs/TabBar/TabsBarViewController.swift
@@ -414,6 +414,17 @@ extension TabsBarViewController: UICollectionViewDataSource {
     updateData()
 
     guard let selectedTab = tabList[destinationIndexPath.row] else { return }
+      
+    // Deselects all tabBarCells before selecting the one that was moved
+    for i in 0...tabs.count - 1 where i != destinationIndexPath.row {
+    if let cell = collectionView.cellForItem(at: IndexPath(row: i, section: 0)) as? TabBarCell {
+      cell.closeButton.isHidden = true
+      cell.deselectedOverlayView.isHidden = false
+      cell.isSelected = false
+      cell.configure()
+    }
+    }
+      
     manager.selectTab(selectedTab)
   }
 }
--- a/Sources/Brave/Frontend/Browser/Tabs/TabBar/TabBarCell.swift
+++ b/Sources/Brave/Frontend/Browser/Tabs/TabBar/TabBarCell.swift
@@ -14,7 +14,7 @@ class TabBarCell: UICollectionViewCell {
     return label
   }()
 
-  private lazy var closeButton: UIButton = {
+  lazy var closeButton: UIButton = {
     let button = UIButton()
     button.addTarget(self, action: #selector(closeTab), for: .touchUpInside)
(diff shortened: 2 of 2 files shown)


soner-yuksel (changes requested):
@davidcardd1 thanks again for the PR. You are on the right track but there are some problems with the logic.

First of all it is not good idea to change the access modifiers and remove `private`, if there is a functionality needs to be implemented it should be carried over a function etc.

Also making all the tabs except the selected one is not feasible, many users have quite a big number of tabs and configuring every single tab when it is not necessary is not a great approach. We should find the cause of this particular problem and solve it for that particular cell.

Sources/Brave/Frontend/Browser/Tabs/TabBar/TabsBarViewController.swift
```diff
@@ -414,6 +414,17 @@ extension TabsBarViewController: UICollectionViewDataSource {
     updateData()
 
     guard let selectedTab = tabList[destinationIndexPath.row] else { return }
+      
+    // Deselects all tabBarCells before selecting the one that was moved
+    for i in 0...tabs.count - 1 where i != destinationIndexPath.row {
+    if let cell = collectionView.cellForItem(at: IndexPath(row: i, section: 0)) as? TabBarCell {
```
soner-yuksel, line 420: nit: if clause should have right Indentation.

soner-yuksel:
Hi @davidcardd1 closing the PR because of inactivity, please feel free to make another PR when you have another solution. Thanks for contribution

PR #7361, opened 2023-04-28
Fix #7123: showing network logo on asset icon

### LINKED ISSUE #7123: Show network logo over asset icon instead of coin logo
### Description: 

iOS version of https://github.com/brave/brave-browser/issues/29166

We should update to show the network logo in the corner of asset icons instead of the coin logo.

nuo-xu:
## Summary of Changes
1. showing network logo on non-native asset icon
2. added a border around network logo to match desktop 
3. add filecoin network logo for two filecoin (evm) networks

## Submitter Checklist:

  - [ ] Light & dark mode

## Test Plan:

Check Portfolio/Token search/NFT tab
all non-native assets should display with a network logo
or if the network is not the official Ethereum network, but uses ETH as gas (e.x Ether on Aurora and Ether on Optimism)

## Screenshots:

https://user-images.githubusercontent.com/1187676/235232844-9857a30c-9c18-4fde-8f4e-54e5b30758d8.mp4

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Sources/BraveWallet/Extensions/BraveWalletSwiftUIExtensions.swift
```diff
@@ -74,6 +78,37 @@ extension BraveWallet.NetworkInfo: Identifiable {
...
+    } else if chainId.caseInsensitiveCompare(BraveWallet.FantomMainnetChainId) == .orderedSame {
+      return "fantom"
+    } else if chainId.caseInsensitiveCompare(BraveWallet.AuroraMainnetChainId) == .orderedSame {
+      return "aurora"
+    } else if chainId.caseInsensitiveCompare(BraveWallet.OptimismMainnetChainId) == .orderedSame {
+      return "optimism"
+    } else {
+      return iconUrls.first
+    }
+  }
```
StephenHeaps: Right now these strings are duplicated in `BraveWallet.NetworkInfo.networkLogo`, `BraveWallet.NetworkInfo.nativeTokenLogo`, and in `NetworkIcon.networkImageInfo`. We should avoid having this much duplicated code, especially strings that won't get caught by the compiler. 

Maybe we could create an enum for each image string? That way we don't miss updating any strings as they're all hardcoded in one location. We might even be able to move these computed properties into the enum, or make them functions in the enum (or can we reduce so only one property is needed total?)?

ex.
```
enum AssetImageName: String {
  case solana = "sol-asset-icon"
  <...>
  
  func networkLogoImageName(for chainId: String) -> String? {
    switch chainId {
    case .solana: return AssetImageName.solana.rawValue
    default: return nil
    }
  }
}
```

Maybe there is another way to reduce the duplicated code if these 2 properties can't be merged into one?

Sources/BraveWallet/NetworkIcon.swift
```diff
@@ -79,6 +79,10 @@ struct NetworkIcon: View {
       return ("optimism", false)
     case BraveWallet.AuroraMainnetChainId:
       return ("aurora", false)
+    case BraveWallet.FilecoinEthereumMainnetChainId:
+      return ("filecoin-asset-icon", false)
+    case BraveWallet.FilecoinEthereumTestnetChainId:
+      return ("filecoin-asset-icon", true)
     default:
       return nil
```
StephenHeaps: Similar to above comment about duplicated code, I think we can make use of an enum to get these string values (perhaps with function accepting a `chainId`), and I _think_ we could use a simple check against `WalletConstants.supportedTestNetworkChainIds` to see if we should grayscale the image or not:
```let isGrayscale = WalletConstants.supportedTestNetworkChainIds.contains(network.chainId)```

StephenHeaps (changes requested):

Sources/BraveWallet/Extensions/BraveWalletSwiftUIExtensions.swift
nuo-xu: updated in [188d673](https://github.com/brave/brave-ios/pull/7361/commits/188d6732e73af0f5cf9271d747aef6b7aedb266b)

Sources/BraveWallet/NetworkIcon.swift
nuo-xu: updated in [188d673](https://github.com/brave/brave-ios/pull/7361/commits/188d6732e73af0f5cf9271d747aef6b7aedb266b)

Sources/BraveWallet/Extensions/BraveWalletSwiftUIExtensions.swift
StephenHeaps: Can we merge these properties? Can we reduce the duplicate code at all? SonarCloud is also flagging the duplicate code (I think it will flag it every time we touch this file).

Sources/BraveWallet/NetworkIcon.swift
StephenHeaps: Can we use that `isGrayscale` flag to reduce code here?

StephenHeaps (approved):

PR #7486, opened 2023-05-19
Fix #7484 - Fix migration crash when parsing an invalid URL

### LINKED ISSUE #7484: Migration crash when user has invalid history URLs
### Description: 
- Somehow users can have a URL in their history that's not parsable as a URL
- This causes the migration to WebKit to crash since we cannot parse their URLs and end up with `UInt(-1)` index, where a negative number cannot be represented as unsigned in Swift.

### Steps to Reproduce 
  1. Somehow get an invalid URL in your history

### Actual result: 
- Crash

### Expected result:
- No crash. It should skip over the bad URL

- Very hard to reproduce

### Brave Version: 
- 1.51 AppStore

### Device details: 
- Any

### Website problems only:
- Not sure

Brandon-T:
## Summary of Changes
- Fix potential issues in migration where users URLs can somehow not be parsed by both Chromium and Foundation.
- This results in their being 0 history to migrate, and gives use `UInt(-1)` which cannot overflow, thus giving a crash.

This pull request fixes #7484

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:
- None as I have no idea how to reproduce the issue.

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Migration/Migration.swift
+++ b/Sources/Brave/Migration/Migration.swift
@@ -122,16 +122,25 @@ public class Migration {
         let historySnapshot = oldTab.urlHistorySnapshot as? [String] ?? []
         
         for url in historySnapshot {
-          guard let url = NSURL(idnString: url) as? URL ?? URL(string: url) else { continue }
+          guard let url = NSURL(idnString: url) as? URL ?? URL(string: url) else {
+            Logger.module.error("Failed to parse URL: \(url) during Migration!")
+            continue
+          }
           if let internalUrl = InternalURL(url), !internalUrl.isAuthorized, let authorizedURL = InternalURL.authorize(url: url) {
             historyURLs.append(authorizedURL)
           } else {
             historyURLs.append(url)
           }
         }
+        
+        if historyURLs.count == 0 {
+          Logger.module.error("User has zero history to migrate!")
+          return
+        }
 
         // currentPage is -webView.backForwardList.forwardList.count
-        let currentPage = (historyURLs.count - 1) + Int(oldTab.urlHistoryCurrentIndex)
+        // If for some reason current page can be negative, we clamp it to [0, inf].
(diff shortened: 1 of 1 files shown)


Sources/Brave/Migration/Migration.swift
```diff
@@ -122,16 +122,25 @@ public class Migration {
         let historySnapshot = oldTab.urlHistorySnapshot as? [String] ?? []
         
         for url in historySnapshot {
-          guard let url = NSURL(idnString: url) as? URL ?? URL(string: url) else { continue }
+          guard let url = NSURL(idnString: url) as? URL ?? URL(string: url) else {
+            Logger.module.error("Failed to parse URL: \(url) during Migration!")
```
iccub, line 126: if this is never mean to happen we can consider adding assertion failure here

iccub (approved):

Sources/Brave/Migration/Migration.swift
soner-yuksel: No lets not add that please, I have no idea how the previous problem happened. It will be better not to have a random crash for an reason which is hard to explain.

soner-yuksel (approved):

Sources/Brave/Migration/Migration.swift
iccub: This would help us catch this problem though? assertion failure is not going to crash outside running it locally

soner-yuksel: Yes totally it will not, thats correct and also we should not be having `migrateTabStateToWebkitState` called after this version, I believe neither do QA. 

I am still thinking if the problem caused by a synced brave internal URL or something, very weird

bsclifton: Is there an ability to send crash report without crashing? we do this in brave-core (`DumpWithoutCrashing`)

PR #7489, opened 2023-05-19
Fix #7491: Change the recovery phrase protection overlay

### LINKED ISSUE #7491: [Security] - Internal #915
### Description: 
https://github.com/brave/internal/issues/915

stoletheminerals:
## Summary of Changes
Changes the default `.redacted(reason: .placeholder)` recovery phrase blur protection with a custom fixed-size overlay.

This pull request fixes https://github.com/brave/internal/issues/915 , https://github.com/brave/brave-ios/issues/7491

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:

Follow the steps in https://github.com/brave/internal/issues/915

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/BraveUI/SwiftUI/CustomPrivacySensitiveModifier.swift
+++ b/Sources/BraveUI/SwiftUI/CustomPrivacySensitiveModifier.swift
@@ -18,8 +18,10 @@ private struct CustomPrivacySensitiveModifier: ViewModifier {
       if isBackgrounded || isCaptured {
         content
           .opacity(0)
-          .overlay(alignment: .center) {RoundedRectangle(cornerRadius: 2)
-              .fill(Color.black.opacity(0.1)).frame(width: 64).padding(.vertical, 2.5)}
+          .overlay(alignment: .center) {
+            RoundedRectangle(cornerRadius: 2)
+              .fill(Color.black.opacity(0.1)).frame(width: 64).padding(.vertical, 2.5)
+            }
       } else {
         content
       }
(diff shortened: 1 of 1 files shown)


StephenHeaps (approved):
Should be fine using fixed size for this 👍 . Checked against iPhone SE 1st gen with display zoom enabled (smallest possible screen size) & largest text size (worst case scenario).
iPhone SE | iPhone SE max text size
--|--

Sources/BraveUI/SwiftUI/CustomPrivacySensitiveModifier.swift
```diff
@@ -17,7 +17,11 @@ private struct CustomPrivacySensitiveModifier: ViewModifier {
     Group {
       if isBackgrounded || isCaptured {
         content
-          .redacted(reason: .placeholder)
+          .opacity(0)
```
kdenhartog, line 20: Could we keep the .redacted here, but utilize the [privacy](https://developer.apple.com/documentation/swiftui/redactionreasons/privacy) reason instead? Seems like that would keep a standardized styling to address the issue.

The thing I'm not certain about is whether or not the use of .redacted versus .opacity might allow the content to be considered non sensitive and then still be accessible in some odd way.

stoletheminerals: `.redacted(reason: .privacy)` doesn't work here for some reason. A comment in this file is probably related to this:
```
/// Custom modifier to behave similar to `.privacySensitive()`
/// The `.privacySensitive()` doesn't work in sheets because the
/// `redactedReasons` doesn't get passed through the environment in sheets.
```
Anyway, under the hood it works the same way, we apply the rule when `isBackgrounded || isCaptured` condition is met, so with `opacity(0)` we make sure the content is never visible.

PS. The logic of _when_ we apply an overlay doesn't change, only the way of _how_ we apply it. Technically `opacity(0)` is enough here and the overlay is added just for better looks.

kdenhartog (approved):
This addresses the issue - approving

PR #7500, opened 2023-05-23
Fix #7476: QR Code Scanning erroneously hides Scrolling Tabs Bar

### LINKED ISSUE #7476: QR Code Scanning erroneously  hides Scrollin Tabs Bar
### Description: 

If Scrolling Tab Bar is enabled and QR code scanning is used from URL Bar, after canceling the QR Code Scan the Tabs Bar disappear for the current tab.

### Steps to Reproduce 
  1. Enable Tabs Bar
  2. Create new tab
  3. Activate the URL bar entry
  4. Press QR code button
  5. Close QR Code button with done
  6. Check scrolling tab bar is disappeared

### Actual result: 

Tabs Bar disappeared for that particular Tab

https://github.com/brave/brave-ios/assets/6643505/8308489b-da92-4804-9e19-14723435e10c

### Expected result:

Tab Bar should never disappear with usage of QR Code Scan

Easily Reproduced

### Brave Version: 

- Can you reproduce this issue with the most recent build from TestFlight? YES
- Can you reproduce this issue with the previous version of the current build from TestFlight? YES
- Can you reproduce this issue with the current build from AppStore? Y

### Device details: 
Any iPhone

### Website problems only:

stoletheminerals:
## Summary of Changes
When a QR button is pressed, we call `leaveOverlayMode` before popping up a QR Scanner:
https://github.com/brave/brave-ios/blob/c12d49888206c6aa0d848b23f37f846b067ebb28/Sources/Brave/Frontend/Browser/Toolbars/UrlBar/TopToolbarView.swift#L632-L635
`leaveOverlayMode` calls `shouldShowTabBar` to determine if it should show the tab bar:
https://github.com/brave/brave-ios/blob/c12d49888206c6aa0d848b23f37f846b067ebb28/Sources/Brave/Frontend/Browser/BrowserViewController.swift#L1436-L1454
At this point `keyboardState` is not `nil` so the function returns `false` and the tab bar remains hidden. Next `topToolbarDidPressQrCodeButton` will be called and `shouldShowTabBar` is not called again.
So the proposed fix is to call `topToolbarDidPressQrCodeButton` first which will collapse the keyboard and after `leaveOverlayMode` is called `keyboardState` will be `nil`, so `shouldShowTabBar` will return `true`

This pull request fixes #7476

## Submitter Checklist:

  - [ ] Light & dark mode

## Screenshots:

https://github.com/brave/brave-ios/assets/10810135/128868a1-8561-487e-9221-8282a3f5a029

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Frontend/Browser/Toolbars/UrlBar/TopToolbarView.swift
+++ b/Sources/Brave/Frontend/Browser/Toolbars/UrlBar/TopToolbarView.swift
@@ -630,8 +630,8 @@ class TopToolbarView: UIView, ToolbarProtocol {
   }
 
   @objc func topToolbarDidPressQrCodeButton() {
-    leaveOverlayMode(didCancel: true)
     delegate?.topToolbarDidPressQrCodeButton(self)
+    leaveOverlayMode(didCancel: true)
   }
   
   @objc private func swipedLocationView() {


soner-yuksel (approved):
Thanks for the simple solution @stoletheminerals, was about to add voice search with similar manner, this is perfect timing.

PR #7503, opened 2023-05-24
Fix #7502: Test additional privacy scripts

### LINKED ISSUE #7502: Add unit tests for cosmetic filtering and request blocking scripts
### Description: 
The following scripts need to be tested to make sure they are functioning properly:
- `SelectorsPollerScript.js`
- `RequestBlockingScript.js`

cuba:
## Summary of Changes

This pull request fixes #7502

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:

This is mostly adding some unit tests which don't need to be tested. There was however some minor changes to the initialization of the cosmetic filtering script. So while we test the scripts itself the initialization of this script is not. So we just need to check that the scripts are injected properly into websites.

Hence we need to:
- [ ] Navigate to several sites and ensure the CF script is hiding elements. No need to test the quality of the script. just that it's working.

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Shared/Extensions/WKWebViewExtensions.swift
+++ b/Sources/Shared/Extensions/WKWebViewExtensions.swift
@@ -74,6 +74,7 @@ public extension WKWebView {
   @discardableResult @MainActor func evaluateSafeJavaScript(
     functionName: String,
     args: [Any] = [],
+    frame: WKFrameInfo? = nil,
     contentWorld: WKContentWorld,
     escapeArgs: Bool = true,
     asFunction: Bool = true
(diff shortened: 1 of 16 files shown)


Sources/Shared/Extensions/WKWebViewExtensions.swift
```diff
@@ -82,11 +83,37 @@ public extension WKWebView {
...
         escapeArgs: escapeArgs,
         asFunction: asFunction) { value, error in
           continuation.resume(returning: (value, error))
-      }
+        }
+    }
+  }
+  
+  @discardableResult
+  @MainActor func evaluateSafeJavaScriptThrowing(
```
Brandon-T, line 96: Why not just make the original code throwing instead of a new function?
Also, prefer `let (result, error) = await evaluateSafeJavascript(....)` instead of doing `let error = result.1`

Brandon-T (changes requested):

Sources/Shared/Extensions/WKWebViewExtensions.swift
cuba: I tried this at first but required quite a few changes to existing places in code which ignore the error. 
The only easy way for me to do it was to put a try? or do an empty catch everywhere its used as I couldn't know how to deal with this error everywhere this function is used. I was afraid of doing this because surely adding a try? or an empty catch would raise some eyebrows (even though technically that's what is happening now). I thought this is a better approach because we can chose to deprecate the original method at any time. Plus there are quite a few places this is used so it would be quite a few changes in code for something that is supposed to be a testing PR.

And thank you for the suggestion for the `let (result, error)` I always forget this is possible.

Brandon-T (approved):

PR #7547, opened 2023-05-31
Fix #7548: Tab Deallocation on main thread

### LINKED ISSUE #7548: Crash - Sometimes when closing a tab, the browser crashes
### Description: 
- Sometimes, when a user closes a tab, the browser crashes.
- Crash log shows it's `Tab.deinit` in the Xcode organizer.

### Steps to Reproduce 
  1. Close some Tabs

### Actual result: 
- Crash

### Expected result:
- Shouldn't crash

- Very very hard to reproduce

### Brave Version: 
- 1.49.x

### Device details: 
- iPhone

Brandon-T:
## Summary of Changes
- Access everything from brave-core weakly. If tab deallocates, all of the objects deallocate under the hood. The weak references become nil. If they are still for some reason doing an operation, well it should be cancelled.

- Fix possible favicon crash

This pull request fixes #7548

## Submitter Checklist:

  - [x] Light & dark mode

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Frontend/Browser/Tab.swift
+++ b/Sources/Brave/Frontend/Browser/Tab.swift
@@ -73,8 +73,7 @@ class Tab: NSObject {
   var blockAllAlerts: Bool = false
   
   private(set) var type: TabType = .regular
-  private let syncTab: BraveSyncTab?
-  let faviconDriver: FaviconDriver?
+  
   
   var redirectURLs = [URL]()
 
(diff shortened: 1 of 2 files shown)


Sources/Brave/Frontend/Browser/Tab.swift
```diff
@@ -269,14 +295,14 @@ class Tab: NSObject {
     self.id = id
     rewardsId = UInt32.random(in: 1...UInt32.max)
     nightMode = Preferences.General.nightModeEnabled.value
-    syncTab = tabGeneratorAPI?.createBraveSyncTab(isOffTheRecord: type == .private)
+    _syncTab = tabGeneratorAPI?.createBraveSyncTab(isOffTheRecord: type == .private)
     
-    if let syncTab = syncTab {
-      faviconDriver = FaviconDriver(webState: syncTab.webState).then {
+    if let syncTab = _syncTab {
```
StephenHeaps, line 300: I don't think matters here but for consistency should this access remain on `syncTab` (instead of `_syncTab`) to match others?

StephenHeaps (approved):

Sources/Brave/Frontend/Browser/Tab.swift
Brandon-T: It shouldn't make a difference since this is done in the `constructor`, and the driver only cares about the internal `webState` which we don't control from the Swift side.

Then in `deinit` both are deallocated weakly :) so no worries.
I think in the constructor we should access the hard reference instead of dealing with weak. Everywhere else we should use weak.

PR #7588, opened 2023-06-08
Fix #5543: Relax SSL Certificate Validation to match all other browsers

### LINKED ISSUE #5543: SSL Trust Validation too strict [does not match Safari's relaxed validation]
### Description: 
- Brave-iOS SSL trust validation is too strict; it validates the BasicX509Policy and Mixed-Content
- All other browsers on iOS validate only the SSLTrustPolicy and do not validate Mixed-Content 
- iOS blocks Mixed-Content by default, so no need to validate it

### Steps to Reproduce 
  1. Visit `https://kevsong.com`
  2. Notice the icon in the URL bar + the certificate viewer says the website is not secure

**Actual result:** 
- Website is not secure

**Expected result:**
- Website should be secure as per other browsers relaxed trust validation

Easily reproduced

**Brave Version:** 
- Any

**Device details:** 
- Any

**Website problems only:**
- Did you check in Safari/Firefox (WkWebView-based browsers)? Yes
- Happens only in Brave-iOS

Brandon-T:
## Security Review
- https://github.com/brave/security/issues/1302

## Summary of Changes
- Relax SSL validation. 
- Use Chromium validation over Apple's validation.
- If Chromium returns a value indicating that the system should handle it, then we use Apple's validation. 
- Disable Apple's X509 validation and only validate it for SSL to match other browsers (including Safari).

This pull request fixes #5543

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:
- Visit https://office.harrisisi.com/index.jsp
- It should show "SECURE" in the URL bar, if Safari and all other browsers show it is secure.

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Frontend/Browser/BrowserViewController.swift
+++ b/Sources/Brave/Frontend/Browser/BrowserViewController.swift
@@ -1721,17 +1721,45 @@ public class BrowserViewController: UIViewController {
         break
       }
       
-      let host = tab.webView?.url?.host
+      guard let scheme = tab.webView?.url?.scheme,
+            let host = tab.webView?.url?.host else {
+        tab.secureContentState = .insecure
+        self.updateURLBar()
+        return
+      }
       
-      Task {
+      let port: Int
+      if let urlPort = tab.webView?.url?.port {
+        port = urlPort
+      } else if scheme == "https" {
+        port = 443
+      } else {
+        port = 80
+      }
+      
+      Task.detached {
         do {
-          try await BraveCertificateUtils.evaluateTrust(serverTrust, for: host)
-          tab.secureContentState = .secure
+          let result = BraveCertificateUtility.verifyTrust(serverTrust,
+                                                           host: host,
+                                                           port: port)
+          // Cert is valid!
+          if result == 0 {
+            tab.secureContentState = .secure
+          } else if result == Int32.min {
+            // Cert is valid but should be validated by the system
+            // Let the system handle it and we'll show an error if the system cannot validate it
+            try await BraveCertificateUtils.evaluateTrust(serverTrust, for: host)
+            tab.secureContentState = .secure
+          } else {
+            tab.secureContentState = .insecure
+          }
         } catch {
           tab.secureContentState = .insecure
         }
         
-        self.updateURLBar()
+        Task { @MainActor in
+          self.updateURLBar()
+        }
       }
     case ._sampledPageTopColor:
       updateStatusBarOverlayColor()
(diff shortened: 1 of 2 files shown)


Sources/Brave/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1721,17 +1721,45 @@ public class BrowserViewController: UIViewController {
...
-          tab.secureContentState = .secure
+          let result = BraveCertificateUtility.verifyTrust(serverTrust,
+                                                           host: host,
+                                                           port: port)
+          // Cert is valid!
+          if result == 0 {
+            tab.secureContentState = .secure
+          } else if result == Int32.min {
+            // Cert is valid but should be validated by the system
+            // Let the system handle it and we'll show an error if the system cannot validate it
```
stoletheminerals, line 1750: Do we know what triggers this?

Brandon-T: It's all handled in Brave-Core:
https://github.com/brave/brave-core/blob/master/ios/browser/api/net/certificate_utility.mm#L193
https://github.com/brave/brave-core/blob/master/ios/browser/api/net/certificate_utility.mm#L173

I can probably check the return type on Line 173 and if it's not equal to `SUCCESS` then I know the cert is invalid no matter what. That means it's both invalid for pinning and invalid as a cert, so I wouldn't have to validate it with any Apple APIs. But my thought when writing the above was that it's more flexible to check the system to see if it's valid on the iOS side and use the Chromium side only for checking if it can be pinned against.

IIRC, when I wrote it, Chromium only checked pinning and didn't validate it against the system certs. But it seems they do now: https://source.chromium.org/chromium/chromium/src/+/main:net/cert/cert_verify_proc_ios.cc;l=392;bpv=0;bpt=1

Chrome iOS still uses Apple to verify certs though, and not the Chromium APIs: https://source.chromium.org/chromium/chromium/src/+/main:ios/web/security/crw_cert_verification_controller.mm;l=202;bpv=0;bpt=1?q=CertVerifyProcIOS&ss=chromium%2Fchromium%2Fsrc

I think that's why I chose to return `Int32.min` and do the same on the iOS side :)
But yeah, the validation is done in Chromium and that's what triggers it.

iccub (approved):

Brandon-T:
Closing in favour of: https://github.com/brave/brave-ios/pull/8309

PR #7638, opened 2023-06-23
Fix #7636: Fix Tab's Title not restoring properly

### LINKED ISSUE #7636: Tabs do not restore their titles only url
### Steps to Reproduce 
  1. Open couple tabs
  2. Close the app
  3. Reopen

### Actual result: 
No titles are restored, only urls

### Brave Version: 

### Website problems only:

Brandon-T:
## Summary of Changes
- `WebView.title` is nil when a tab is lazily restored. The old tab logic did not support this, so the titles were always loaded as every webView immediately loaded
- To fix it, we just check if the webView is nil, and if it is, we show the restoring title instead of a URL.

This pull request fixes #7636

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:
- Open 3 tabs
- Kill the app
- Restore the app
- Notice: Tab Title should now be correct instead of a URL.

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Frontend/Browser/Tab.swift
+++ b/Sources/Brave/Frontend/Browser/Tab.swift
@@ -507,7 +507,7 @@ class Tab: NSObject {
       return Strings.Hotkey.newTabTitle
     }
 
-    if let url = self.url, !InternalURL.isValid(url: url), let shownUrl = url.displayURL?.absoluteString {
+    if let url = self.url, !InternalURL.isValid(url: url), let shownUrl = url.displayURL?.absoluteString, webView != nil {
       syncTab?.setTitle(shownUrl)
       return shownUrl
     }


Sources/Brave/Frontend/Browser/Tab.swift
```diff
@@ -507,7 +507,7 @@ class Tab: NSObject {
       return Strings.Hotkey.newTabTitle
     }
 
-    if let url = self.url, !InternalURL.isValid(url: url), let shownUrl = url.displayURL?.absoluteString {
+    if let url = self.url, !InternalURL.isValid(url: url), let shownUrl = url.displayURL?.absoluteString, webView != nil {
```
soner-yuksel, line 510: Oh so here it is setting the title again to URL

soner-yuksel (approved):

PR #7639, opened 2023-06-25
Ref #3939: Add support to Apple Screen Time Limits

### LINKED ISSUE #3939: Implement screentime
### Description: 
See this API https://developer.apple.com/documentation/screentime/stwebpagecontroller
It may be possible to block restricted pages based on screentime limits

### Steps to Reproduce 
  1.
  2.
  3.

**Actual result:** 

**Expected result:**

**Brave Version:** 

**Device details:** 

**Website problems only:**

domcorvasce:
## Summary of Changes

This pull request fixes #3939. It adds support for:

<details>
  <summary>tracking time spent on individual websites (unless you're in incognito mode)</summary>
  
</details> 

<details>
  <summary>enforcing usage limits per single web page</summary>
  
</details>

The Screen Time API is not well documented. I benefitted from reading https://github.com/mozilla-mobile/firefox-ios/pull/15111.

There is no way to test this feature using the Simulator&mdash;I've tried but Screen Time does not work. You need a real device.

## Submitter Checklist:

  - [x] Light & dark mode

### Background

1. Build the app on your real device
3. Go to Settings > Screen Time > turn it on if you didn't already
4. Go to Settings > Screen Time > App limits > Add Limit and add a one minute limit on a website of your choice

### Test 1

1. Given you visit the website in a non-private tab
2. Given you browse the website for one minute
3. Then you should see the Screen Time limit over view

### Test 2

1. Given you visit the website in a private tab
2. Given you browse the website for one minute
3. Then you won't see the Screen Time limit over view nor the page usage will appear in the Screen Time activity log

### Test 3

1. Given you visit the website in a non-private tab
2. Given you switch to an empty tab
3. Given you wait for one minute
4. Then you won't see a dedicated entry for the internal URL in the Screen Time activity log 

### Test 4

1. Given you visit the website in a non-private tab
2. Given you switch to an empty tab
3. Given you wait for one minute to show the Screen Time block view
4. Then the ST block view should go away when you change the current tab URL

#### Test 5

1. Given you visit the website in a non-private tab
2. Given you browse the website for one minute
3. Given you switch to a private tab and visit the same website
4. Then you should see the Screen Time limit over view

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Frontend/Browser/Tab.swift
+++ b/Sources/Brave/Frontend/Browser/Tab.swift
@@ -177,7 +177,7 @@ class Tab: NSObject {
     willSet {
       url = newValue
       previousComittedURL = committedURL
-      screenTimeViewController.url = isPrivate ? nil : url
+      screenTimeViewController.url = url
     }
   }
 
--- a/Sources/Brave/Frontend/Browser/Tab.swift
+++ b/Sources/Brave/Frontend/Browser/Tab.swift
@@ -13,6 +13,7 @@ import Data
 import os.log
 import BraveWallet
 import Favicon
+import ScreenTime
 
 protocol TabContentScriptLoader {
   static func loadUserScript(named: String) -> String?
@@ -65,6 +66,7 @@ enum TabSecureContentState {
 class Tab: NSObject {
   let id: UUID
   let rewardsId: UInt32
+  let screenTimeViewController = STWebpageController()
 
   var onScreenshotUpdated: (() -> Void)?
   var rewardsEnabledCallback: ((Bool) -> Void)?
@@ -73,7 +75,6 @@ class Tab: NSObject {
   var blockAllAlerts: Bool = false
   
   private(set) var type: TabType = .regular
-  
   
   var redirectURLs = [URL]()
 
@@ -176,9 +177,10 @@ class Tab: NSObject {
     willSet {
       url = newValue
       previousComittedURL = committedURL
+      screenTimeViewController.url = isPrivate ? nil : url
     }
   }
-  
+
   /// The previous url that was set before `comittedURL` was set again
   private(set) var previousComittedURL: URL?
   
@@ -307,6 +309,7 @@ class Tab: NSObject {
 
     super.init()
     self.type = type
+    self.screenTimeViewController.suppressUsageRecording = type.isPrivate
   }
 
   weak var navigationDelegate: WKNavigationDelegate? {
@@ -342,6 +345,11 @@ class Tab: NSObject {
       }
       let webView = TabWebView(frame: .zero, tab: self, configuration: configuration!, isPrivate: isPrivate)
       webView.delegate = self
+
+      webView.addSubview(screenTimeViewController.view)
+      screenTimeViewController.view.snp.makeConstraints {
+        $0.edges.equalToSuperview()
+      }
 
       webView.accessibilityLabel = Strings.webContentAccessibilityLabel
       webView.allowsBackForwardNavigationGestures = true


iccub:
Code looks legit, thanks @domcorvasce.
This PR will stay open for a bit, we have to test it well then make a security review for it

Sources/Brave/Frontend/Browser/BrowserViewController.swift
```diff
@@ -779,14 +782,26 @@ public class BrowserViewController: UIViewController {
...
+
     view.addSubview(alertStackView)
     view.addSubview(bottomTouchArea)
     view.addSubview(topTouchArea)
     view.addSubview(bottomBarKeyboardBackground)
     view.addSubview(footer)
     view.addSubview(header)
     view.addSubview(statusBarOverlay)
-    
+    view.addSubview(screenTimeViewContainer.view)
```
soner-yuksel: Can we use snapkit for the sake of unity?

```
stackView.snp.makeConstraints {
  screenTimeViewContainer.view.edges.equalToSuperview()
}
```

iccub (changes requested):
There's one change required.
Currently this implementation screen limit covers entire screen because it's added to the BrowserViewController.
We have to add it somewhere at the webview constraints level so the user can change to another tab.
It would also have to be one `STWebpageController` per tab.
I will try to update this code, find a good place for it

domcorvasce:
> Currently this implementation screen limit covers entire screen because it's added to the BrowserViewController.
> We have to add it somewhere at the webview constraints level so the user can change to another tab.

I was going for Safari behaviour but I double-checked and Safari attaches the controller to the specific tab.
So, yeah, we need to align that. I'll give it a try on the weekend ⚡ 

> It would also have to be one STWebpageController per tab.

Yet we only actively track the active tabs. [`suppressUsageRecording`](https://developer.apple.com/documentation/screentime/stwebpagecontroller/3552409-suppressusagerecording?language=objc) should be enabled once a tab becomes inactive.

domcorvasce:
@iccub I moved the logic into `Tab` and I've checked that tabs switch is possible. It now follows Safari's behaviour (check top PR description for the up-to-date screenshot).

One quick note about [Tab.swift:180](https://github.com/brave/brave-ios/pull/7639/files#diff-e1de1b26b97c278da85da2f7ed234633875e7038d34c41306b48cbdb2d82329fR180): even though we suppress usage recording, assuming the user already maxed out the available time we need to set the active URL to `null` when within a private tab so the ST view is not shown.

PS: Yes, I double-checked that inactive tabs don't count towards ST.

iccub (changes requested):
The placement is good now, I am able to switch tabs.

I noticed one bug that bypasses the screen limit though.

STR:
1. Open new tab
2. Enter url of the website you passed the limit on
3. It will show screen limit screen correctly 
4. Now press back arrow to take you to NTP, then forward back to the website

Result is the website does not show the limit screen anymore, you can refresh and navigate within that tab without any limits

Don't feel pressured to fix that if you don't want to, we can try to find the cause of the problem too

iccub:
I also noticed the screen limit is ignored in private mode. Safari does not do that. It is a product decision however, we have to decide on that behavior.
Should be fine since all screen time data is saved locally only

domcorvasce:
> Result is the website does not show the limit screen anymore, you can refresh and navigate within that tab without any limits

Good catch. Now I’m eager to learn why it’s happening 😅
I see the following error message when switching back to the initial page (and it seems to only happen with that one):

```
The connection to service with pid XYZ named com.apple.ScreenTime.ScreenTimeWebExtension.viewservice
was interrupted, but the message was sent over an additional proxy
and therefore this proxy has become invalid.
```

> I also noticed the screen limit is ignored in private mode. Safari does not do that. It is a product decision however, we have to decide on that behavior.

I see. They’re attaching the ST controller to block websites that hit the usage limit in non private tabs but they don’t actively track page usage in private tabs (try browsing a website in private mode and it won’t show up in your Screen Time breakdown)

arealestate:
Question -- would this update also replicate Safari's behavior of syncing time spent on individual websites with ScreenTime, as opposed to aggregating and simply showing cumulative time spent on "Brave Browser"?

Thanks for building this! Have been waiting for this feature for a loooooong time! You all rock!

domcorvasce:
> Would this update also replicate Safari's behavior of syncing time spent on individual websites with ScreenTime, as opposed to aggregating and simply showing cumulative time spent on "Brave Browser"?

Yep

arealestate:
> > Would this update also replicate Safari's behavior of syncing time spent on individual websites with ScreenTime, as opposed to aggregating and simply showing cumulative time spent on "Brave Browser"?
> 
> Yep

Grazie mille!

iccub:
small update on this one. i ran this branch for a while and i've been getting multiple random crash related to screen time. we are investigating the probme

link1day:
I have Brave browser freshly downloaded for iOS. Screentime /time limit websites  is not working.  Is there a solution on this page, can someone explain further? Do I need to just download something from here to add to that browser?

bgala:
Can this support extent to private browsing mode?

iccub:
@bgala the time limits I hope yes.
Recording web usage no.

We will try to match what Safari does

iccub:
There's been couple bigger problems and I made a new one here https://github.com/brave/brave-ios/pull/7983

PR #7689, opened 2023-07-06
Fix #7692: Improve cosmetic filters script performance

### LINKED ISSUE #7692: Web compat issues on youtube when scrolling/reloading page [iPad]
### Description: 
Web compat issues on youtube when scrolling/reloading page [iPad]

Related to: https://github.com/brave/brave-ios/issues/7655

### Steps to Reproduce 
  1. Install 1.52.x
  2. Navigate to youtube.com
  3. Play some videos, scroll through the comments/related videos
  4. Try reloading the page and adding video to playlist and then return from playlist

### Actual result: 
Content fails to load/starts to freeze up

### Expected result:
Content should load and navigation should be fluid

Always

### Brave Version:  Any 1.52.x

### Device details: 
iPadOS

cuba:
## Summary of Changes

This pull request fixes #7692

In this PR are a number of fixes:
1. Fixed a potential for `usePolling` from being called more than once. This is the main fix here. This is a really bad issue because of the following:
The poller is designed to run for only 10 seconds and then switch back to the mutation observer. Because it sets a global variable `selectorsPollingIntervalId` a second poller would replace that variable, this means that when a second poller starts and sets the variable, the first poller is no longer referenced and can not be canceled. After 10 seconds it starts another mutation observer, the second poller is cancelled but not the first one, it then keeps running starting a new mutation observer every time it runs. Each new mutation observer then may start new pollers. This results in mutation observers and pollers starting with no end killing the web-page and making it completely unusable. The only fix at this point is to refresh the page. Because YouTube is a single page application that almost never reloads, this problem persists when navigating.

This is solved with the following line:
```
if (selectorsPollingIntervalId !== undefined) { return }
```

2. Synced code that was different from desktop. On a mutation observer when a new node is added it will add elements into a queue rather than adding the selectors right away. We extract the selectors only when calling `sendPendingSelectorsIfNeeded`. This is reflecting a change on desktop [here](https://github.com/brave/brave-core/commit/276b8063057911996f22a5de781bdb76a9556f28).

3. Synced changes that happened in desktop that were missing on iOS found [here](https://github.com/brave/brave-core/commit/78bb36e83ce1ec4644a05145aa056aef2fa648c9). This removes an unnecessary call to `queueSelectorsFromMutations` (`queueAttrsFromMutations` on desktop) and clears the  `notYetQueriedElements`.

4. Throttle sending of selectors, moving stylesheet and rewriting css rules so they don't happen often when new elements are frequently added/modified such as in those cases on YouTube when scrolling through comments, these are seen by adding `moveStyleThrottled`, `setRulesOnStylesheetThrottled` and `sendPendingSelectorsThrottled`.

## Submitter Checklist:

  - [ ] Light & dark mode

## Test Plan:

See STR in issue

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Sandboxed/SelectorsPollerScript.js
+++ b/Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Sandboxed/SelectorsPollerScript.js
@@ -59,6 +59,10 @@ window.__firefox__.execute(function($) {
   const scoreCalcIntervalMs = 1000
   // The begin of the time frame to calc |currentMutationScore|.
   let currentMutationStartTime = performance.now()
+  // Elements that are not yet queried for selectors
+  const notYetQueriedElements = []
+  // The query to perform when extracting classes and ids
+  const classIdWithoutHtmlOrBody = '[id]:not(html):not(body),[class]:not(html):not(body)'
 
   const CC = {
     allSelectors: new Set(),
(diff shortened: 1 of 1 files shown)


pes10k:
cc @antonok-edm too, incase we need to mirror this in brave-core

cuba:
This PR is replaced with the following: https://github.com/brave/brave-ios/pull/7741

In this PR, the commits are broken down more logically according to their purpose so its easier to follow and see what exactly is fixed.

PR #7878, opened 2023-08-15
Fix #7849: Update Sync Local Authentication Logic

### LINKED ISSUE #7849: Update Sync Local Authentication Logic
### Description: 
Linked to https://github.com/brave/brave-ios/issues/7436
Slack: https://bravesoftware.slack.com/archives/C0816QH3R/p1689972393401329?thread_ts=1685503730.825809&cid=C0816QH3R

Please update the logic when showing the PIN as follows

- Request a PIN/Biometrics when the user decides to sync All or Passwords
- Request a PIN/Biometrics when the user wants to add new device to sync

1. ask for auth when user toggle password-sync option or wants to add a device to sync
2. if no auth enabled at all you are not allowed to disable or enable password sync
3. If no auth set up when you tap on the password sync toggle show user an alert that they need to enable pincode/faceid. 

And take the user to iOS settings, or privacy settings if possible
Use the current lockscreen UI, alert should be shown for point 3

soner-yuksel:
## Summary of Changes

This pull request is changing the criteria for Sync Local Authentication Logic criterias.

Previous version has to authenticate user if user is viewing sync setting or creating a sync chain. 

New criteria suggests  

1. Ask for authentication when user toggle password-sync option or wants to add a device to sync
2. If no auth enabled at all you are not allowed to disable or enable password sync
3. If no auth set up when you tap on the password sync toggle show user an alert that they need to enable pincode/faceid.

This pull request fixes #7849

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan _ Screenshots::

Test 1 

-Create a or join a Sync Chain
-Disable PIN Code entirely
-Try to toggle password sync

test662

Test 2 

-Create a or join a Sync Chain
-Disable FaceID - Touch ID but do not disable Pincode
-Check adding device and toggling password logic (this should include cancel)
-Enable Face ID or Touch Id again
-Check adding device and toggling password logic (this should include also cancel functionality)

https://github.com/brave/brave-ios/assets/6643505/e87facf4-0f82-4a20-8ba1-d82b4c7c8655

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Frontend/Sync/SyncSettingsTableViewController.swift
+++ b/Sources/Brave/Frontend/Sync/SyncSettingsTableViewController.swift
@@ -264,7 +264,6 @@ class SyncSettingsTableViewController: SyncViewController, UITableViewDelegate, 
 
   @objc private func didToggleSyncType(_ toggle: UISwitch) {
     let toggleExistingStatus = !toggle.isOn
-    toggle.setOn(toggleExistingStatus, animated: false)
     
     askForAuthentication() { [weak self] status, error in
       guard let self = self, status else {
--- a/Sources/Brave/Frontend/Sync/SyncSettingsTableViewController.swift
+++ b/Sources/Brave/Frontend/Sync/SyncSettingsTableViewController.swift
@@ -84,14 +84,18 @@ class SyncSettingsTableViewController: SyncViewController, UITableViewDelegate, 
        syncAPI: BraveSyncAPI,
        syncProfileService: BraveSyncProfileServiceIOS,
        tabManager: TabManager,
-       windowProtection: WindowProtection?,
-       requiresAuthentication: Bool) {
+       windowProtection: WindowProtection?) {
     self.isModallyPresented = isModallyPresented
     self.syncAPI = syncAPI
     self.syncProfileService = syncProfileService
     self.tabManager = tabManager
     
-    super.init(windowProtection: windowProtection, requiresAuthentication: requiresAuthentication, isModallyPresented: isModallyPresented)
+    // Local Authentication (Biometric - Pincode) needed only for actions
+    // Enabling - disabling password sync and add new device
+    super.init(windowProtection: windowProtection,
+               requiresAuthentication: false,
+               isModallyPresented: isModallyPresented,
+               dismissPresenter: false)
   }
 
   required init?(coder: NSCoder) {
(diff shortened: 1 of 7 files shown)


Sources/Brave/Frontend/Sync/SyncSettingsTableViewController.swift
```diff
@@ -259,25 +263,36 @@ class SyncSettingsTableViewController: SyncViewController, UITableViewDelegate,
...
+      toggle.setOn(!toggleExistingStatus, animated: false)
+
+      switch toggle.tag {
+      case SyncDataTypes.bookmarks.rawValue:
+        Preferences.Chromium.syncBookmarksEnabled.value = !toggleExistingStatus
+      case SyncDataTypes.history.rawValue:
+        Preferences.Chromium.syncHistoryEnabled.value = !toggleExistingStatus
+      case SyncDataTypes.passwords.rawValue:
+        Preferences.Chromium.syncPasswordsEnabled.value = !toggleExistingStatus
+      case SyncDataTypes.openTabs.rawValue:
```
Brandon-T: I think this can be:

```swift
guard let syncDataType = SyncDataTypes(rawValue: toggle.tag) else {
   Logger.module.error("Invalid Sync DataType")
   return
}

switch syncDataType {
case .bookmarks: ...
case .history: ...
case .passwords: ...
case .openTabs: ...
}
```

with no `default` case so that if the types ever change, it would give a compilation error when the case is not implemented.

Sources/Brave/Frontend/Sync/SyncSettingsTableViewController.swift
```diff
@@ -307,11 +322,15 @@ class SyncSettingsTableViewController: SyncViewController, UITableViewDelegate,
 
   @objc
   private func onSyncInternalsTapped() {
-    let syncInternalsController = syncAPI.createSyncInternalsController().then {
-      $0.title = Strings.braveSyncInternalsTitle
+    askForAuthentication() { [weak self] status, error in
+      guard let self = self, status else { return }
+      
+      let syncInternalsController = syncAPI.createSyncInternalsController().then {
```
Brandon-T: Do you need `self.syncAPI`?

Brandon-T (changes requested):

Sources/Brave/Frontend/Widgets/LoadingViewController.swift
```diff
@@ -37,6 +37,15 @@ public class LoadingViewController: UIViewController {
 public class AuthenticationController: LoadingViewController {
   enum AuthViewType {
     case sync, tabTray
+    
+    var description: String {
```
Brandon-T: I would rename this to something else.
`description` is pretty much discouraged everywhere: https://developer.apple.com/documentation/swift/string/description

Same for dictionary and so on.

Brandon-T (changes requested):

Sources/Brave/Frontend/Sync/SyncSettingsTableViewController.swift
soner-yuksel: Yeah great catch fix at https://github.com/brave/brave-ios/pull/7878/commits/2873a2ad927225f02157498543c1e57e183f24ee

Sources/Brave/Frontend/Sync/SyncSettingsTableViewController.swift
soner-yuksel: Makes sense, and seems cleaner fix at https://github.com/brave/brave-ios/pull/7878/commits/2873a2ad927225f02157498543c1e57e183f24ee

Sources/Brave/Frontend/Widgets/LoadingViewController.swift
soner-yuksel: Fixed at https://github.com/brave/brave-ios/pull/7878/commits/2873a2ad927225f02157498543c1e57e183f24ee

fmarier (approved):
I didn't review the code, but the description and screenshot/video look good to me. It seems to match what was agreed to.

The only thing I will note is that asking for biometric is only important IMO before **enabling** password sync. We could let users **disable** password syncing without authenticating. Right now, it seems like we prompt no matter what. It's fine like that, but not really necessary

Brandon-T (approved):

PR #7890, opened 2023-08-17
Fix#7792: Add cookie consent notice popup

### LINKED ISSUE #7792: [iOS] Opt-out cookie consent notice
## Description
Implement the cookie consent notice as opt-out. Same display logic as the old notice before it was disabled.

image

- Got it = keep cookie consent filter enabled (it's enabled by default).
- Stop blocking = disable cookie consent filter.

### Figma
https://www.figma.com/file/cL22PpqneTYxJePORxUBKz/Cookie-consent-blocker?type=design&node-id=1242%3A44596&mode=design&t=FSGniL2oAsfIVxUT-1

cuba:
## Summary of Changes
Add a dialogue informing the user of cookie consent notice blocking

This pull request fixes #7792 
## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:

Launch the application several times and go through some onboarding. At one point you should get this dialogue.

## Screenshots:

Simulator Screenshot - iPhone 14 Pro - 2023-08-17 at 09 40 57
Simulator Screenshot - iPhone 14 Pro - 2023-08-17 at 09 41 09

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController+Callout.swift
```diff
@@ -181,6 +183,25 @@ extension BrowserViewController {
     present(controller, animated: true)
   }
   
+  private func presentCookieNotificationBlockingCallout(skipSafeGuards: Bool = false) {
```
iccub, line 186: Do we show this popup for users who already saw the original 'opt in' popup?

Sources/Onboarding/ProductNotifications/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,97 @@
...
+    fatalError("init(coder:) has not been implemented")
+  }
+  
+  public override func viewDidLoad() {
+    super.viewDidLoad()
+    view.backgroundColor = UIColor.braveBackground
+  }
+}
+
+extension Task where Success == Never, Failure == Never {
```
iccub: I believe we have this helper already in the codebase in `AsyncAwait` file

iccub (changes requested):
One design decision to make sure is correct. I believe we should not show the opt out for existing users, as they saw the first opt in prompt already

Sources/Onboarding/ProductNotifications/CookieNotificationBlockingConsentView.swift
```diff
@@ -0,0 +1,97 @@
...
+          
+          Text(Strings.blockCookieConsentNoticesPopupTitle).font(.title)
+          Text(Strings.blockCookieConsentNoticesPopupDescription).font(.body)
+        }
+        .padding(Self.textPadding)
+        .foregroundColor(Color(UIColor.braveLabel))
+        .multilineTextAlignment(.center)
+        .fixedSize(horizontal: false, vertical: true)
+        
+        Button(Strings.blockCookieConsentNoticesDismissButtonTitle,
```
iccub, line 35: In Figma there's "Stop blocking" button as well. Do we not add it?

cuba:
> One design decision to make sure is correct. I believe we should not show the opt out for existing users, as they saw the first opt in prompt already

Well existing users will not get this popup because I'm using the same setting value that opt in users got. Their settings will stay untouched and this popup won't appear since existing users already got a chance to chose what they wanted

Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController+Callout.swift
cuba: Again, no. It used the same setting as before as the name fit

Sources/Onboarding/ProductNotifications/CookieNotificationBlockingConsentView.swift
cuba: I understood that this is no longer needed. Perhaps I was mistaken.

Sources/Onboarding/ProductNotifications/CookieNotificationBlockingConsentView.swift
cuba: It is defined yes, but it's in the `Brave` package. Brave uses `Onboarding` as a dependency (not the other way around) so we would have to move this extension to something like "Shared". This might result in much larger changes.

cuba: That being said that sleep is no longer used so I can remove that extension (it was used for the animation which we don't have anymore)

cuba: Removed

iccub (approved):
we had some small UI concerns. they are low priority and can be fixed separately

Issue #2440, opened 2020-04-02
Implement certificate pinning for Brave/BraveSoftware domains


jumde:
### Description: 
Certificate Pinning reduces the impact of MITM attacks. We should PIN certificates for *.brave.com and *.bravesoftware.com domains to ensure the responses from these domains can be trusted. 

Certificate Pinning is already implemented: https://github.com/brave/brave-ios/blob/f8b4d52df2cea77ac624a8621def3825f89511dd/BraveShared/CertificatePinning.swift - Just needs to be enabled.

iccub:
Does this block the super referrer work or can be implemented later/separately?

diracdeltas:
@iccub not blocking

PR #8027, opened 2023-09-05
Fix #8026: Fixed BasicAuthCredentialsManager to use credentials from WebKit

### LINKED ISSUE #8026: Store Brave staging basic auth credentials for endpoint usage
Since we dropped the sqlcipher and non apple basic auth credential storage this has been broken.
The basic auth creds save to webkit storage but when you issue a separate endpoint it does not forward those saved credentials.
We will fix it by injecting those credential is a Brave domain is detected.

This should help with Playlist and Search staging endpoints

Brandon-T:
## Summary of Changes

This pull request fixes #8026

## Submitter Checklist:

- [] *Unit Tests* are updated to cover new or changed functionality
  - [x] Light & dark mode

## Test Plan:
- Test Brave-Search APIs working on Staging
- Test Playlist Shared Folders working on Staging

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Frontend/Browser/Playlist/BasicAuthCredentials.swift
+++ b/Sources/Brave/Frontend/Browser/Playlist/BasicAuthCredentials.swift
@@ -13,13 +13,12 @@ class BasicAuthCredentialsManager: NSObject, URLSessionDataDelegate {
   
   public static let validDomains: Set<String> = [
     // Brave Search
-    "search.brave.com", "search.brave.software",
-    "search.bravesoftware.com", "safesearch.brave.com",
+    "search.brave.software", "search.bravesoftware.com",
     "safesearch.brave.software", "safesearch.bravesoftware.com",
     "search-dev-local.brave.com",
     
     // Playlist
-    "playlist.bravesoftware.com", "playlist.brave.com"
+    "playlist.bravesoftware.com"
   ]
   
   static func setCredential(origin: String, credential: URLCredential?) {
--- a/Sources/Brave/Frontend/Browser/Playlist/BasicAuthCredentials.swift
+++ b/Sources/Brave/Frontend/Browser/Playlist/BasicAuthCredentials.swift
@@ -5,44 +5,87 @@
 
 import Foundation
 import UIKit
+import BraveCore
+import os.log
 
 class BasicAuthCredentialsManager: NSObject, URLSessionDataDelegate {
-  private let domains: [String]
+  private static var credentials = [String: URLCredential]()  // ["origin:port": credential]
   
-  init(for domains: [String]) {
-    self.domains = domains
+  public static let validDomains: Set<String> = [
+    // Brave Search
+    "search.brave.com", "search.brave.software",
+    "search.bravesoftware.com", "safesearch.brave.com",
+    "safesearch.brave.software", "safesearch.bravesoftware.com",
+    "search-dev-local.brave.com",
+    
+    // Playlist
+    "playlist.bravesoftware.com", "playlist.brave.com"
+  ]
+  
+  static func setCredential(origin: String, credential: URLCredential?) {
+    credentials[origin] = credential
   }
 
   func urlSession(
     _ session: URLSession,
     didReceive challenge: URLAuthenticationChallenge
   ) async -> (URLSession.AuthChallengeDisposition, URLCredential?) {
-    guard challenge.protectionSpace.authenticationMethod == NSURLAuthenticationMethodHTTPBasic ||
-            challenge.protectionSpace.authenticationMethod == NSURLAuthenticationMethodHTTPDigest ||
-            challenge.protectionSpace.authenticationMethod == NSURLAuthenticationMethodNTLM else {
-      return (.performDefaultHandling, nil)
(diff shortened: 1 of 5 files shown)


iccub (approved):

Sources/Brave/Frontend/Browser/Playlist/BasicAuthCredentials.swift
```diff
@@ -5,44 +5,87 @@
...
   ) async -> (URLSession.AuthChallengeDisposition, URLCredential?) {
-    guard challenge.protectionSpace.authenticationMethod == NSURLAuthenticationMethodHTTPBasic ||
-            challenge.protectionSpace.authenticationMethod == NSURLAuthenticationMethodHTTPDigest ||
-            challenge.protectionSpace.authenticationMethod == NSURLAuthenticationMethodNTLM else {
-      return (.performDefaultHandling, nil)
-    }
-
-    if !domains.contains(challenge.protectionSpace.host) {
-      return (.performDefaultHandling, nil)
+    // Certificate Pinning
```
stoletheminerals, line 32: What's the reason to have a separate pinning check here?

Sources/Brave/Frontend/Browser/Playlist/BasicAuthCredentials.swift
```diff
@@ -5,44 +5,87 @@
...
 
 class BasicAuthCredentialsManager: NSObject, URLSessionDataDelegate {
-  private let domains: [String]
+  private static var credentials = [String: URLCredential]()  // ["origin:port": credential]
   
-  init(for domains: [String]) {
-    self.domains = domains
+  public static let validDomains: Set<String> = [
+    // Brave Search
+    "search.brave.com", "search.brave.software",
```
stoletheminerals: Do we need a basic auth for search.brave.com?

stoletheminerals (commented):
A few questions to better understand the logic behind this change.
1. Are basic auth creds going to be hardcoded or will it require a user to visit an auth protected endpoint and enter the creds first?
2. Is this change aimed for API calls to these endpoints (e.g an API call to Brave Search suggestion endpoint) and not normal web browsing?

Sources/Brave/Frontend/Browser/Playlist/BasicAuthCredentials.swift
Brandon-T: I have it there in case the API call needs to be pinned.

This is used only for API calls and not for WebKit calls, so I make sure that those are also pinned.

Sources/Brave/Frontend/Browser/Playlist/BasicAuthCredentials.swift
Brandon-T: No I don't think we do. I can remove it. The old code had it, so I copied that over. But I just checked and it isn't needed :)

Brandon-T:
> A few questions to better understand the logic behind this change.
> 
> 1. Are basic auth creds going to be hardcoded or will it require a user to visit an auth protected endpoint and enter the creds first?
> 
> 2. Is this change aimed for API calls to these endpoints (e.g an API call to Brave Search suggestion endpoint) and not normal web browsing? 

Both correct.

It requires the user to visit the page first. This is true for both playlist and brave search and will be true for any future API calls. Otherwise the dev would have to hardcode it, and that would require a separate sec review.

Atm, we always have to visit the page anyway and we get promoted for credential input.

This change is aimed only at API calls and not for normal web browsing :)

Sources/Brave/Frontend/Browser/Playlist/BasicAuthCredentials.swift
stoletheminerals: Do we not already pin brave hosts in API calls? The ssl pinning is meant to be for API calls specifically. That's why I'm surprised we have additional code here.

Brandon-T: We pin API calls for playlist and DAU, but I don't think we had pinning for search APIs as we didn't really make any (the calls go to Google as the fallback). AFAIK, all search calls happened in the webView as the user searches via typing into the page itself. All our WebKit calls are pinned.

stoletheminerals: Maybe not as a part of this PR, but can we enable SSL pinning for other Brave API calls as well? All Web3 calls to our proxies are not pinned, ideally we need to pin all the API calls to Brave domains in https://github.com/brave/brave-core/blob/master/chromium_src/net/tools/transport_security_state_generator/input_file_parsers.cc
cc @StephenHeaps

stoletheminerals (approved):

Sources/Brave/Frontend/Browser/Playlist/BasicAuthCredentials.swift
iccub: Ticket is here https://github.com/brave/brave-ios/issues/2440
We will triage it next week

PR #8030, opened 2023-09-05
Fix #8019: Fix issues with farbling plugins on iOS 17

### LINKED ISSUE #8019: [iOS 17] Fingerprinting is broken and showing the same values on both pages and after relaunching Brave when Shields enabled
### Description: 

[iOS 17] Fingerprinting is broken and showing the same values on `both pages` and after `relaunching Brave` when `Shields enabled`

### Steps to Reproduce 
  1. Launch Brave
  2. Go to https://dev-pages.brave.software/fingerprinting/farbling.html
  3. Make sure `Brave Shields` is enabled > Generate fingerprints
  4. Check results for `Plugins`, `Hardware connectivity and `Speech voices`
  5. For each separate feature the value must be the same in every column, but different for each feature
  6. Close and relaunch Brave > Run the test again
  7. Compare values with results in the previous step  > Observe that values are the same
  8. Open another page https://dev-pages.bravesoftware.com/fingerprinting/farbling.html
  9. Run the test and compare results > Observe that values are the same
  
Note: Hardware concurrency is hit or miss. We farble number of cores between 2-8 (on iOS 14 and can be less on older devices). With such a low variety of numbers, it's possible that both both test pages get the same value. This can be seen by killing the app and relaunching it. A new random value is given to the farbler and you will see the same values repeating on average 1/7 times as there are only 1/7 possible values we can use.

### Actual result: 

When `Brave Shields` is enabled on a device with `iOS 17` the values for farbling tests are the same. On iPad with iPadOS 16.5.1 the values are different as expected, so I believe this is `iOS 17` specific issue.

|1|2|3|4|
|:-:|:-:|:-:|:-:|
|  |  |  |  |

### Expected result:

Results must have different values for farbling tests.

Easily reproduced

### Brave Version: 

Brave build: 1.57 (23.9.1.15)
BraveCore: 1.57.53 (116.0.5845.114)

### Device details: 

iPhone 14 [iOS 17.0]

### Website problems only:

cuba:
## Summary of Changes

This pull request fixes #8019

## Submitter Checklist:

  - [ ] Light & dark mode

## Test Plan:

STR in issue

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Paged/FarblingProtectionScript.js
+++ b/Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Paged/FarblingProtectionScript.js
@@ -247,25 +247,41 @@ window.__firefox__.execute(function($) {
     // always be within the expected range of -1 and 1.
     // This small decrease should not affect affect legitimite users of this api.
     // But will affect fingerprinters by introducing a small random change.
-    const fudgeFactor = args['fudgeFactor']
-    farbleAudio(fudgeFactor)
+    try {
+      const fudgeFactor = args['fudgeFactor']
+      farbleAudio(fudgeFactor)
+    } catch (error) {
+      console.error(`Failed to farble audio: ${error}`)
+    }
 
     // Fake data that is to be used to construct fake plugins
-    const fakePluginData = args['fakePluginData']
-    farblePlugins(fakePluginData)
+    try {
+      const fakePluginData = args['fakePluginData']
+      farblePlugins(fakePluginData)
+    } catch (error) {
+      console.error(`Failed to farble plugins: ${error}`)
+    }
 
     // A value representing a fake voice name that will be used to add a fake voice
-    const fakeVoiceName = args['fakeVoiceName']
-    // This value is used to get a random index between 0 and voices.length
-    // It's important to have a value between 0 - 1 in order to be within the
-    // array bounds
-    const randomVoiceIndexScale = args['randomVoiceIndexScale']
-    farbleVoices(fakeVoiceName, randomVoiceIndexScale)
+    try {
+      const fakeVoiceName = args['fakeVoiceName']
+      // This value is used to get a random index between 0 and voices.length
+      // It's important to have a value between 0 - 1 in order to be within the
+      // array bounds
+      const randomVoiceIndexScale = args['randomVoiceIndexScale']
+      farbleVoices(fakeVoiceName, randomVoiceIndexScale)
+    } catch (error) {
+      console.error(`Failed to farble voices: ${error}`)
+    }
 
     // This value lets us pick a value between 2 and window.navigator.hardwareConcurrency
     // It is a value between 0 and 1. For example 0.5 will give us 3 and
     // thus return 2 + 3 = 5 for hardware concurrency
-    const randomHardwareIndexScale = args['randomHardwareIndexScale']
-    farbleHardwareConcurrency(randomHardwareIndexScale)
+    try {
+      const randomHardwareIndexScale = args['randomHardwareIndexScale']
+      farbleHardwareConcurrency(randomHardwareIndexScale)
+    } catch (error) {
+      console.error(`Failed to farble hardware concurrency: ${error}`)
+    }
   })();
 });
--- a/Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Paged/FarblingProtectionScript.js
+++ b/Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Paged/FarblingProtectionScript.js
@@ -145,9 +145,7 @@ window.__firefox__.execute(function($) {
         
         // Farble the `item` method on the plugins array
         const originalItem = window.navigator.plugins.item
-        pluginsPrototype.item = function () {
-          let index = arguments[0]
-          
+        pluginsPrototype.item = function (index) {
           if (index < originalPluginsLength) {
             return Reflect.apply(originalItem, this, arguments)
           } else {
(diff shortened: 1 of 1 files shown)


Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Paged/FarblingProtectionScript.js
```diff
@@ -130,34 +130,45 @@ window.__firefox__.execute(function($) {
         // We need the original length so we can reference it (as we will change it)
         const plugins = window.navigator.plugins
         const originalPluginsLength = plugins.length
-
+        const pluginsPrototype = Object.getPrototypeOf(window.navigator.plugins)
```
Brandon-T, line 133: Why are we still using `plugins`? It's deprecated: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/plugins and will be removed.

This code doesn't check for the existence of it before using it or attempting to get the prototype of it. If it ever returned `null` or `undefined` this would cause the entire script to fail loading as it'd give an error.

Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Paged/FarblingProtectionScript.js
```diff
@@ -130,34 +130,45 @@ window.__firefox__.execute(function($) {
...
+        const originalItem = window.navigator.plugins.item
+        pluginsPrototype.item = function () {
+          let index = arguments[0]
+          
           if (index < originalPluginsLength) {
-            return Reflect.apply(originalItemFunction, plugins, arguments)
+            return Reflect.apply(originalItem, this, arguments)
           } else {
-            return plugins[index]
+            const farbledIndex = index - originalPluginsLength
```
Brandon-T, line 152: If originalPluginsLength > index, you'd end up with a negative farbledIndex here.
Is it possible to have an originalPluginsLength > index?

Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Paged/FarblingProtectionScript.js
```diff
@@ -130,34 +130,45 @@ window.__firefox__.execute(function($) {
...
-
-        // Patch `PluginArray.item(index)` function to return the correct item
-        // otherwise it returns `undefined`
-        const originalItemFunction = plugins.item
-        window.PluginArray.prototype.item = function (index) {
+        
+        // Farble the `item` method on the plugins array
+        const originalItem = window.navigator.plugins.item
+        pluginsPrototype.item = function () {
+          let index = arguments[0]
```
Brandon-T: Why do `pluginsPrototype.item = function()` + `let index = arguments[0]` instead of `pluginsPrototype.item = function(index)`?

Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Paged/FarblingProtectionScript.js
```diff
@@ -130,34 +130,45 @@ window.__firefox__.execute(function($) {
...
-            return plugins[index]
+            const farbledIndex = index - originalPluginsLength
+            return fakePlugins[farbledIndex]
           }
         }
+        
+        // Farble the `namedItem` method on the plugins array
+        const originalNamedItem = window.navigator.plugins.namedItem
+        pluginsPrototype.namedItem = function () {
+          let name = arguments[0]
```
Brandon-T: Same thing. `.namedItem = function(name)`

Brandon-T (changes requested):

Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Paged/FarblingProtectionScript.js
cuba: There is a check just 5 lines above

Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Paged/FarblingProtectionScript.js
cuba: Yes I could change this. I prefer arguments in case some new params are added to the API. though highly unlikely.

Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Paged/FarblingProtectionScript.js
cuba: this is in an `else` of an `if` check which guarantees index is smaller than originalPluginsLength so the index may only be equal or greater than original plugins length. A negative value is impossible here.

cuba: I should also point out that if a user were to pass a negative value, this would fall into the if statement which then would get passed to the original method and handled as if unfarbled.

Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Paged/FarblingProtectionScript.js
Brandon-T: Hmm, if new parameters were added, you'd have to change the code to use those anyway, so you could always change the function signature.

Brandon-T (approved):

PR #8048, opened 2023-09-11
Fix #7835: Add Suppress button to Allow link to switch apps? 

### LINKED ISSUE #7835: [hackerone] Add `Suppress` button to `Allow link to switch apps?` pop-up
https://hackerone.com/reports/2098015

soner-yuksel:
## Summary of Changes

Handling possible abuse of external application alerts by adding suppress alert dialog when alerts triggered to be shown multiple times. 

This pull request fixes #7835

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:

- Test to open a website that triggers facetime-audio repeatedly
- Check open external alert is shown and website cant trigger more alerts while pop up is active
- Do not allow external app launch
- Check instead of open external alert, an alert with suppress app launch is presented
- Check suppress works as expected
- Check show switch app alert works expected

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Frontend/Browser/BrowserViewController.swift
+++ b/Sources/Brave/Frontend/Browser/BrowserViewController.swift
@@ -2778,6 +2778,13 @@ extension BrowserViewController: TabDelegate {
     handleIPFSSchemeURL(url, visitType: .unknown)
   }
 
+  func didReloadTab(_ tab: Tab) {
+    // Resetting External Alert Properties
+    externalAppAlertCounter = 0
+    isExternalAppAlertPresented = false
+    isExternalAppAlertSuppressed = false
+  }
+  
   @MainActor
   private func isPendingRequestAvailable() async -> Bool {
     let privateMode = privateBrowsingManager.isPrivateBrowsing
--- a/Sources/Brave/Frontend/Browser/BrowserViewController.swift
+++ b/Sources/Brave/Frontend/Browser/BrowserViewController.swift
@@ -266,12 +266,6 @@ public class BrowserViewController: UIViewController {
   
   var processAddressBarTask: Task<(), Never>?
   var topToolbarDidPressReloadTask: Task<(), Never>?
-  
-  /// Boolean tracking custom url-scheme alert presented
-  var isExternalAppAlertPresented = false
-  var externalAppAlertCounter = 0
-  var isExternalAppAlertSuppressed = false
-  var externalAppURLOrigin: String?
 
   public init(
     windowId: UUID,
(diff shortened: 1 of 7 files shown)


stoletheminerals:
Thanks for fixing this @soner-yuksel. Unfortunately it has the same issue as before, just a different prompt now.
The PoC I'm using:
```html
<a id="link" href="maps://test">OPEN MAPS</a>
<script>
while(true){
document.getElementById('link').click();
}
</script>
```

This will keep opening the `Suppress alerts` prompt now, instead of the `Allow link to switch app` prompt

https://github.com/brave/brave-ios/assets/10810135/ab687955-5d8c-4f2d-ba3d-8a41742c962a

I'd change a logic here a little bit to make it similar to JavaScript alert suppression. We also don't need a new prompt for this.
1) First and second prompt - we show the user `Allow link to switch app` message as before.
2) On third and all consecutive prompts - we add a button or a checkbox to suppress all the next prompts on this page. Ideally the button should be in the same prompt.
3) The counter should reset if the user reloads the page

soner-yuksel:
@stoletheminerals the logic is changed like you requested

1. First and second prompt - we show the user Allow link to switch app message as before.

2. On third and all consecutive prompts - we SWAP the "Dont Allow" button with "Suppressed Alerts" for all the next prompts on this page.

3. The counter should reset if the user reloads the page and different URL should reset values

stoletheminerals:
@soner-yuksel 
This script will exploit the fact that the counter is URL specific and not tab specific:
```html
<a id="link" href="maps://test">OPEN MAPS</a>
<script>
const stateObj = { foo: "bar" };

function delay(time) {
  return new Promise(resolve => setTimeout(resolve, time));
}

function clickLoop() {
	delay(1000).then(() => {
		history.pushState(stateObj, '', Math.random());
		document.getElementById('link').click();
		clickLoop();
	})
}
clickLoop();

</script>
```

In Safari it also produces infinite prompts, the difference is that the prompt doesn't block the browser's UI and it's possible to close the tab. In Brave this basically blocks the whole browser.

stoletheminerals:
One issue I noticed, while it now stops prompts from appearing on the tab that calls these prompts, if you navigate to another tab or just press `All tabs` button, prompts will reappear again:

https://github.com/brave/brave-ios/assets/10810135/8e94c72c-5383-4385-a0b0-8ed4f8951d8b

soner-yuksel:
> One issue I noticed, while it now stops prompts from appearing on the tab that calls these prompts, if you navigate to another tab or just press `All tabs` button, prompts will reappear again:
> 

Added a check where the active controller is not bvc not to show external scheme alerts, this will cover all the action related with screen change imo.

stoletheminerals (approved):
lgtm now, thank you!

Brandon-T (approved):

PR #8288, opened 2023-10-19
Fix #8292: Change the minimum number of engines available

### LINKED ISSUE #8292: Increase the number of engines on older devices
### Description: 
The number of max engines is set to a value that is just too little. On an iPhone 6s it's 5 and on an iPhone X it's 15.

### Steps to Reproduce 
  1. Install the application on various devices
  2. Go to the filter list settings and enable as many engines as possible

### Actual result: 
4 on iPhone 6s, 14 on iPhone X

### Expected result:
The number is arbitrary and depends on the memory of the device but a floor of 19 and ceiling of 39 should be imposed on all devices. (note the total is N filter lists + 1 as the default filter list takes 1 spot)

### Brave Version: 

### Website problems only:

cuba:
## Summary of Changes

This pull request fixes #8292

## Submitter Checklist:

  - [ ] Light & dark mode

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/WebFilters/ShieldStats/Adblock/AdBlockStats.swift
+++ b/Sources/Brave/WebFilters/ShieldStats/Adblock/AdBlockStats.swift
@@ -13,9 +13,12 @@ import os.log
 public actor AdBlockStats {
   /// The max number of enabled filter lists depending on the amount memory available to the device
   static var maxNumberOfAllowedFilterLists: Int = {
-    let memory = Int(ProcessInfo.processInfo.physicalMemory / 1073741824)
-    ContentBlockerManager.log.debug("Memory: \(memory)")
-    return max(min(5 * memory, 40), 20)
+    // Get the number of gigs (memory in bytes divided by the size of a gig in bytes)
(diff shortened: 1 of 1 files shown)


soner-yuksel (commented):
Why? Can you please add a proper explanation here as a comment why max 20?

why are we first checking memory * 5

cuba:
> Why? Can you please add a proper explanation here as a comment why max 20?
> 
> why are we first checking memory * 5

Sorry the value is very misleading. I will rename it to `numberOfGigs`. It basically allows a minimum of 20 filter  lists and max of 40 filter lists with the actual value somewhere in between depending on the amount of memory (number of gigs) on the device.

cuba:
> Why? Can you please add a proper explanation here as a comment why max 20?
> 
> why are we first checking memory * 5

also max is a ceiling, not a floor so max(x, 20) is not a max of 20 but a minimum of 20...it's a max of the two values. Confuses me too. So first we take a y = max(x, 20) giving us a floor of 20, then we take a min(y, 40) giving us ceiling of 40.

soner-yuksel (approved):

PR #8306, opened 2023-10-23
Fix #8305 - Ref #8297: Allow Syncing All History - Update brave-core to 1.60.105

### LINKED ISSUE #8297: Bump Brave Core to 1.60.x
### Description: 

Bumping Brave-Core to latest 1.60.x

As per [brave-browser Release schedule](https://github.com/brave/brave-browser/wiki/Brave-Release-Schedule):
Channels will migrate October 25, 2023, Released on October 31, 2023, using Chromium 119

### LINKED ISSUE #8305: Allow Syncing All History
With Chromium enables History datatype and with CR119 the old TypedUrls datatype, which was used for History Sync, is wiped out from Chromium codebase.

The Desktop and Android change is covering most of the implementation for History API on iOS side with
https://github.com/brave/brave-core/pull/20329

Over that PR iOS side will have the changes with PR https://github.com/brave/brave-core/pull/20642

This PR will be merged together with CR119 uplift and "Allow Syncing All History" should be enabled with 1.60.

The reflection of this change on iOS repo is we have to remove typed url detection while adding item to History so All Item can be synced automatically. And since this separation is not needed anymore it has to be removed and cleaned from repository.

This will only effect History Items so The adding removing History with or without sync will be enough for testing.

soner-yuksel:
## Summary of Changes

This pull request fixes #8297
This pull request fixes #8305

This pull request is removing typedNavigation array and VisitType recording for navigation. This is necesserarry to allow all the History items to be synced properly.

https://github.com/brave/brave-core/pull/20654 CR119
1.60.105 brave-core has CR119 changes that includes Sync All History

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:
Test History adding / deleting with or without Sync Chain and History type is enabled.
To have Sync History work both recipients should be 1.60+ for both desktop and mobile.

Important not for any OS: only those history entries are synced, which were created after enabling history sync. Entries visited before enabling history sync are not sent to server and are not synced.

## Screenshots:

test2112

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Frontend/Browser/Toolbars/BottomToolbar/Menu/HistoryViewController.swift
+++ b/Sources/Brave/Frontend/Browser/Toolbars/BottomToolbar/Menu/HistoryViewController.swift
@@ -274,7 +274,7 @@ class HistoryViewController: SiteTableViewController, ToolbarUrlActionsProtocol 
       }
       
       dismiss(animated: true) {
-        self.toolbarUrlActionsDelegate?.select(url: url, visitType: .typed)
+        self.toolbarUrlActionsDelegate?.select(url: url)
       }
     }
 
--- a/Sources/Brave/Frontend/Browser/Toolbars/BottomToolbar/Menu/HistoryViewController.swift
+++ b/Sources/Brave/Frontend/Browser/Toolbars/BottomToolbar/Menu/HistoryViewController.swift
@@ -188,8 +188,11 @@ class HistoryViewController: SiteTableViewController, ToolbarUrlActionsProtocol 
       UIAlertAction(
         title: Strings.History.historyClearActionTitle, style: .destructive,
         handler: { [weak self] _ in
-          guard let self = self else { return }
-          
+          guard let self = self, let allHistoryItems = historyFRC?.fetchedObjects  else {
+            return
+          }
+
+          // Deleting Local History
           self.historyAPI.deleteAll {
             // Clearing Tab History with entire history entry
             self.tabManager.clearTabHistory() {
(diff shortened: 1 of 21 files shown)


Sources/Brave/Frontend/Browser/Toolbars/BottomToolbar/Menu/HistoryViewController.swift
```diff
@@ -199,6 +202,11 @@ class HistoryViewController: SiteTableViewController, ToolbarUrlActionsProtocol
             // Clearing History should clear Recently Closed
             RecentlyClosed.removeAll()
           }
+          
+          // Asking Sync Engine To Remove Visits
+          for historyItems in allHistoryItems {
+            self.historyAPI.removeHistory(historyItems)
```
Brandon-T, line 208: Hmm, when removing `ALL` history, is there no API in sync to ask it to remove all instead of the specific history above? Maybe we can export an API that can just remove all the history without the need to get the `fetchedObjects`?

Maybe not for now this PR though, but might be worth it in the future.

Brandon-T (approved):

Sources/Brave/Frontend/Browser/Toolbars/BottomToolbar/Menu/HistoryViewController.swift
soner-yuksel: The problem is this new BrowsingHistoryService doesnt have it. I checked ofcourse and normal History service method doesnt trigger Sync delete directives.

But later we can adjust add methods also using BrowsingHistoryService but I dont see any hurry to do it for bnow like you said.

PR #8473, opened 2023-11-23
Fix #8445: View recovery phrases button is not tappable

### LINKED ISSUE #8445: Unable to view seed phrase in Brave Wallet
The panel pops up, but the “view my recovery phrase” isn’t working correctly to display my seed phrase. @nuo-xu tested in 1.60 TF build and was reproducing as well.

nuo-xu:
## Summary of Changes
This issue is caused by the `ScrollView` has the wrong frame even though screen size is big enough for it to expand.
With further investigation, wrong frame caused by keyboard not being dismissed in the previous step which cause SwiftUI thinks the screen height is the actual height minus keyboard height. 
Attached screenshots below to see the height differences. 

This pull request fixes #8445

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:

Apparently, only @kdenhartog and I can repro this. I'm adding him as reviewer to see if he can verify.

## Screenshots:

When button is not tappable | When button is tappable
--- | ---
IMG_3281 | IMG_3282

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/BraveWallet/Crypto/Onboarding/BackupWalletView.swift
+++ b/Sources/BraveWallet/Crypto/Onboarding/BackupWalletView.swift
@@ -35,8 +35,11 @@ struct BackupWalletView: View {
       return
     }
     
-    // password filed resign first responder in case ScrollView gets changed
-    // which could effect the next step. Details see #8445
+    // Dismiss keyboard for password field before leaving this view #8445.
+    // Without dismissal, it's possible the scroll view in the detail view
+    // has incorrect frame (sitting above the area the keyboard used to be)
+    // which can show the scroll view contents out of the scroll bounds
+    // but will block tap interactions.
     resignFirstResponder()
     
     keyringStore.recoveryPhrase(password: password) { words in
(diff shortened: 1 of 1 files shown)


Sources/BraveWallet/Crypto/Onboarding/BackupWalletView.swift
```diff
@@ -34,6 +34,11 @@ struct BackupWalletView: View {
       // user can press return in field to execute when continue button is disabled
       return
     }
+    
+    // password filed resign first responder in case ScrollView gets changed
```
StephenHeaps: nit: hard to understand the actual issue / why this fixes it.
```suggestion
    // Dismiss keyboard for password field before leaving this view #8445.
    // Without dismissal, it's possible the scroll view in the detail view
    // has incorrect frame (sitting above the area the keyboard used to be)
    // which can show the scroll view contents out of the scroll bounds
    // but will block tap interactions.
```

StephenHeaps (approved):
I'm still unable to reproduce the bug, but I'm not seeing issues with this solution either (from Portfolio banner or Accounts tab). Just one nit to be a bit more descriptive, maybe we could also include the screenshot with red showing `ScrollView` bounds in the PR description.

kdenhartog (approved):
I'm unable to reproduce this anymore even when trying to get the keyboard to stay open like I had before, but tested this branch and found it was working. I'd think it would be fine to merge.

PR #8609, opened 2024-01-03
Fix #8608: Authenticate when disabling browser lock

### LINKED ISSUE #8608: Add Browser Lock Authentication when disabling Browser Lock
## Feature Overview 
- We should authenticate the user when attempting to disable browser lock from settings

Brandon-T:
## Summary of Changes
Fixes: https://hackerone.com/reports/2302346

This pull request fixes #8608

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:
1. Turn on Browser Lock (requires system passcode/faceid/touchid)
2. Turn it off.
3. It should require authentication to turn it off.

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Frontend/Settings/SettingsViewController.swift
+++ b/Sources/Brave/Frontend/Settings/SettingsViewController.swift
@@ -485,7 +485,7 @@ class SettingsViewController: TableViewController {
       let privateTabsRow = Row(
         text: Strings.TabsSettings.privateTabsSettingsTitle,
         selection: { [unowned self] in
-          let vc = UIHostingController(rootView: PrivateTabsView(tabManager: tabManager))
+          let vc = UIHostingController(rootView: PrivateTabsView(tabManager: tabManager, askForAuthentication: self.askForLocalAuthentication))
           self.navigationController?.pushViewController(vc, animated: true)
         },
         image: UIImage(braveSystemNamed: "leo.product.private-window"),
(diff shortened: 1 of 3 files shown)


Sources/Brave/Frontend/Settings/SettingsViewController.swift
```diff
@@ -656,11 +656,24 @@ class SettingsViewController: TableViewController {
...
       rows: [
-        .boolRow(
-          title: Strings.Privacy.browserLock,
+        Row(
+          text: Strings.Privacy.browserLock,
           detailText: Strings.Privacy.browserLockDescription,
-          option: Preferences.Privacy.lockWithPasscode,
-          image: UIImage(braveSystemNamed: "leo.biometric.login")),
+          image: UIImage(braveSystemNamed: "leo.biometric.login"),
+          accessory: .view(SwitchAccessoryView(initialValue: Preferences.Privacy.lockWithPasscode.value, valueChange: { isOn in
```
kylehickinson, line 663: `boolRow` has an argument that takes a `toggleValue` and `valueChange` argument to do just this fyi

Brandon-T: But no image argument, so that's why I used this one.

stoletheminerals:
@Brandon-T can you also add it to this switch?

Brandon-T:
> @Brandon-T can you also add it to this switch?
> 
> 
> 
> 
> 
> 
> 
> 

I did. When you press the switch, it asks for auth.

stoletheminerals (approved):

Sources/Brave/Frontend/Settings/Tabs/PrivateTabsView.swift
```diff
@@ -86,7 +89,15 @@ struct PrivateTabsView: View {
...
           OptionToggleView(title: browsingLockTitle,
                            subtitle: nil,
-                           option: Preferences.Privacy.privateBrowsingLock)
+                           option: privateBrowsingLock) { newValue in
+            if !newValue {
+              askForAuthentication(.general) { success, _ in
+                if success {
+                  privateBrowsingLock.value = newValue
+                }
+              }
```
kylehickinson: Brought up in DM but this will still update the preference immediately and currently does not reset it back to false if auth fails. Likely need to replace this with the underlying view that gives you access to a Binding, then create a custom Binding that doesn't set the pref to false at all until auth completes

kylehickinson (changes requested):

Sources/Brave/Frontend/Settings/Tabs/PrivateTabsView.swift
stoletheminerals: Thanks @kylehickinson , good catch! I tried this with `Browser lock` switch and it didn't work, but didn't check `Private tabs`.

Brandon-T: Fixed so the preference only changes after authentication.

kylehickinson (approved):

PR #8651, opened 2024-01-12
Fix #8650: Fix SSL Status on Offline pages and Http pages

### LINKED ISSUE #8650: SSL Error shows on interstitial pages
### Description: 
 - When on an interstitial page such as the "No Internet Connection" page, the page shows as `insecure` which can be misleading.
 - When on a `http` page, it shows the wrong status (warning) instead of (insecure).

### Steps to Reproduce 
  1. Test `http://neverssl.com`
  2. Test turning off internet on any page and refreshing it until you get the "No internet connection" interstitial page

### Actual result: 
  1. Shows "Warning" instead of "Insecure"
  2. Shows "Insecure" instead of "Secure"

### Expected result:
  1. Should show "Insecure"
  2. Should show "Secure"

- Easily Reproduced

### Brave Version: 

- 1.61.x, 1.62.x

### Device details: 
- Any

### Website problems only:
- Yes

Brandon-T:
## Summary of Changes
- When there is no internet or we're on an internal page, we need to show our internal page as secure ONLY IF the page is not an invalid cert page.

- When on an internal page, no serverTrust exists, and hasSecureContent is triggered, causing the page to show as insecure since SSL is missing. We now double check that it is actually insecure. SSL security takes priority over mixed-content.

This pull request fixes #8650

## Submitter Checklist:

  - [x] Light & dark mode

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`

### DIFF
--- a/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+WKNavigationDelegate.swift
+++ b/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+WKNavigationDelegate.swift
@@ -67,11 +67,14 @@ extension BrowserViewController: WKNavigationDelegate {
     // check if web view is loading a different origin than the one currently loaded
     if let selectedTab = tabManager.selectedTab,
        selectedTab.url?.origin != webView.url?.origin {
-      // reset secure content state to unknown until page can be evaluated
-      if let url = webView.url, !InternalURL.isValid(url: url) {
-        selectedTab.secureContentState = .unknown
-        updateToolbarSecureContentState(.unknown)
+      if let url = webView.url {
+        if !InternalURL.isValid(url: url) {
+          // reset secure content state to unknown until page can be evaluated
+          selectedTab.secureContentState = .unknown
+          updateToolbarSecureContentState(.unknown)
+        }
       }
+      
       // new site has a different origin, hide wallet icon.
       tabManager.selectedTab?.isWalletIconVisible = false
       // new site, reset connected addresses
@@ -672,8 +675,18 @@ extension BrowserViewController: WKNavigationDelegate {
 
   public func webView(_ webView: WKWebView, didCommit navigation: WKNavigation!) {
     guard let tab = tab(for: webView) else { return }
+
     // Set the committed url which will also set tab.url
     tab.committedURL = webView.url
+    
+    // Server Trust and URL is also updated in didCommit
+    // However, WebKit does NOT trigger the `serverTrust` observer when the URL changes, but the trust has not.
+    // WebKit also does NOT trigger the `serverTrust` observer when the page is actually insecure (non-https).
+    // So manually trigger it with the current trust.
+    observeValue(forKeyPath: KVOConstants.serverTrust.keyPath,
+                 of: webView,
+                 change: [.newKey: webView.serverTrust as Any, .kindKey: 1],
+                 context: nil)
     
     // Need to evaluate Night mode script injection after url is set inside the Tab
     tab.nightMode = Preferences.General.nightModeEnabled.value
--- a/Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController.swift
+++ b/Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController.swift
@@ -1817,9 +1817,33 @@ public class BrowserViewController: UIViewController {
         if let url = tab.webView?.url, url.isReaderModeURL {
           break
         }
+        
         tab.secureContentState = .mixedContent
       }
-
+      
+      if let url = tab.webView?.url,
+         InternalURL.isValid(url: url),
+         let internalUrl = InternalURL(url) {
+        
+        if internalUrl.isErrorPage {
+          if ErrorPageHelper.certificateError(for: url) != 0 {
+            // Cert validation takes precedence over all other errors
+            tab.secureContentState = .invalidCert
+          } else if NetworkErrorPageHandler.isNetworkError(errorCode: ErrorPageHelper.errorCode(for: url)) {
+            // Network error takes precedence over missing cert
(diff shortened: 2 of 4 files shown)


Sources/Brave/Frontend/Browser/BrowserViewController/BVC+WKNavigationDelegate.swift
```diff
@@ -67,11 +67,14 @@ extension BrowserViewController: WKNavigationDelegate {
     // check if web view is loading a different origin than the one currently loaded
     if let selectedTab = tabManager.selectedTab,
        selectedTab.url?.origin != webView.url?.origin {
-      // reset secure content state to unknown until page can be evaluated
-      if let url = webView.url, !InternalURL.isValid(url: url) {
-        selectedTab.secureContentState = .unknown
-        updateToolbarSecureContentState(.unknown)
+      if let url = webView.url {
```
soner-yuksel, line 70: Is this change necessary here ?

Sources/Brave/Frontend/Browser/BrowserViewController/BVC+WKNavigationDelegate.swift
```diff
@@ -672,9 +675,19 @@ extension BrowserViewController: WKNavigationDelegate {
...
     guard let tab = tab(for: webView) else { return }
+
     // Set the committed url which will also set tab.url
     tab.committedURL = webView.url
     
+    // Server Trust and URL is also updated in didCommit
+    // However, WebKit does NOT trigger the `serverTrust` observer when the URL changes, but the trust has not.
+    // WebKit also does NOT trigger the `serverTrust` observer when the page is actually insecure (non-https).
+    // So manually trigger it with the current trust.
+    observeValue(forKeyPath: KVOConstants.serverTrust.keyPath,
```
soner-yuksel, line 686: Isnt it supposed to be marked under main queue?

Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -1817,9 +1817,33 @@ public class BrowserViewController: UIViewController {
         if let url = tab.webView?.url, url.isReaderModeURL {
           break
         }
+        
         tab.secureContentState = .mixedContent
       }
-
+      
+      if let url = tab.webView?.url,
```
soner-yuksel, line 1824: Hmm this is hasOnlySecureContent right ? which should mean **web view loaded all resources on the page through securely encrypted connections**. Is there a reason we are checking internal is error ?

Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -1848,10 +1872,19 @@ public class BrowserViewController: UIViewController {
             internalUrl.isErrorPage {
 
             if ErrorPageHelper.certificateError(for: url) != 0 {
+              // Cert validation takes precedence over all other errors
               tab.secureContentState = .invalidCert
+            } else if NetworkErrorPageHandler.isNetworkError(errorCode: ErrorPageHelper.errorCode(for: url)) {
+              // Network error takes precedence over missing cert
+              // Because we cannot determine if a cert is missing yet, if we cannot connect to the server
+              // Our network interstitial page shows
+              tab.secureContentState = .localhost
```
soner-yuksel, line 1881: Oh now this makes a lot of sense actually

Sources/Brave/Frontend/Browser/Helpers/ErrorPageHelper.swift
```diff
@@ -135,18 +135,26 @@ class ErrorPageHelper {
...
-
+      
+      return errCode
+    }
+    return 0
+  }
+  
+  static func certificateError(for url: URL) -> Int {
+    let errCode = errorCode(for: url)
+    if errCode != 0 {
```
soner-yuksel, line 163: Can you add comment here what does error code 0 mean and why do we limit that code?

Sources/Brave/Frontend/Browser/BrowserViewController/BVC+WKNavigationDelegate.swift
Brandon-T: No. That will create a race condition if I Dispatch it. It should be done exactly in-line in the commit, right after `tab.url` is assigned. So that cert status and URL is updated at the exact same time.

This is already on the Main Queue because `didCommit` is called on main :) So no need to dispatch again which can delay the update.

Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController.swift
Brandon-T: Yes. When you're on a local-host page such as the interstitial pages, it will always show as `hasSecureContentOnly = false` because there's no local ssl cert for those pages.

So when you're offline for example, it will show as insecure even though it's not (no evaluation was done). But it's insecure because `hasSecureContentOnly = false`. So in addition to that check, we have to check if it's our interstitial pages. We know they're secure because they're ours and it's local and it's offline. So we override the check and mark it secure.

Sources/Brave/Frontend/Browser/Helpers/ErrorPageHelper.swift
Brandon-T: It means there's none. When there is an errorCode, `CFNetworkError(rawValue: ...)` minimum value is `1`, not `0`. So `0 = errSecNo`. IE: 0 = no error.

soner-yuksel (approved):

soner-yuksel (approved):

PR #8729, opened 2024-02-02
Use list_catalog.json file in iOS

### LINKED ISSUE #35210: Expose the first party list in iOS
## Description 

We don't use the first party list in iOS yet. We need to make this list available to iOS

Note: this is a temporary solution. A real solution should be [here](https://github.com/brave/brave-browser/issues/35209), which is a much bigger and invasive change which has a ton of other benefits.

Additional work will need to be done in iOS to use these new services:
https://github.com/brave/brave-ios/issues/7611

cuba:
## Summary of Changes

This pull request fixes https://github.com/brave/brave-browser/issues/35210

MUST Be merged in response to https://github.com/brave/brave-core/pull/21883

This actually has a number of benefits:
1. All lists are now filter lists. so no additional code if we add new hidden lists such as first party list or iOS only lists
2. Will very likely use this mechanism for any launch only lists too to increase launch performance (but this is not yet done and no such list exists yet)
3. Brings us one step closer to how desktop functions.

## Submitter Checklist:

  - [x] Light & dark mode

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


cuba:
moved to brave-core

PR #8749, opened 2024-02-08
Fix compiling only necessary files during launch

cuba:
## Summary of Changes

This PR prepares for the list_catalog.json and fixes a number of issues during launch

1. Don't compile custom filter lists upon launch (this could have been causing load hangs for people with custom filter lists)
2. Explicitly don't compile shown filter lists upon launch. (it will compile hidden filter lists but those do not appear until we switch to list_catalog.json). There is no change in behaviour because there was a bug in the code which causes no filter lists to compile
3. Add signposts for some better launch profiling

Note except for part 2, none of this should affect launch performance. Later work will be done to improve launch performance (we will attempt to save and load dat files in the future which should give a performance boost)

## Submitter Checklist:

  - [x] Light & dark mode

## Test Plan:

Test adblocking and launch performance.

## Reviewer Checklist:

  - `QA/(Yes|No)`
  - `bug` / `enhancement`


cuba:
merged with brave-core's list catalog PR

